PR #11, opened 2023-06-02
feat: integrate sdk + nft contract for onboarding

rthomare:
# This Stack
This pull request introduces a comprehensive end-to-end decentralized application that serves as a prime example for users of our SDK, providing a practical demonstration of various interactions and functionalities, fostering a solid foundation for their own development endeavors.

# This PR
Compiles and update NFT Contract and ABI, and adds first version of onboarding

# Test Plan
https://github.com/alchemyplatform/aa-sdk/assets/3498866/f7c29d10-3fbd-4c75-a670-0fc30955806a

PR #12, opened 2023-06-02
feat: integrate sdk and refactor onboarding

rthomare:
# This Stack
This pull request introduces a comprehensive end-to-end decentralized application that serves as a prime example for users of our SDK, providing a practical demonstration of various interactions and functionalities, fostering a solid foundation for their own development endeavors.

# This PR
- Refactors onboarding, files structures, and onboarding
- Also adopts SDK into onboarding.

# Test Plan
https://github.com/alchemyplatform/aa-sdk/assets/3498866/f7c29d10-3fbd-4c75-a670-0fc30955806a


examples/alchemy-daapp/package.json
```diff
@@ -9,6 +9,7 @@
     "lint": "next lint"
   },
   "dependencies": {
+    "@alchemy/aa-core": "file:../../packages/core",
```
moldy530, line 12: oh... hmm what's going on here? does this not work with just `"*"` as the version?

moldy530 (commented):
looks like something funky with git history here... this seems really similar to https://github.com/alchemyplatform/aa-sdk/pull/11

examples/alchemy-daapp/package.json
rthomare: if we do * it'll ask for a version whenever we run yarn. Doing it this way forces the example to use the version that is specified in the core package.json.

rthomare: Screenshot 2023-06-05 at 10 00 20 PM

packages/core/inject-version.ts
```diff
@@ -1,13 +1,14 @@
...
 import { dirname, resolve } from "path";
 import { fileURLToPath } from "url";
-import packageJSON from "./package.json" assert { type: "json" };
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
 const VERSION_FILE_PATH = "src/version.ts";
 
+const packageJSON = JSON.parse(readFileSync("./package.json").toString());
```
rthomare, line 10: Stops tsc from having json import issues.

moldy530: oh weird... it shouldn't be having issues since we have `"resolveJsonModule" : true` in compiler options and the `assert `on the import. what version of TS are you on?

Issue #15, opened 2023-06-04
Running aa-sdk in AWS lambda to fire user operations


yashh:
### [REQUIRED] Environment

- Browser version:
- AA SDK version:
- Package:

#### Relevant code or sample repro:

```
import { Context, APIGatewayProxyResult, APIGatewayEvent } from 'aws-lambda';
import {
    alchemyPaymasterAndDataMiddleware,
    SimpleSmartContractAccount,
    SmartAccountProvider,
    type SimpleSmartAccountOwner,
} from "@alchemy/aa-core";
import { mnemonicToAccount } from "viem/accounts";
import { sepolia } from "viem/chains";
import { toHex } from "viem";
import fetch from "node-fetch";
global.fetch = fetch;


const SIMPLE_ACCOUNT_FACTORY_ADDRESS =
    "0x9c7938fCaaDd7Ad94F6B5b4A2fd4efb8219Exxxx";
const MNEMONIC = process.env.MNEMONIC
const ENDPOINT = "https://eth-sepolia.g.alchemy.com/v2/xxxxxx"
const CHAINID = sepolia
const ENTRYPOINT = "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2790"
const PAYMASTER_POLICY_ID = "xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxx"


// 1. define the EOA owner of the Smart Account
// This is just one exapmle of how to interact with EOAs, feel free to use any other interface
const ownerAccount = mnemonicToAccount(MNEMONIC);
// All that is important for defining an owner is that it provides a `signMessage` and `getAddress` function
const owner: SimpleSmartAccountOwner = {
    // this should sign a message according to ERC-191
    signMessage: async (msg) =>
        ownerAccount.signMessage({
            message: toHex(msg),
        }),
    getAddress: async () => ownerAccount.address,
};

// 2. initialize the provider and connect it to the account
let provider = new SmartAccountProvider(
    ENDPOINT, // rpcUrl
    ENTRYPOINT, // entryPointAddress
    CHAINID // chain
).connect(
    (rpcClient) =>
        new SimpleSmartContractAccount({
            entryPointAddress: ENTRYPOINT,
            chain: CHAINID,
            factoryAddress: SIMPLE_ACCOUNT_FACTORY_ADDRESS,
            rpcClient,
            owner,
            // optionally if you already know the account's address
[...]

PR #16, opened 2023-06-05
chore: update to latest version of viem and update docs

moldy530:
This updates to the latest version of viem and updates the README to fix an error in the `sendUserOperation` example call.

- A breaking change was introduced in the last merge and was not reflected in the README properly. (See https://github.com/alchemyplatform/aa-sdk/issues/15)

- I made viem a peerDependency so that the library can use the users local viem. This way, users don't need to wait for us to merge changes if they want to use the latest viem version

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,16 @@ Alchemy's Account Abstraction Kit is an SDK that enables easy interactions with 
 via `yarn`:
 
 ```bash
-yarn add @alchemy/aa-core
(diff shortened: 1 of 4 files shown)


rthomare (approved):

moldy530:
@moldy530 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/16).

PR #19, opened 2023-06-08
refactor: modify pamyaster middleware to be able to account for gas

moldy530:


packages/core/src/provider/base.ts
```diff
@@ -267,6 +267,10 @@ export class SmartAccountProvider<
     struct: UserOperationStruct
   ): Promise<UserOperationStruct> => {
     struct.paymasterAndData = "0x";
+    // by default add a 25k overhead to verification gas limit
+    struct.verificationGasLimit =
```
dancoombs, line 271: This is no longer needed, rundler does this automatically

moldy530: ok cool will drop this and then it should be sufficient to just support overriding gas on paymaster middleware

moldy530:
I'll update the docs for the paymaster middleware and add to the readme. We might want to use something like JS-docs or something to auto gen documentation so we can just keep it in the comments on the methods

dancoombs (approved):

PR #21, opened 2023-06-12
Kernel Smart Account Wallet Support

whatmanathinks:


packages/core/src/provider/base.ts
```diff
@@ -40,7 +40,7 @@ export const noOpMiddleware: AccountMiddlewareFn = async (
 ) => struct;
 
 // borrowed from ethers.js
-function defineReadOnly<T, K extends keyof T>(
+export function defineReadOnly<T, K extends keyof T>(
```
moldy530, line 43: we might want to move this to a `utils` file / folder

packages/core/src/signer/hd-account.ts
```diff
@@ -0,0 +1,31 @@
+import {SmartAccountSigner} from "../interfaces/signer";
+import {Address} from "abitype";
+import {HDAccount, HDOptions, Hex, toHex} from "viem";
+import {mnemonicToAccount, privateKeyToAccount} from "viem/accounts";
+
+export class HdAccountSigner implements SmartAccountSigner {
```
moldy530, line 6: this is nice!

packages/core/src/account/base.ts
```diff
@@ -30,7 +36,7 @@ export abstract class BaseSmartContractAccount<
   TTransport extends SupportedTransports = Transport
 > implements ISmartContractAccount
 {
-  protected isDeployed?: boolean = undefined;
+  protected isDeployed?: DeploymentState = DeploymentState.undefined;
```
moldy530, line 39: nit: this doesn't need to be marked as `?`

packages/core/src/account/base.ts
```diff
@@ -17,6 +17,12 @@ import type {
 import type { ISmartContractAccount } from "./types.js";
 import type { BatchUserOperationCallData } from "../types.js";
 
+export enum DeploymentState {
+  undefined = '0x0',
+  not_deployed = '0x1',
+  deployed = '0x2',
```
moldy530, line 23: nit: can we have the keys here be all uppercase?

moldy530 (commented):
thanks for contributing to our SDK @ManaOmlabs! 

This looks like a great use of it, but had a couple notes:
1. I think we should make the changes proposed here to `BaseAccount` and the introduction of the `hd-account` signer impl is a great utility that I agree is lacking
2. the actual implementation of the Kernel account should live in a separate package and potentially in a separate repo that depends on this SDK. I was going to remove the `alchemyPaymaster` and other alchemy specific configs from `core` and move them to an alchemy specific package. The goal of the `core` repo was to serve as an extensible base for interacting with erc-4337. 

btw is kernel account here referencing a ZeroDev kernel smart contract account? curious what's backing this

keon:
> the actual implementation of the Kernel account should live in a separate package and potentially in a separate repo that depends on this SDK. I was going to remove the alchemyPaymaster and other alchemy specific configs from core and move them to an alchemy specific package. The goal of the core repo was to serve as an extensible base for interacting with erc-4337.

Creating our own package makes sense, for our immediate use-case. But for the package to be extensible and support multiple smart contract standards, I wish there was a clean separation between the core and its extensions (or presets). I don't have any idea right now, but will share some thoughts as I gain more insights. 

> btw is kernel account here referencing a ZeroDev kernel smart contract account? 

Yes, it is ZeroDev kernel, we found it simple & extensible among the smart contract wallets that we can productionize immediately.

ManaOmlabs:
> btw is kernel account here referencing a ZeroDev kernel smart contract account? curious what's backing this

yes kernel is zerodev's kernel smart contrac

packages/core/src/provider/base.ts
ManaOmlabs: noted

ManaOmlabs:
> 2\. the actual implementation of the Kernel account should live in a separate package and potentially in a separate repo that depends on this SDK. I was going to remove the `alchemyPaymaster` and other alchemy specific configs from `core` and move them to an alchemy specific package. The goal of the `core` repo was to serve as an extensible base for interacting with erc-4337.



Yes maybe we can have core package where there just is just the core interface implementation and then we can have dedicated package where everybody can contribute and add different implementations built on top of it (wallet contracts, paymasters, bundlers, etc). It can sort of work like giant template factory everybody references from

moldy530:
yea I kinda like the idea of having a package like `aa-contracts` and structuring it similarly to how rainbowkit has a `walletConnectors` folder. Lemme think about it a bit more because it looks like some contracts require a different `Provider` impl or extension of the `SmartAccountProvider`. The ideal case is that the contract defs will work with the base provider so that usage is plug and play via `connect()`

cc @rthomare and @dancoombs

packages/core/src/interfaces/signer.ts
```diff
@@ -0,0 +1,6 @@
+import {Address} from "abitype";
+
+export interface SmartAccountSigner {
```
moldy530, line 3: the rest of the repo follows the pattern of having these live in a `types.ts` file in their namespace, so would probs put this in `signer/types.ts` alongside the `hd-account`

packages/core/src/account/kernel.ts
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+            return this.encodeExecuteAction(target, value, data, 0n)
+        }
+    }
+
+    async encodeExecuteDelegate (target: Hex, value: bigint, data: Hex): Promise<Hex> {
+        return this.encodeExecuteAction(target, value, data, 1n)
+    }
+
+    async signWithEip6492(msg: string | Uint8Array | Hex): Promise<Hex> {
```
moldy530, line 63: I'm actually not super familiar with 6492, so this might be a dumb question. Should this not just be the default `signMessage` implementation below rather than having 2 different sign message functions? I imagine the factory supports the 6492 since it's being used to sign here, so there shouldn't be any issues on account creation right?

packages/core/src/__tests__/kernel-account.test.ts
```diff
@@ -0,0 +1,199 @@
...
+import {concatHex, encodeAbiParameters, Hex, parseEther} from "viem";
+import {polygonMumbai} from "viem/chains";
+import {mnemonicToAccount} from "viem/accounts";
+import {parseAbiParameters} from "abitype";
+
+
+describe("Kernel Account Tests", () => {
+    const chain = polygonMumbai
+    const config: KernelHdConfig = {
+        mnemonic: process.env.OWNER_MNEMONIC,
```
moldy530, line 11: currently working through a way to test this stuff without requiring an environment specific mnemonic (these tests definitely won't pass for me since I'm using a different mnemonic from y'all)

moldy530:
@keon @ManaOmlabs 

What I'll do is create a sub package tomorrow that's empty with readme on adding new account types for support -- `@alchemy/aa-accounts`

Will update y'all here when it's done

packages/core/src/__tests__/kernel-account.test.ts
ManaOmlabs: Will push some more changes today which would allow tests tests to be run irrespective of wallet

moldy530:
@ManaOmlabs @keon PR is up here https://github.com/alchemyplatform/aa-sdk/pull/23

Lemme know what y'all think about the structure outlined in the readme of the PR. Since y'all are the first contributors would be great to have y'all's feedback there too

ManaOmlabs:
> @ManaOmlabs @keon PR is up here #23
> 
> Lemme know what y'all think about the structure outlined in the readme of the PR. Since y'all are the first contributors would be great to have y'all's feedback there too

@moldy530 overall looks good. 
Contributors may also add signers. If signer is based on generic viem implementations, then add in core else add within their implementation if customization needed.


I will also modify PR to have the latest changes and segregate both. Test cases also will be pushed today (wallet agnostic ones)

packages/core/src/account/base.ts
```diff
@@ -30,7 +36,7 @@ export abstract class BaseSmartContractAccount<
   TTransport extends SupportedTransports = Transport
 > implements ISmartContractAccount
 {
-  protected isDeployed?: boolean = undefined;
+  protected isDeployed: DeploymentState = DeploymentState.UNDEFINED;
```
moldy530, line 39:
```suggestion
  protected deploymentState: DeploymentState = DeploymentState.UNDEFINED;
```

packages/accounts/src/kernel-zerodev/validator/base.ts
```diff
@@ -0,0 +1,52 @@
...
+    abstract async signMessage(message: string | Uint8Array | Hex): Promise<Hex>
+
+    getAddress(): Hex {
+        return this.validatorAddress
+    }
+
+    async getOwner (): Promise<Hex> {
+        return this.owner.getAddress();
+    }
+    async getSignature(userOpHash: Uint8Array | string | Hex): Promise<Hex> {
```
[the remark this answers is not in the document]
moldy530: is this meant to be used exclusively for userOp signing?

packages/core/src/account/kernel.ts
ManaOmlabs: the entrypoint contract doesn't accept it yet

packages/core/src/account/base.ts
ManaOmlabs: This is fixed in separate PR

packages/core/src/account/base.ts
```diff
@@ -117,4 +123,18 @@ export abstract class BaseSmartContractAccount<
 
     return this.accountAddress;
   }
+
+
+  // Extra implementations
+  async isAccountDeployed(): Promise<Boolean> {
```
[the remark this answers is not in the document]
ManaOmlabs: done

packages/accounts/src/kernel-zerodev/validator/base.ts
```diff
@@ -0,0 +1,52 @@
...
+    // enable = '0x00000002',
+}
+
+export interface BaseValidatorParams {
+    validatorAddress: Hex
+    mode: ValidatorMode
+    owner: SmartAccountSigner
+}
+
+export abstract class BaseValidator {
```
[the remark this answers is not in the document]
ManaOmlabs: It is a contruct introduced by kernel to separate validation phase. It expects few extra bits to signature based on config

packages/accounts/src/kernel-zerodev/validator/ecdsa.ts
```diff
@@ -0,0 +1,13 @@
...
+import {Hex, hexToBytes} from "viem";
+
+export class ECDSAValidator extends BaseValidator {
+    async signMessage(message: string | Uint8Array | Hex): Promise<Hex> {
+        if (typeof message === "string" && message.startsWith("0x")) {
+            message = hexToBytes(message as Hex);
+        } else if (typeof message === "string") {
+            message = new TextEncoder().encode(message);
+        }
+        return this.owner.signMessage(message)
```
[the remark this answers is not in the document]
ManaOmlabs: Moved. thanks

packages/accounts/src/kernel-zerodev/validator/base.ts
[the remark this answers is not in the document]
ManaOmlabs: There actually wasn't difference. Corrected it 

Pls review https://github.com/alchemyplatform/aa-sdk/pull/26/files

I have segregated kernell and this PR so that we can release core first with base changes and then use core version here.

@moldy530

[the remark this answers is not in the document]
moldy530: just reviewd #26

packages/accounts/src/index.ts
```diff
@@ -1,2 +1,11 @@
...
 // Don't use wildcard exports, instead use named exports
+
+//kernel exports
+export { KernelFactoryAbi } from "./kernel-zerodev/abis/KernelFactoryAbi";
+export { KernelAccountAbi } from "./kernel-zerodev/abis/KernelAccountAbi";
+export type { KernelSmartAccountParams } from "./kernel-zerodev/account";
+export { KernelSmartContractAccount } from "./kernel-zerodev/account";
+export { KernelAccountProvider } from "./kernel-zerodev/provider";
+export { BaseValidator } from "./kernel-zerodev/validator/base";
+export { ECDSAValidator } from "./kernel-zerodev/validator/ecdsa";
```
moldy530, line 11: one nit on these names, maybe we should also prefix these with `Kernel*` as well. That way if some other account also has a concept of validators they won't clash

packages/accounts/src/kernel-zerodev/provider.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    constructor(
+        rpcProvider: string | PublicErc4337Client<TTransport>,
+        private entryPointAddress: Address,
+        private chain: Chain,
+        readonly account?: KernelSmartContractAccount,
+        opts?: SmartAccountProviderOpts
+    ) {
+        super(rpcProvider,entryPointAddress,chain,account,opts)
+    }
```
moldy530, line 24: is this constructor necessary? the `entryPointAddress` and `chain` are now protected properties. Only thing I can see that's specific is the requirement for `KernelSmartContractAccount` on the account type.

we could update the interface for `ISmartAccountProvider` and the base provider to accept one more generic:
```ts
interface ISmartAccountProvider<
  TTransport extends SupportedTransports = Transport,
  TSmartContractAccount extends BaseSmartContractAccount = BaseSmartContractAccount,
>
```
and then here you would have 

```ts
export class KernelAccountProvider<
    TTransport extends SupportedTransports = Transport
> extends SmartAccountProvider<TTransport, KernelSmartContractAccount>
```

this would reduce the boilerplate for future contributors

packages/accounts/src/kernel-zerodev/account.ts
```diff
@@ -0,0 +1,119 @@
...
+    TTransport extends Transport | FallbackTransport = Transport
+> extends BaseSmartAccountParams<TTransport> {
+    owner: SmartAccountSigner;
+    factoryAddress: Address;
+    index?: bigint;
+    defaultValidator: BaseValidator
+    validator?: BaseValidator
+}
+
+export class KernelSmartContractAccount<
```
moldy530, line 27: do ZeroDev kernel account support batch executions?

packages/accounts/src/index.ts
moldy530, line 11: also I'm not seeing the `ecdsa` validator file >.>

moldy530 (commented):
just had a few more comments, but LGTM otherwise

packages/accounts/src/kernel-zerodev/account.ts
ManaOmlabs: It does. But we will add implementation in later PR. as of now, it will show not supported

packages/accounts/src/index.ts
ManaOmlabs: Yeah, still need to clean up these today

packages/accounts/src/kernel-zerodev/provider.ts
ManaOmlabs: Noted. will do

packages/accounts/src/index.ts
ManaOmlabs: Ok

moldy530:
Ah one more comment -- once other changes are addressed, it'd be good to add a README.md in side `src/accounts/kernel-zerodev` for consumers to reference

packages/accounts/src/kernel-zerodev/provider.ts
ManaOmlabs: https://github.com/alchemyplatform/aa-sdk/pull/28/files

packages/core/src/__tests__/kernel-account.test.ts
ManaOmlabs: This is sorted now. You should be able to run test cases irrespective of wallet @moldy530

ManaOmlabs:
@moldy530 
- test cases are now generic
- made changes based on all review comments.
-  added `readme.md` file. 

Before merging this, we need to merge this fix
https://github.com/alchemyplatform/aa-sdk/pull/28/files

ManaOmlabs:
@moldy530 let me know if any more comments in this PR :)

packages/accounts/src/kernel-zerodev/provider.ts
```diff
@@ -0,0 +1,46 @@
...
+export class KernelAccountProvider extends SmartAccountProvider<HttpTransport> {
+  gasEstimator: AccountMiddlewareFn = async (struct) => {
+    const request = deepHexlify(await resolveProperties(struct));
+    const estimates = await this.rpcClient.estimateUserOperationGas(
+      request,
+      this.entryPointAddress
+    );
+
+    estimates.verificationGasLimit =
+      (BigInt(estimates.verificationGasLimit) * 130n) / 100n;
```
moldy530, line 18: oh interesting, where's the extra 30% overhead come from?

moldy530 (approved):
LGTM had a question about the gasEstimator change for KernelProvider, but not blocker

packages/accounts/src/kernel-zerodev/provider.ts
ryanli-me: curious to learn about this 30% overhead too

PR #22, opened 2023-06-12
feat: add alchemy sub-package

moldy530:
This adds a new subpackage that can be used if you're interacting with Alchemy infrastructure exclusively.

The new AlchemyProvider handles all of the configuration that matches our Rundler and Gas Manager configs
and makes it easier to interact with alchemy in general.

TODO: Once PR is approved, update the README to include docs on how to use the new package + Provider (tests also serve as decent docs for now too)


moldy530:
* main
  * **PR #22** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/22" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/22?utm_source=stack-comment).

packages/alchemy/src/chains.ts
```diff
@@ -0,0 +1,50 @@
...
+  // testnets
+  defineChainStrategy(goerli.id, GasFeeStrategy.FIXED, 0n),
+  defineChainStrategy(sepolia.id, GasFeeStrategy.FIXED, 0n),
+  defineChainStrategy(polygonMumbai.id, GasFeeStrategy.FIXED, 0n),
+  defineChainStrategy(optimismGoerli.id, GasFeeStrategy.FIXED, 0n),
+  defineChainStrategy(arbitrumGoerli.id, GasFeeStrategy.FIXED, 0n),
+  // mainnets
+  defineChainStrategy(mainnet.id, GasFeeStrategy.PRIORITY_FEE_PERCENTAGE, 25n),
+  defineChainStrategy(polygon.id, GasFeeStrategy.PRIORITY_FEE_PERCENTAGE, 25n),
+  defineChainStrategy(optimism.id, GasFeeStrategy.PRIORITY_FEE_PERCENTAGE, 5n),
```
dancoombs, line 48: This should be `BASE_FEE_PERCENTAGE`

packages/alchemy/src/middleware/gas-fees.ts
```diff
@@ -0,0 +1,56 @@
...
+    const block = await provider.rpcClient.getBlock({ blockTag: "latest" });
+    const baseFeePerGas = block.baseFeePerGas;
+    if (baseFeePerGas == null) {
+      throw new Error("baseFeePerGas is null");
+    }
+    const maxPriorityFeePerGas = BigInt(
+      await provider.rpcClient.getMaxPriorityFeePerGas()
+    );
+    // add 25% overhead to ensure mine
+    // because we're using bigint, 5n/4n = 1, so we need to scale by 100
```
dancoombs, line 33: Really? Even if you multiply by 5 first and then divide by 4?

packages/alchemy/src/middleware/gas-fees.ts
```diff
@@ -0,0 +1,56 @@
...
+    const baseFeeScaled = (baseFeePerGas * 125n) / 100n;
+
+    const prioFee = ((): bigint => {
+      switch (feeMode.strategy) {
+        case GasFeeStrategy.FIXED:
+          return maxPriorityFeePerGas + feeMode.value;
+        case GasFeeStrategy.BASE_FEE_PERCENTAGE:
+          return (baseFeeScaled * feeMode.value) / 100n;
+        case GasFeeStrategy.PRIORITY_FEE_PERCENTAGE:
+          // add 10% to required priority fee to ensure mine
```
dancoombs, line 43: Not for this PR, but a general comment. These arbitrary % increases were mostly random guesses by me. We should strive to provide more flexibility for users.

The solution here that @dphilipson and I discussed is some sort of `rundler_feeHistory` endpoint that allows users to estimate their base/priority fees based on prior value and how quickly they want their UO mined. Similar to those fast/normal/slow options that wallets sometimes provide.

When ERC4337 moves to its P2P mempool, these calculations definitely won't apply, and we will need better tracking of mempool priority fees to give users better hints to get their transaction mined at the lowest cost.

packages/alchemy/src/chains.ts
moldy530: Ah was going off of: https://github.com/OMGWINNING/rundler-infra/pull/32/files#diff-33c874a999d6039f3477b2d5511a90d4fcf9ac41b3548913cf802de517ecfafcR100

packages/alchemy/src/middleware/gas-fees.ts
moldy530: pre-coffee brain fart... yea it works just fine with that order because math

packages/alchemy/src/__tests__/simple-account.test.ts
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  it("should correctly sign the message", async () => {
+    expect(
+      // TODO: expose sign message on the provider too
+      await signer.account.signMessage(
+        "0xa70d0af2ebb03a44dcd0714a8724f622e3ab876d0aa312f0ee04823285d6fb1b"
+      )
+    ).toBe(
+      "0xd16f93b584fbfdc03a5ee85914a1f29aa35c44fea5144c387ee1040a3c1678252bf323b7e9c3e9b4dfd91cca841fc522f4d3160a1e803f2bf14eb5fa037aae4a1b"
```
adam-alchemy, line 56: Does this depends on the specific value of the private key in the env var? If so, we could use a test key for this purpose (like the output of a string's hash, similar to foundry's [`makeAddrAndKey`](https://github.com/foundry-rs/forge-std/blob/66bf4e2c92cf507531599845e8d5a08cc2e3b5bb/src/StdCheats.sol#L429-L433)).

packages/alchemy/src/chains.ts
dancoombs: Oh shoot, my bad, that's correctly configured in staging but not prod. `BASE_FEE_PERCENTAGE` is the correct value, I'll update that location.

packages/alchemy/src/__tests__/simple-account.test.ts
moldy530: yea it does depend on a specific key in the env var.

I can make that change to test signing, but will still need some live key with funds in it for testing the send user op methods

dancoombs (approved):

PR #32, opened 2023-06-20
fix: support contract deployment by a smart contract account

moldy530:
this supports the use case where a user wants to deploy a contract using a smart contract account

this is done by sending the deploy data to the zeroAddress.

While the sendUserOpeartion method supports this, sendTransaction does not because it does not properly handle undefined target.

So we update the API to support undefined target in the call so that if left undefined (most notably an issue in the ethers adapter when using ContractFactory), we use the zeroAddress


moldy530:
* main
  * **PR #32** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/32" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/32?utm_source=stack-comment).

moldy530:
also added a utility method for deploying contracts when using the core provider

packages/core/src/provider/types.ts
```diff
@@ -128,6 +133,14 @@ export interface ISmartAccountProvider<
    */
   sendTransactions: (request: RpcTransactionRequest[]) => Promise<Hash>;
 
+  /**
+   * This takes contract deploy parameters and crafts a UserOperation to deploy the contract.
```
rthomare, line 137: This also sends the user operation right?

rthomare (approved):

dancoombs (approved):

packages/core/src/provider/types.ts
moldy530: yup! lemme update that

moldy530:
hmmm actually this might not work as expected. The UO is submitted correctly, but the contract doesn't deploy. I think doing `address(0).call(deployData)` doesn't actually work on chain. So need to figure this out. It might be that it's not exactly possible with the current impl of SimpleAccount

moldy530:
Ok yea, closing this out because this won't work. 

What we need to do is one of the two options below which are based on the `deploy` function in `FactoryAssembly` in this example: https://solidity-by-example.org/app/create2/

1. Write a smart contract account that supports a `deploy` method
2. deploy a singleton `GenericAssemblyFactory` like the factory in the example that exposes  `deploy` function that takes in the contract deploy data + salt to pass to CREATE2

PR #39, opened 2023-06-28
Kernel Gas Estimation Fixes

ManaOmlabs:


packages/accounts/src/kernel-zerodev/provider.ts
```diff
@@ -7,40 +7,11 @@ import {
...
 export class KernelAccountProvider extends SmartAccountProvider<HttpTransport> {
-  gasEstimator: AccountMiddlewareFn = async (struct) => {
-    const request = deepHexlify(await resolveProperties(struct));
-    const estimates = await this.rpcClient.estimateUserOperationGas(
-      request,
-      this.entryPointAddress
-    );
-
-    estimates.verificationGasLimit =
-      (BigInt(estimates.verificationGasLimit) * 130n) / 100n;
```
ManaOmlabs, line 18: @moldy530 

I didn't have accurate answer to your question about requiring 30% extra gas usage for kernel. Turns out, it was because i added an inaccurate dummy signature which leads to underestimation of gas. Fixed in this update

ManaOmlabs:
Build will only pass once new version from main branch is released

moldy530:
ah nice good catch! will let you know once I push an update

ManaOmlabs:
@moldy530 ready to review

packages/accounts/src/kernel-zerodev/__tests__/kernel-account.test.ts
```diff
@@ -198,22 +196,31 @@ describe("Kernel Account Tests", () => {
   //     await expect(result).resolves.not.toThrowError();
   // });
 
+  // it("sendUserOperation batch should execute properly", async () => {
```
moldy530, line 199: @ManaOmlabs , the aa infra on alchemy is now open to the public so it should be possible to make these tests pass here. However, @avasisht23 is working on splitting tests into integration vs unit tests so we'll make sure to port these over as integration tests since they require an alchemy api key

packages/accounts/src/kernel-zerodev/provider.ts
```diff
@@ -1,46 +1,12 @@
...
-      }
-      // @ts-ignore
-      return this.account.signWithEip6492(data);
-    } else {
-      return super.request(args);
+  signMessage = async (msg: string | Uint8Array): Promise<Hash> => {
+    if (!this.account) {
+      throw new Error("account not connected!");
     }
+    // @ts-ignore
```
moldy530, line 9: can we make this `@ts-ignore: <reason for ignore>` also we shouldn't use `@ts-ignore` and instead use `@ts-expect-error`

packages/accounts/src/kernel-zerodev/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { KernelUserOperationCallData } from "./types";
+
+export const encodeCall = (call: KernelUserOperationCallData): string => {
+  const data = toBytes(call.data);
+  const encoded = encodePacked(
+    ["uint8", "address", "uint256", "uint256", "bytes"],
+    [
+      call.delegateCall ? 1 : 0,
+      call.target,
+      call.value || BigInt(0),
```
moldy530, line 11: nit:
```suggestion
      call.value ?? BigInt(0),
```

packages/accounts/src/kernel-zerodev/__tests__/kernel-account.test.ts
ManaOmlabs: Sure thing. I will be on lookout for the respective PR

packages/accounts/src/kernel-zerodev/provider.ts
ManaOmlabs: done

packages/accounts/src/kernel-zerodev/utils.ts
ManaOmlabs: done

ManaOmlabs:
@moldy530 let me know if any more changes needed

packages/accounts/src/kernel-zerodev/account.ts
```diff
@@ -67,6 +70,19 @@ export class KernelSmartContractAccount<
     return this.encodeExecuteAction(target, value, data, 1);
   }
 
+  override async encodeBatchExecute(
+    _txs: BatchUserOperationCallData
+  ): Promise<`0x${string}`> {
+    // @ts-ignore
+    const multiSendData: `0x${string}` =
+      "0x" + _txs.map((tx) => encodeCall(tx)).join("");
```
moldy530, line 78: Btw you could use the `concat` util from viem and just do 

```
concat(_txs.map(...))
```

packages/accounts/src/kernel-zerodev/account.ts
```diff
@@ -67,6 +70,19 @@ export class KernelSmartContractAccount<
     return this.encodeExecuteAction(target, value, data, 1);
   }
 
+  override async encodeBatchExecute(
+    _txs: BatchUserOperationCallData
+  ): Promise<`0x${string}`> {
+    // @ts-ignore
```
moldy530, line 76: What's the error here?

packages/accounts/src/kernel-zerodev/provider.ts
```diff
@@ -1,46 +1,12 @@
...
-      }
-      // @ts-ignore
-      return this.account.signWithEip6492(data);
-    } else {
-      return super.request(args);
+  signMessage = async (msg: string | Uint8Array): Promise<Hash> => {
+    if (!this.account) {
+      throw new Error("account not connected!");
     }
+    // @ts-expect-error: signWith6492 exists only on kernel account which is expected
```
moldy530, line 9: Oh you could fix this by either overriding the account param on the kernel provider so it has more specific typing or making the provider take a generic `A extends BaseSmartContractAccount = BaseSmartContractAccount`

moldy530 (approved):
LGTM with some nits and comments on TS errors

Won't be online today much so can take a look again tmr

packages/accounts/src/kernel-zerodev/account.ts
ManaOmlabs: done

packages/accounts/src/kernel-zerodev/account.ts
ManaOmlabs: solved

packages/accounts/src/kernel-zerodev/provider.ts
ManaOmlabs: solved

packages/accounts/src/kernel-zerodev/__tests__/kernel-account.test.ts
ManaOmlabs: Handled

ManaOmlabs:
@moldy530 everything handled :)

PR #42, opened 2023-06-29
fix: local account signers were not correctly handling hex inputs

moldy530:
There was a bug in the hd-account and private key signer. it was incorrectly handling hex inputs, this addresses that bug. 

I also merged the implementations for private-key signer and hd-account signer into one (and marked the old impls as deprecated to avoid breaking peoples stuff for now)


moldy530:
* main
  * **PR #42** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/42" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/42?utm_source=stack-comment).

moldy530:
Test runs below. The one test that failed failed because a tx was still in the mempool. We'll fix this when we split integration tests and unit tests out

image

packages/core/src/__tests__/private-key-signer.test.ts
moldy530, line 1: the tests in this file were moved to `local-account.test.ts`

avasisht23 (approved):

PR #48, opened 2023-07-06
fix: always import with file extension

moldy530:
Adds an eslint config to make sure file extensions are used and fixes the missing ones


rthomare (approved):

dphilipson:
Can you explain why we're doing this? It's just kinda odd looking at first glance: I can't think of a single time I've seen imports with a `.js` extension before.

dphilipson:
Huh, interesting. Thanks!

PR #49, opened 2023-07-09
feat(core): use exponential backoff on waitForUserOperationTransaction

avasisht23:
Requirements:
- update the delay to be an exponential backoff instead of fixed time intervals in `waitForUserOperationTransaction`
- parametrize the backoff / cooldown multiplier in `SmartAccountProviderOpts`
- add tests for `core/src/provider/base.ts`

Tests: see `provider.test.ts`


avasisht23:
* main
  * **PR #49** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/49" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/49?utm_source=stack-comment).

packages/core/src/__tests__/provider.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      await providerMock.waitForUserOperationTransaction("0xTHIS_IS_A_TEST");
+      expect(retryMsDelays).toEqual([2_000]);
+      expect(getUserOperationReceiptMock).toHaveBeenCalledTimes(1);
+    });
+
+    it("should retry twice with exponential delay for waitForUserOperationTransaction", async () => {
+      const providerMock = new SmartAccountProvider(
+        "ALCHEMY_RPC_URL",
+        "0xENTRYPOINT_ADDRESS",
+        polygonMumbai
```
avasisht23, line 66: wondering how to mock this?

packages/core/src/__tests__/provider.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+describe(
+  "Base Tests",
+  () => {
+    let retryMsDelays: number[] = [];
+
+    beforeEach(() => {
+      vi.useFakeTimers();
+      vi.spyOn(global, "setTimeout").mockImplementation(
+        // @ts-ignore: Mock implementation doesn't need to return a Timeout.
```
avasisht23, line 15: copied from: https://github.com/alchemyplatform/alchemy-sdk-js/blob/b4df3768/test/unit/backoff.test.ts#L14

packages/core/src/__tests__/provider.test.ts
```diff
@@ -0,0 +1,136 @@
...
+        "0xENTRYPOINT_ADDRESS",
+        polygonMumbai
+      );
+
+      const getUserOperationReceiptMock = vi
+        .spyOn(providerMock, "getUserOperationReceipt")
+        .mockImplementation(() => {
+          return Promise.resolve({
+            receipt: { transactionHash: "0xMOCK_USER_OP_RECEIPT" },
+          } as unknown as UserOperationReceipt);
```
avasisht23, line 42: I casted to `unknown` to avoid ballooning the mocked return value with irrelevant fields for the tested code path. lmk if there's a better way!

packages/core/src/__tests__/provider.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      const getUserOperationReceiptMock = vi
+        .spyOn(providerMock, "getUserOperationReceipt")
+        .mockImplementationOnce(() =>
+          Promise.reject("Failed request, must retry")
+        )
+        .mockImplementation(() => {
+          return Promise.resolve({
+            receipt: { transactionHash: "0xMOCK_USER_OP_RECEIPT" },
+          } as unknown as UserOperationReceipt);
+        });
```
moldy530, line 78: we could probably break this out into a `givenGetUserOperationFailsNTimes(n)` function which we can reuse throughout the tests and do something like

```ts
const givenGetUserOperationFailsNTimes = (times: number) => {
    let retries = times;
    const mock = vi.spyOn(providerMock, "getUserOperationReceipt")
        .mockImplementation(() => {
            if (retries > 0) return Promise.reject("Failed request, must retry");

            return Promise.resolve({
                receipt: { transactionHash: "0xMOCK_USER_OP_RECEIPT" },
            } as unknown as UserOperationReceipt);    
        });
    // can do the assert here if it's needed
    return mock;
}
```

(i usually follow naming and structuring my tests based on https://martinfowler.com/bliki/GivenWhenThen.html)

packages/core/src/__tests__/provider.test.ts
```diff
@@ -0,0 +1,136 @@
...
+        .mockImplementation(() => {
+          return Promise.resolve({
+            hash: "0xMOCK_TXN_HASH",
+          } as unknown as Transaction);
+        });
+      expect(getTransactionMock.getMockName()).toEqual("getTransaction");
+
+      await providerMock.waitForUserOperationTransaction("0xTHIS_IS_A_TEST");
+      expect(retryMsDelays).toEqual([2_000, 3_000]);
+      expect(getUserOperationReceiptMock).toHaveBeenCalledTimes(2);
```
moldy530, line 94: going off the above comment, could have a shared function for this too that's like `thenExpectRetriesToBe(expectedRetryMsDelays, expectedMockCalls, getUserOperationReceiptMock)`

packages/core/src/provider/base.ts
```diff
@@ -181,16 +191,18 @@ export class SmartAccountProvider<
   waitForUserOperationTransaction = async (hash: Hash): Promise<Hash> => {
     for (let i = 0; i < this.txMaxRetries; i++) {
       await new Promise((resolve) =>
-        setTimeout(resolve, this.txRetryIntervalMs)
+        setTimeout(
+          resolve,
+          this.txRetryIntervalMs * Math.pow(this.txRetryMulitplier, i)
```
moldy530, line 196: nit: usually for exponential retries, there's also random jitter added to the final value here -- usually a random number in the hundreds of milliseconds

logic is that if the reason these retries are happening because we went down, we don't want our clients to keep us down by causing a thundering herd scenario. So you randomly stagger each of the clients to try to avoid all of them making calls at the same time

packages/core/src/__tests__/provider.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      const getUserOperationReceiptMock = vi
+        .spyOn(providerMock, "getUserOperationReceipt")
+        .mockImplementation(() => {
+          return Promise.resolve({
+            receipt: { transactionHash: "0xMOCK_USER_OP_RECEIPT" },
+          } as unknown as UserOperationReceipt);
+        });
+      expect(getUserOperationReceiptMock.getMockName()).toEqual(
+        "getUserOperationReceipt"
+      );
```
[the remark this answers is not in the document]
avasisht23: No cases where it changes. Just saw it as an example way to test and kept it in, but it's superfluous - will remove! J

packages/core/src/__tests__/provider.test.ts
avasisht23: sorry I guess for "this" I meant the `polygonMumbai` parameter

packages/core/src/__tests__/provider.test.ts
avasisht23: tweaked this a bit b/c logic is slightly than what I'm doing, but this a helpful framing for me going forward!

moldy530 (approved):

PR #54, opened 2023-07-11
refactor: rearrange tests

avasisht23:
Requirements:
- for unit tests, have a subfolder called `__tests__` which contains tests for relevant files in a folder
-- e.g. `provider/__test__/base.test.ts` if `provider/base.ts` has testable classes/methods etc.
- for e2e/integration tests, have a folder called `e2e-tests` at the root directory of each package which contains e2e testing for that package

Tests (offline to show unit tests work - failures are e2e, due to underfunding including `sendUserOperation`):
image
image

Tests (online - the only ones that fail are due to lack of funds in wallet / test timeout):
image
image


avasisht23:
* main
  * **PR #54** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/54" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/54?utm_source=stack-comment).

packages/core/src/__tests__/local-account.test.ts
moldy530, line 1: as we add more tests I think it's gonna start to get really crowded in the root folder. I would either have these be in `signer/__tests__` or just leave as is

packages/accounts/src/kernel-zerodev/test_e2e/kernel-account.test.ts
moldy530, line 1: I think the way these tests are written they can run in a __tests__ folder as part of PRs because they don't depend on a specific wallet or API key

packages/alchemy/test_e2e/simple-account.test.ts
moldy530, line 1: not all the tests in this file are necessarily e2e tests. I think the only ones are the e2e are the ones that try to execute a UO successfully

packages/core/test_e2e/simple-account.test.ts
moldy530, line 1: same here

packages/ethers/test_e2e/simple-account.test.ts
moldy530, line 1: same here

packages/alchemy/test_e2e/simple-account.test.ts
avasisht23: I was able to remove a few I consider to be unit tests (i.e. able to run offline, not read/writing from an external service like blockchain SCWs). 

But, that means I considered more as-written tests as e2e than just the ones sending an UO - i.e. `estimateUserOperationGas`, `sendUserOperation`, `eth_call` for "getCounterFactualAddress." I think is the right split. Lmk what you think.

packages/accounts/src/kernel-zerodev/test_e2e/kernel-account.test.ts
avasisht23: there were some I was able to remove a few I consider to be unit tests (i.e. able to run offline, not read/writing from an external service like blockchain SCWs). 

But, that means I considered more as-written tests as e2e than just the ones sending an UO / dependent on a specific wallet or API key.

packages/accounts/src/kernel-zerodev/e2e-tests/mocks/mock-signer.ts
```diff
@@ -0,0 +1,13 @@
+import type { Address, Hex, SmartAccountSigner } from "@alchemy/aa-core";
```
avasisht23, line 1: mocked like how zerodev did, in order to avoid using .env variables like owner nemonic for unit tests so devs can run the unit tests locally. I couldn't use a dummy nemonic like in local-account because the checksum invalid with the ethers library.

packages/ethers/src/__tests__/mocks/mock-signer.ts
```diff
@@ -0,0 +1,13 @@
+import type { Address, Hex, SmartAccountSigner } from "@alchemy/aa-core";
+
+export class MockSigner implements SmartAccountSigner {
```
avasisht23, line 3: mocked like how zerodev did, in order to avoid using .env variables like owner nemonic for unit tests so devs can run the unit tests locally. I couldn't use a dummy nemonic like in local-account because the checksum invalid with the ethers library.

packages/accounts/src/kernel-zerodev/e2e-tests/kernel-account.test.ts
```diff
@@ -95,32 +95,6 @@ describe("Kernel Account Tests", () => {
...
-        "0xA7b2c01A5AfBCf1FAB17aCf95D8367eCcFeEb845",
-        1n,
-        "0x234"
-      )
-    ).eql(
-      "0x51945447000000000000000000000000a7b2c01a5afbcf1fab17acf95d8367eccfeeb84500000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000022340000000000000000000000000000000000000000000000000000000000000"
-    );
-  });
-
   it("signWithEip6492 should correctly sign the message", async () => {
```
moldy530, line 98: this one looks like a unit test

packages/alchemy/src/__tests__/provider.test.ts
moldy530, line 1: nit on the file naming here: this is a `simple-account.test`

packages/core/e2e-tests/simple-account.test.ts
moldy530, line 1: nit: this still contains tests for getting valid signature and counterfactual, which are unit tests

packages/ethers/src/__tests__/mocks/mock-signer.ts
moldy530: For unit tests, I think we should use some kind of hard coded mnemonic or private key like https://github.com/wagmi-dev/viem/blob/main/src/_test/constants.ts#L6C8-L6C74 

The problem with hard coding the `signMessage` here is that it won't catch issues with the underlying signer. The sign message tests are actually there because at some point viem wasn't signing `hash` messages correctly so I think we should always use the underlying signer type for that

moldy530 (commented):
couple nits about further breaking apart tests

moldy530 (approved):

moldy530 (approved):

avasisht23:
@avasisht23 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/54).

avasisht23:
@avasisht23 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/54).

PR #56, opened 2023-07-12
ci: add unit tests to on-pull-request action

avasisht23:
Requirements
- add unit tests to on-pull-request action
- enforce running on-pull-request on PRs merging to any branch, to account for stacked PRs

Tests: see output of CI below (which passes)

---

## PR-Codex overview
This PR focuses on updating the GitHub Actions workflow file `on-pull-request.yml`. 

### Detailed summary:
- Renamed the job `build_and_lint` to `build_and_lint_and_test`
- Added a new job `Unit Test` to run the `yarn test` command

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #55, opened 2023-07-12
chore: separate e2e and integration test scripts

avasisht23:
Requirement: separate e2e and integration test scripts.

Tests:
- Unit (15 tests)
image

- E2E (26 tests)
image

---

## PR-Codex overview
The focus of this PR is to add end-to-end testing configurations for different packages. 

### Detailed summary:
- Added `vitest.config.e2e.ts` files for `core`, `ethers`, `accounts`, and `alchemy` packages.
- Imported `configDefaults` and `defineProject` from `vitest/config` in the new configuration files.
- Added `test:run-e2e` scripts in the `package.json` files for running end-to-end tests.
- Excluded `**/e2e-tests/**/*.test.ts` files in the new configuration files.
- Excluded `**/__tests__/**/*.test.ts` files in the new configuration files.
- Updated the `test` script in the root `package.json` to run `lerna run test:run`.
- Added `test:e2e` script in the root `package.json` to run `lerna run test:run-e2e`.

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #57, opened 2023-07-12
ci: add e2e tests to publish-package action

avasisht23:
Requirements:
- add env variables for secrets to run e2e tests
- run e2e tests as part of publish-package action
- add flag to only deploy if e2e tests pass
- add publish flag for publish-package action

Tests:
- see this successful CI run for publish flag true (would publish but prints the flags both being true instead as test)  https://github.com/alchemyplatform/aa-sdk/actions/runs/5628745012/job/15252957921

- see this successful CI run for publish flag false (skips deploy because flag is false) https://github.com/alchemyplatform/aa-sdk/actions/runs/5628721929/job/15252905766

- see this correctly unsuccessful CI run for failed e2e tests (skips deploy because e2e failed even if publish flag is true) https://github.com/alchemyplatform/aa-sdk/actions/runs/5617640123

---

## PR-Codex overview
This PR focuses on enhancing the test:e2e script and adding a workflow_dispatch event to trigger the workflow manually. 

### Detailed summary
- Updated the test:e2e script to include the "--concurrency=1" flag
- Added a workflow_dispatch event with an input parameter for publishing a new package

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -1,11 +1,11 @@
 name: Publish new Package
 
 # workflow dispatch requires a maintainer to go to the Actions tab and manually trigger the workflow
-on: 
+on:
   workflow_dispatch:
     inputs:
       publish:
-        description: 'Flag to publish new package'
+        description: "Flag to publish new package"
         required: true
         default: false
         type: boolean
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -55,15 +55,15 @@ jobs:
 
       - name: Artifact of E2E tests
         id: artifact
-        if: steps.e2e.outcome == "success"
-        run: echo "e2e=true" >> "$GITHUB_OUTPUT"
+        if: steps.e2e.outcome == 'success'
+        run: echo 'e2e=true' >> "$GITHUB_OUTPUT"
 
   # Lerna publish will, compute the new version, update files, run precommit hooks, tag, publish, and push change log
   deploy:
     name: Deploy
     runs-on: ubuntu-latest
     needs: build_and_test
-    if: inputs.publish && (needs.build_and_test.outputs.artifact == "true")
+    if: inputs.publish && (needs.build_and_test.outputs.artifact == 'true')
     steps:
       - name: Publish using Lerna
         # run: yarn lerna publish --no-private --yes --no-verify-access
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -55,7 +55,7 @@ jobs:
 
       - name: Artifact of E2E tests
         id: artifact
-        if: steps.e2e.outcome == 'success'
+        if: steps.e2e.outcome == "success"
         run: echo "e2e=true" >> "$GITHUB_OUTPUT"
 
   # Lerna publish will, compute the new version, update files, run precommit hooks, tag, publish, and push change log
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -36,8 +36,6 @@ jobs:
       - name: Build
         run: yarn build
 
-      # TODO: we still need to make the tests work in CI environments
-      # will come back to this
       - name: E2E Test
         env:
           API_KEY: ${{ secrets.API_KEY }}
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -17,7 +17,7 @@ jobs:
     name: Build and Test
     runs-on: ubuntu-latest
     outputs:
-      e2e: ${{ steps.artifact.outputs.e2e }}
+      artifact: ${{ steps.artifact.outputs.e2e }}
     steps:
       - uses: actions/checkout@v3
         with:
@@ -46,6 +46,7 @@ jobs:
         run: yarn build
 
       - name: E2E tests
+        id: e2e
         env:
           API_KEY: ${{ secrets.API_KEY }}
           OWNER_MNEMONIC: ${{ secrets.OWNER_MNEMONIC }}
@@ -62,8 +63,8 @@ jobs:
     name: Deploy
     runs-on: ubuntu-latest
     needs: build_and_test
-    if: ${{ inputs.publish }} && ${{ needs.build_and_test.outputs.e2e }}
+    if: ${{ inputs.publish }} && ${{ needs.build_and_test.outputs.artifact }}
     steps:
       - name: Publish using Lerna
         # run: yarn lerna publish --no-private --yes --no-verify-access
-        run: echo "${{ inputs.publish }} ${{ needs.build_and_test.outputs.e2e }}"
+        run: echo "${{ inputs.publish }} ${{ needs.build_and_test.outputs.artifact }}"
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -44,5 +44,5 @@ jobs:
         run: yarn test:e2e
 
       # Lerna publish will, compute the new version, update files, run precommit hooks, tag, publish, and push change log
-      - name: Publish using Lerna
-        run: yarn lerna publish --no-private --yes --no-verify-access
+      # - name: Publish using Lerna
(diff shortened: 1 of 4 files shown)


avasisht23:
* main
  * **PR #54** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/54" target="_blank">Graphite</a> 
    * **PR #55** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/55" target="_blank">Graphite</a> 
      * **PR #56** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/56" target="_blank">Graphite</a> 
        * **PR #57** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/57" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/57?utm_source=stack-comment).

avasisht23:
* main
  * **PR #54** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/54" target="_blank">Graphite</a> 
    * **PR #55** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/55" target="_blank">Graphite</a> 
      * **PR #56** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/56" target="_blank">Graphite</a> 
        * **PR #57** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/57" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/57?utm_source=stack-comment).

moldy530 (commented):
can we add configuration params to this workflow? Did your test actually publish something?

EDIT: oh I see that it didn't. ok cool. Let's add the config params to this still so that we can leverage it in the future (probably pretty useful cuz we could theoretically run it manually on branches)

avasisht23:
> can we add configuration params to this workflow? Did your test actually publish something?
> 
> EDIT: oh I see that it didn't. ok cool. Let's add the config params to this still so that we can leverage it in the future (probably pretty useful cuz we could theoretically run it manually on branches)

ah I should have clarified: I looked into config + conditional. Conditionals only work at the job level, not the job-step level. We'd need to refactor the publish action to have publish be a separate job after the install and build and test job. That cool?

.github/workflows/publish-package.yml
```diff
@@ -36,11 +45,25 @@ jobs:
...
+        if: steps.e2e.outcome == 'success'
+        run: echo 'e2e=true' >> "$GITHUB_OUTPUT"
 
-      # Lerna publish will, compute the new version, update files, run precommit hooks, tag, publish, and push change log
+  # Lerna publish will, compute the new version, update files, run precommit hooks, tag, publish, and push change log
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+    needs: build_and_test
+    if: inputs.publish && (needs.build_and_test.outputs.artifact == 'true')
```
avasisht23, line 66: from docs: "you may omit the ${{ }} expression syntax because GitHub Actions automatically evaluates the if conditional as an expression. Using the ${{ }} expression syntax turns the contents into a string, and strings are truthy."

I also check against the stringified boolean because the artifact variable `e2e` I believe is still a string.

.github/workflows/publish-package.yml
```diff
@@ -36,11 +45,25 @@ jobs:
...
+        id: e2e
+        env:
+          API_KEY: ${{ secrets.API_KEY }}
+          OWNER_MNEMONIC: ${{ secrets.OWNER_MNEMONIC }}
+          PAYMASTER_POLICY_ID: ${{ secrets.PAYMASTER_POLICY_ID }}
+        run: yarn test:e2e
+
+      - name: Artifact of E2E tests
+        id: artifact
+        if: steps.e2e.outcome == 'success'
```
avasisht23, line 58: from [here](https://stackoverflow.com/a/70549615) on how to get the result of the e2e step. referenced by id.

.github/workflows/publish-package.yml
```diff
@@ -36,11 +45,25 @@ jobs:
...
+        run: echo 'e2e=true' >> "$GITHUB_OUTPUT"
```
avasisht23, line 59: from example in [docs](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs) on saving the "build and test" job's e2e step's output to condition whether to run the next job "deploy"

.github/workflows/publish-package.yml
moldy530: is this necessary, I thought this is what `needs` is for below? like if one of these things fails causing `build_and_test` to fail, then the needs constraint isn't met? I could be wrong actually because it's been a minute since I've used needs

moldy530: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-requiring-successful-dependent-jobs yea I think you should be fine with just needs

.github/workflows/publish-package.yml
```diff
@@ -36,11 +45,25 @@ jobs:
...
 
-      # Lerna publish will, compute the new version, update files, run precommit hooks, tag, publish, and push change log
+  # Lerna publish will, compute the new version, update files, run precommit hooks, tag, publish, and push change log
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+    needs: build_and_test
+    if: inputs.publish && (needs.build_and_test.outputs.artifact == 'true')
+    steps:
```
moldy530, line 67: does this not need checkout, install, and build? I think it does because the artifacts don't transfer to the next job?

.github/workflows/publish-package.yml
avasisht23: i just realized i can just do the conditionals on a job-step, not just jobs. so just going to consolidate deploy into the build_and_test job to become build_and_test_and_deploy, conditional on the publish flag and successful e2e. 

this avoids duplicating checkout, install, and build (per your below comment)

.github/workflows/publish-package.yml
avasisht23: i just realized i can just do the conditionals on a job-step, not just jobs. so just going to consolidate deploy into the build_and_test job to become build_and_test_and_deploy, conditional on the publish flag and successful e2e. 

this avoids duplicating checkout, install, and build.

.github/workflows/publish-package.yml
```diff
@@ -36,11 +43,15 @@ jobs:
...
+        id: e2e
+        env:
+          API_KEY: ${{ secrets.API_KEY }}
+          OWNER_MNEMONIC: ${{ secrets.OWNER_MNEMONIC }}
+          PAYMASTER_POLICY_ID: ${{ secrets.PAYMASTER_POLICY_ID }}
+        run: yarn test:e2e
 
       # Lerna publish will, compute the new version, update files, run precommit hooks, tag, publish, and push change log
       - name: Publish using Lerna
+        if: inputs.publish && steps.e2e.outcome == 'success'
```
avasisht23, line 56: i just realized i can just do the conditionals on a job-step, not just jobs. so just going to consolidate deploy into the build_and_test job to become build_and_test_and_deploy, conditional on the publish flag and successful e2e.

this avoids duplicating checkout, install, and build.

moldy530 (approved):

avasisht23:
@avasisht23 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/57).

avasisht23:
[Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/57) couldn't merge this pull request because a downstack PR #60 failed to merge.

moldy530:
@moldy530 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/57).

moldy530:
@moldy530 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/57).

PR #61, opened 2023-07-27
feat: support passing fee overrides to the send user op method

moldy530:
This adds the ability to override the gas fees of a user op

useful if you need to replace a user operation

---

### Detailed summary
- Added a new type `UserOperationOverrides` to represent gas price overrides in user operation requests.
- Modified the `sendUserOperation` method to accept an optional `overrides` parameter.
- Updated the `sendUserOperation` method to apply the overrides if they are passed in.
- Updated the `SmartAccountProvider` class to use the `overrides` parameter when calling `sendUserOperation` method.

### DIFF
--- a/packages/core/src/provider/base.ts
+++ b/packages/core/src/provider/base.ts
@@ -4,8 +4,8 @@ import {
   type Chain,
   type Hash,
   type RpcTransactionRequest,
-  type Transport,
   type Transaction,
+  type Transport,
 } from "viem";
 import { arbitrum, arbitrumGoerli } from "viem/chains";
 import { BaseSmartContractAccount } from "../account/base.js";
@@ -18,6 +18,7 @@ import {
   isValidRequest,
   type BatchUserOperationCallData,
   type UserOperationCallData,
+  type UserOperationOverrides,
   type UserOperationReceipt,
   type UserOperationResponse,
   type UserOperationStruct,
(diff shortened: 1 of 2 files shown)


moldy530:
* main
  * **PR #61** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/61" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/61?utm_source=stack-comment).

packages/core/src/provider/base.ts
```diff
@@ -235,7 +264,9 @@ export class SmartAccountProvider<
       this.gasEstimator,
       this.feeDataGetter,
       this.paymasterDataMiddleware,
-      this.customMiddleware ?? noOpMiddleware
+      this.customMiddleware ?? noOpMiddleware,
+      // This applies the overrides if they've been passed in
+      async (struct) => ({ ...struct, ...overrides })
```
moldy530, line 269: note: the reason we don't short circuit the feeDataGetter if the override is passed in is because it's possible to pass in only a partial override

packages/core/src/provider/base.ts
```diff
@@ -175,15 +181,37 @@ export class SmartAccountProvider<
...
       };
     });
 
-    const { hash } = await this.sendUserOperation(batch);
+    const bigIntMax = (...args: bigint[]) => {
+      if (!args.length) {
+        return undefined;
+      }
+
+      return args.reduce((m, c) => (m > c ? m : c));
```
avasisht23, line 196: I believe you should have an "initialValue"? [(docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce). Perhaps 0?

packages/core/src/provider/base.ts
avasisht23: why is it async? do we also need to await if so?

avasisht23 (approved):
lgtm! few qs, non-blocking.

packages/core/src/provider/base.ts
moldy530: in the case of initialValue not being supplied it uses `args[0]` as the first value and starts with `c` as `args[1]`

packages/core/src/provider/base.ts
moldy530: It's async because the signature of the pipeAsync requires the functions to be of type `T => Promise<T>` which is what this is doing. Making it `async` implies that the response is wrapped in promise

PR #71, opened 2023-08-08
feat(alchemy): use rundler_getLocalRequiredFees for fee estimate

dancoombs:
### Detailed summary
- This PR adds the `RPC_URL` and `API_KEY` constants to several test files.
- It removes the `GasFeeStrategy` enum and `GasFeeMode` interface from the `gas-fees.ts` file.
- It updates the import statements in some test files to include the `RPC_URL` constant.
- It updates the `withAlchemyGasFeeEstimator` function in the `gas-fees.ts` file to remove the `feeMode` and `maxPriorityFeeBufferPercent` parameters.

> The following files were skipped due to too many changes: `packages/alchemy/src/middleware/gas-fees.ts`

### DIFF
--- a/packages/accounts/src/kernel-zerodev/e2e-tests/kernel-account.test.ts
+++ b/packages/accounts/src/kernel-zerodev/e2e-tests/kernel-account.test.ts
@@ -19,14 +19,14 @@ import {
 import { KernelAccountProvider } from "../provider.js";
 import type { KernelUserOperationCallData } from "../types.js";
 import { KernelBaseValidator, ValidatorMode } from "../validator/base.js";
-import { API_KEY, OWNER_MNEMONIC } from "./constants.js";
+import { RPC_URL, API_KEY, OWNER_MNEMONIC } from "./constants.js";
 import { MockSigner } from "./mocks/mock-signer.js";
 
 describe("Kernel Account Tests", () => {
   //any wallet should work
   const config = {
     chain: polygonMumbai,
-    rpcProvider: `${polygonMumbai.rpcUrls.alchemy.http[0]}/${API_KEY}`,
+    rpcProvider: RPC_URL !== undefined ? RPC_URL : `${polygonMumbai.rpcUrls.alchemy.http[0]}/${API_KEY}`,
     validatorAddress: "0x180D6465F921C7E0DEA0040107D342c87455fFF5" as Address,
     accountFactoryAddress:
       "0x5D006d3880645ec6e254E18C1F879DAC9Dd71A39" as Address,
--- a/packages/alchemy/src/middleware/gas-fees.ts
+++ b/packages/alchemy/src/middleware/gas-fees.ts
@@ -1,55 +1,17 @@
 import type { AlchemyProvider } from "../provider.js";
-
-export enum GasFeeStrategy {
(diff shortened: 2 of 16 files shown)


packages/core/src/provider/base.ts
```diff
@@ -293,8 +293,8 @@ export class SmartAccountProvider<
     const initCode = await this.account.getInitCode();
     const uoStruct = await asyncPipe(
       this.dummyPaymasterDataMiddleware,
-      this.gasEstimator,
       this.feeDataGetter,
+      this.gasEstimator,
```
dancoombs, line 297: @moldy530 can you check my logic on this change. We want the fees to get estimated and applied before gas is estimated as gas estimation can sometimes rely on fees (e.g. as it will on Optimism). Does this have that intended impact?

packages/alchemy/src/middleware/gas-fees.ts
```diff
@@ -1,55 +1,17 @@
...
-          throw new Error("fee mode not supported");
-      }
-    })();
-
-    return {
-      maxPriorityFeePerGas: prioFee,
-      maxFeePerGas: baseFeeScaled + prioFee,
-    };
+    const result = await (
+      provider.rpcClient as ClientWithAlchemyMethods
```
moldy530, line 9: I think we should update the typing on the `AlchemyProvider` to be this so we don't have to typecast

packages/core/src/provider/base.ts
moldy530: yup! pipe executes sequentially so this will do fees then gas

packages/accounts/src/kernel-zerodev/e2e-tests/kernel-account.test.ts
```diff
@@ -19,14 +19,14 @@ import {
...
-import { API_KEY, OWNER_MNEMONIC } from "./constants.js";
+import { RPC_URL, API_KEY, OWNER_MNEMONIC } from "./constants.js";
 import { MockSigner } from "./mocks/mock-signer.js";
 
 describe("Kernel Account Tests", () => {
   //any wallet should work
   const config = {
     chain: polygonMumbai,
-    rpcProvider: `${polygonMumbai.rpcUrls.alchemy.http[0]}/${API_KEY}`,
+    rpcProvider: RPC_URL !== undefined ? RPC_URL : `${polygonMumbai.rpcUrls.alchemy.http[0]}/${API_KEY}`,
```
moldy530, line 29: what's the use case for this? 

also nit:
```suggestion
    rpcProvider: RPC_URL != null ? RPC_URL : `${polygonMumbai.rpcUrls.alchemy.http[0]}/${API_KEY}`,
```

dancoombs: To be able to run the e2e tests against something other than Alchemy prod

dancoombs: Can put this in a separate PR, just needed it for my current testing.

moldy530: all good, we can keep it here

PR #74, opened 2023-08-09
feat(alchemy): use rundler_maxPriorityFeePerGas for fee estimate

dancoombs:
Use the new `rundler_maxPriorityFeePerGas` instead of hardcoded fee modes during Alchemy provider fee estimation.

---

### Focus of the PR:
This PR focuses on refactoring and improving the gas fee handling and estimation logic in the Alchemy package.

### Detailed summary:
- The `GasFeeStrategy` enum and `GasFeeMode` interface have been removed from the `gas-fees.ts` file.
- The `GasFeeStrategy` enum and `GasFeeMode` interface have been removed from the `OnboardingController.ts` file.
- The `withAlchemyGasFeeEstimator` function in the `gas-fees.ts` file has been updated to accept `baseFeeBufferPercent` and `maxPriorityFeeBufferPercent` as parameters instead of `feeMode`.
- The `withAlchemyGasFeeEstimator` function in the `gas-fees.ts` file now uses the `alchemyClient.request` method to fetch the `maxPriorityFeePerGas` value.
- The `withAlchemyGasFeeEstimator` function in the `gas-fees.ts` file now calculates the `baseFeeIncrease` and `prioFeeIncrease` values based on the `baseFeeBufferPercent` and `maxPriorityFeeBufferPercent` parameters.
- The `withAlchemyGasManager` function in the `gas-manager.ts` file now uses the `provider.rpcClient` property instead of `config.provider` to access the RPC client.
- The `withAlchemyGasManager` function in the `gas-manager.ts` file now accepts `baseFeeBufferPercent` and `maxPriorityFeeBufferPercent` as parameters instead of `provider`.
- The `alchemyPaymasterAndDataMiddleware` function in the `gas-manager.ts` file now uses the `provider.rpcClient` property instead of `config.provider` to access the RPC client.

> The following files were skipped due to too many changes: `packages/alchemy/src/middleware/gas-manager.ts`, `packages/alchemy/src/provider.ts`

### DIFF
--- a/packages/alchemy/e2e-tests/simple-account.test.ts
+++ b/packages/alchemy/e2e-tests/simple-account.test.ts
@@ -88,10 +88,66 @@ describe("Simple Account Tests", () => {
 
   it("should successfully execute with alchemy paymaster info", async () => {
     const newSigner = signer.withAlchemyGasManager({
-      provider: signer.rpcClient,
       policyId: PAYMASTER_POLICY_ID,
       entryPoint: ENTRYPOINT_ADDRESS,
     });
+
+    const result = await newSigner.sendUserOperation({
+      target: await newSigner.getAddress(),
+      data: "0x",
+    });
+    const txnHash = signer.waitForUserOperationTransaction(result.hash as Hash);
+
+    await expect(txnHash).resolves.not.toThrowError();
+  }, 50000);
+
+  it("should successfully override fees in alchemy paymaster", async () => {
+    const newSigner = signer
+      .withAlchemyGasManager({
+        policyId: PAYMASTER_POLICY_ID,
+        entryPoint: ENTRYPOINT_ADDRESS,
+      })
+      .withFeeDataGetter(async () => ({
+        maxFeePerGas: 1n,
+        maxPriorityFeePerGas: 1n,
+      }));
+
+    // this should fail since we set super low fees
+    await expect(
+      async () =>
+        await newSigner.sendUserOperation({
+          target: await newSigner.getAddress(),
+          data: "0x",
+        })
+    ).rejects.toThrow();
+  }, 50000);
+
+  it("should successfully use paymaster with fee opts", async () => {
+    const newSigner = new AlchemyProvider({
+      apiKey: API_KEY,
+      rpcUrl: RPC_URL,
+      chain,
+      entryPointAddress: ENTRYPOINT_ADDRESS,
+      feeOpts: {
+        baseFeeBufferPercent: 50n,
+        maxPriorityFeeBufferPercent: 50n,
+        preVerificationGasBufferPercent: 50n,
+      },
+    })
+      .connect(
+        (provider) =>
+          new SimpleSmartContractAccount({
+            entryPointAddress: ENTRYPOINT_ADDRESS,
+            chain,
+            owner,
+            factoryAddress: SIMPLE_ACCOUNT_FACTORY_ADDRESS,
+            rpcClient: provider,
+          })
+      )
+      .withAlchemyGasManager({
+        policyId: PAYMASTER_POLICY_ID,
+        entryPoint: ENTRYPOINT_ADDRESS,
+      });
 
     const result = await newSigner.sendUserOperation({
       target: await newSigner.getAddress(),
--- a/packages/alchemy/src/middleware/gas-fees.ts
+++ b/packages/alchemy/src/middleware/gas-fees.ts
@@ -1,54 +1,31 @@
 import type { AlchemyProvider } from "../provider.js";
 
-export enum GasFeeStrategy {
-  DEFAULT = "DEFAULT",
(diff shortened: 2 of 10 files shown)


moldy530 (dismissed):

moldy530 (dismissed):

packages/alchemy/e2e-tests/simple-account.test.ts
```diff
@@ -101,4 +100,60 @@ describe("Simple Account Tests", () => {
...
+
+  it("should successfully override fees in alchemy paymaster", async () => {
+    const newSigner = signer
+      .withAlchemyGasManager({
+        policyId: PAYMASTER_POLICY_ID,
+        entryPoint: ENTRYPOINT_ADDRESS,
+      })
+      .withFeeDataGetter(async () => ({
+        maxFeePerGas: 1n,
+        maxPriorityFeePerGas: 1n,
```
dancoombs, line 112: TODO: replace these once tested.

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -2,49 +2,14 @@ import {
...
-      }>;
-    }["request"];
-};
+import type { Address, Transport } from "viem";
+import type { ClientWithAlchemyMethods } from "./client.js";
 
 export interface AlchemyGasManagerConfig {
   policyId: string;
   entryPoint: Address;
-  provider: PublicErc4337Client;
```
dancoombs, line 47: Removing this allows chaining the `.connect()` call with `.withAlchemyGasManager()` without having to pass the rpc client in twice. The RPC client is passed through the `AlchemyProvider`.

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -70,23 +35,40 @@ export const withAlchemyGasManager = <
...
         maxFeePerGas: 0n,
         maxPriorityFeePerGas: 0n,
       }))
       .withPaymasterMiddleware({
         paymasterDataMiddleware: async (struct) => {
+          const userOperation: UserOperationRequest = deepHexlify(
+            await resolveProperties(struct)
+          );
+
+          var feeOverride = undefined;
```
moldy530, line 50: use `let` if you want mutable or `const` if immutable. `var` is no bueno because it get's hoisted into global scope lol

packages/alchemy/src/middleware/gas-fees.ts
```diff
@@ -1,54 +1,31 @@
 import type { AlchemyProvider } from "../provider.js";
 
-export enum GasFeeStrategy {
```
rthomare, line 3: good kill!

packages/alchemy/src/provider.ts
```diff
@@ -107,6 +131,12 @@ export class AlchemyProvider extends SmartAccountProvider<HttpTransport> {
       );
     }
 
-    return withAlchemyGasManager(this, config);
+    if (this.feeOptsSet) {
```
rthomare, line 134: I worry about this a little bit. It'd be nice to just fix the withAlchemyGasManager to not override right? 

Specifically by not calling the code block below conditionally in `withAlchemyGasManager` if the fee ops are set in the config? 
```
  .withFeeDataGetter(async () => ({
     maxFeePerGas: 0n,
     maxPriorityFeePerGas: 0n,
  }))
```

rthomare: curious on @moldy530's thoughts

moldy530: oh wait... I have a thing for this already https://github.com/alchemyplatform/aa-sdk/blob/main/packages/core/src/provider/base.ts#L311

you could potentially just override the sendUserOperation method here to do something like:

```ts
sendUserOperation = async (
    data: UserOperationCallData | BatchUserOperationCallData,
    overrides?: UserOperationOverrides
): Promise<SendUserOperationResult> => {
  super.sendUserOperation(data, overrides ?? this.overrides)
}
```

dancoombs: You can't override the fees after the paymaster has signed them

dancoombs: This code switches between the two middleware functions based on if the fee opts are set. If no fees are set we should use the middleware where we make gas estimation a no-op.

Not really sure I see another way to do this? The issue isn't really the fee estimation logic, its the removal of the gas estimation logic.

We could also remove the usage of `alchemy_requestGasAndPaymasterAndData` entirely from this code, that would keep this cleaner (but we would lose test coverage there).

moldy530: gotcha yea so requestGasAndPaymaster data overrides all both the fee estimates and gas estimates performed. So if you have any overrides set, then that would mean that those don't get respected if you want to use a paymaster which by default gets gas + fee estimates. 

I'm good with the approach here, it makes sense.

moldy530 (approved):

PR #81, opened 2023-08-16
feat: add jwt support for Alchemy providers

avasisht23:
### Focus of the PR:
This PR focuses on adding a JWT (JSON Web Token) property to the `AlchemyProviderConfig` and `ConnectionConfig` types in order to support authentication with Alchemy API.

### Detailed summary:
- Added `ConnectionConfig` type to support different connection configurations.
- Added `jwt` property to the `ConnectionConfig` type to store the JSON Web Token.
- Updated `AlchemyProviderConfig` type to include `jwt` property.
- Updated `AlchemyProvider` constructor to accept `jwt` property in the connection configuration.
- Added a test to ensure the JWT property is correctly passed to the `createPublicErc4337Client` function.
- Updated imports and removed unused imports.

### DIFF
--- a/packages/alchemy/src/provider.ts
+++ b/packages/alchemy/src/provider.ts
@@ -1,12 +1,13 @@
 import {
   BaseSmartContractAccount,
   SmartAccountProvider,
+  createPublicErc4337Client,
   deepHexlify,
   resolveProperties,
   type AccountMiddlewareFn,
   type SmartAccountProviderOpts,
 } from "@alchemy/aa-core";
-import type { Address, Chain, HttpTransport } from "viem";
+import { type Address, type Chain, type HttpTransport } from "viem";
 import {
   arbitrum,
   arbitrumGoerli,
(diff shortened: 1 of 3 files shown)


avasisht23:
* main
  * **PR #81** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/81" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/81?utm_source=stack-comment).

packages/alchemy/src/provider.ts
```diff
@@ -1,33 +1,35 @@
...
-import type { ClientWithAlchemyMethods } from "./middleware/client.js";
 
 type ConnectionConfig =
   | {
       apiKey: string;
       rpcUrl?: undefined;
     }
-  | { rpcUrl: string; apiKey?: undefined };
+  | { rpcUrl: string; apiKey?: undefined }
+  | { rpcUrl: string; apiKey?: undefined; jwt: string };
```
moldy530, line 32: This is a breaking change, we need to support the case where RPC url is set but the other two are not set still.

Also nit: while we're here can we make the `?: undefined` -> `?: never`

moldy530 (approved):

packages/alchemy/src/provider.ts
```diff
@@ -1,33 +1,35 @@
...
-import type { ClientWithAlchemyMethods } from "./middleware/client.js";
 
 type ConnectionConfig =
   | {
       apiKey: string;
-      rpcUrl?: undefined;
+      rpcUrl?: never;
     }
-  | { rpcUrl: string; apiKey?: undefined };
+  | { rpcUrl: string; apiKey?: never }
```
moldy530, line 31: oh right, I ran into a typing thing if I didn't also have `jwt?: never` in all the other options. Then you can do `connectionConfig.jwt != null ? {} : {headers: ...}` down below

I wonder if this causes any problems elsewhere re typing. probs not

moldy530 (approved):

moldy530 (approved):

PR #88, opened 2023-08-25
fix: pass overrides from tx to user op

raymondfeng:
The PR makes sure TX overrides to be passed into user op.

---

### Detailed summary
- The focus of this PR is to update the `buildUserOperation` function in the `base.ts` file.
- Notable changes include:
  - The `buildUserOperation` function now accepts an additional `overrides` parameter.
  - The `overrides` parameter is passed as an argument when calling the `buildUserOperation` function in the `sendTransactions` function.


moldy530 (approved):
oh good catch!

moldy530 (approved):

PR #90, opened 2023-08-27
feat: add `JsonRpcAccountSigner`

### LINKED ISSUE #45: Cannot read properties of undefined: `getAddress` when querying for CounterFactual address
### [REQUIRED] Environment
image

- Browser version: Chrome - 114.0.5735.198 
- AA SDK version: 0.1.0-alpha.14
- Package: alchemy/aa-core

### [REQUIRED] Describe the problem

When someone tries to query for the CounterFactual address of the `SimpleSmartContractAccount` then the error Cannot read properties of undefined: `getAddress` is thrown

#### How to reproduce:

```javascript
let provider = new SmartAccountProvider(
      "[rpc_url]",
      ENTRYPOINT_ADDRESS,
      sepolia
).connect((rpcClient) => {
    return new SimpleSmartContractAccount({
        entryPointAddress: ENTRYPOINT_ADDRESS,
        chain: sepolia,
        factoryAddress: SIMPLE_ACCOUNT_FACTORY_ADDRESS,
        rpcClient,
        owner: customClient,
    });
});

// Trying to query for the CounterFactual address of Smart Account
let address = await provider.account.getAddress();
```

therealharpaljadeja:
This PR adds support for [`JsonRpcAccount`](https://viem.sh/docs/clients/wallet.html#json-rpc-accounts) type `WalletClient` from `viem` to be used with `aa-sdk`

Fixes #45 

How to use:

```ts
import { useWalletClient } from "wagmi";
const { data: walletClient } = useWalletClient();

let provider = new AlchemyProvider({
    apiKey: process.env.NEXT_PUBLIC_ALCHEMY_API_KEY,
    chain: sepolia,
    entryPointAddress: ENTRYPOINT_ADDRESS,
}).connect((rpcClient) => {
    return new SimpleSmartContractAccount({
        entryPointAddress: ENTRYPOINT_ADDRESS,
        chain: sepolia,
        factoryAddress: SIMPLE_ACCOUNT_FACTORY_ADDRESS,
        rpcClient,
        owner: new JsonRpcAccountSigner(walletClient),
    });
});
```

---

### Focus of the PR:
This PR focuses on adding a new `WalletClientSigner` class to the codebase.

### Detailed summary:
- Added `WalletClientSigner` class to `packages/core/src/signer/wallet-client.ts`
- Imported necessary dependencies from `viem` and `./types`
- Implemented methods `getAddress`, `signMessage`, and `signTypedData` in the `WalletClientSigner` class


packages/core/src/signer/json-rpc-account.ts
```diff
@@ -0,0 +1,40 @@
+import { getAddress, type ByteArray, type Hex, type WalletClient } from "viem";
+import type { SmartAccountSigner } from "./types";
+import type { SignTypedDataParams } from "../account/types";
+
+export class JsonRpcAccountSigner implements SmartAccountSigner {
```
moldy530, line 5: nit:
```suggestion
export class WalletClientSigner implements SmartAccountSigner {
```

packages/core/src/signer/json-rpc-account.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  getAddress: () => Promise<`0x${string}`> = async () => {
+    let addresses = await this.client.getAddresses();
+    return getAddress(addresses[0]);
+  };
+
+  readonly signMessage: (
+    message: string | Hex | ByteArray
+  ) => Promise<`0x${string}`> = async (message) => {
+    if (typeof message === "string") {
```
moldy530, line 20: `Hex` is a string too, but needs to be passed as a `raw` parameter below

```suggestion
    if (typeof message === "string" && !message.startsWith("0x")) {
```

packages/core/src/signer/json-rpc-account.ts
therealharpaljadeja: Complying with this, since there is a `LocalAccountSigner`

image

moldy530: JsonRpcAccountSigner is too generic a name here. Viem's `WalletClient` isn't the only implementation of a JSON RPC provider, the `PublicClient` is as well. The ask here is to make the name specific to the use case. This signer is used to adapt a `WalletClient` to `SmartAccountSigner` so this should be a `WalletClientSigner` or some name that makes the use case more clear. 

`LocalAccountSigner` refers to the `viem` type [LocalAccount](https://github.com/wagmi-dev/viem/blob/cf4d5d8c666b7fd78a221974bb27162c0c6334b5/src/accounts/types.ts#L49). A `WalletClient` can actually be initialized with a `LocalAccount` as well as you pointed out. So it's not entirely accurate to call it just `JsonRpcAccountSigner`

packages/core/src/signer/json-rpc-account.ts
moldy530, line 1: sorry missed the file name too, mind updating this to be `wallet-client.ts`?

moldy530 (approved):

noam-alchemy:
@therealharpaljadeja - please check the lint errors that are blocking merging.

packages/core/src/signer/wallet-client.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  getAddress: () => Promise<`0x${string}`> = async () => {
+    let addresses = await this.client.getAddresses();
+    return getAddress(addresses[0]);
+  };
+
+  readonly signMessage: (
+    message: string | Hex | ByteArray
+  ) => Promise<`0x${string}`> = async (message) => {
+    if (typeof message === "string" && !message.startsWith("0x")) {
```
moldy530, line 20: you can probably fix the lint error by doing
```suggestion
    if (typeof message === "string" && !isHex(message)) {
```

isHex is exported by viem. that was my bad, totally forgot that helper existed

moldy530 (approved):

PR #112, opened 2023-10-03
feat: add `dropAndReplaceUserOperation`

### LINKED ISSUE #82: FR: support `dropAndReplace` flow for UserOperations
**Is your feature request related to a problem? Please describe.**
On occasion fee markets move and UOs get stuck as pending in the mempool. In those instances, it's often desirable to resubmit the UO so that it clears faster. To do so, it needs to be the higher of (1.1x the last price OR the new fee market price). 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

therealharpaljadeja:
@moldy530 I wasn't able to test `dropAndReplaceUserOperation`

When I tried overriding gas prices I got the invalid paymaster signature error, I was thinking of overriding with less gas price so that the UserOp gets stuck and then replacing it with a UserOp with no override.

telegram-cloud-photo-size-5-6115924363560728530-y

This is how I am overriding.

telegram-cloud-photo-size-5-6115924363560728531-y

Closes #82

---

### Detailed summary
- Added a new utility function `bigIntMax` to calculate the maximum value among a list of `bigint` values.
- Refactored code in `provider/base.ts` to use the new `bigIntMax` function.
- Created a new method `_runMiddlewareStack` to run the middleware stack and apply overrides to the user operation data.
- Updated the `sendUserOperation` method to use the `_runMiddlewareStack` method.
- Added a new method `dropAndReplaceUserOperation` to drop and replace a user operation with updated values.
- Updated the `dropAndReplaceUserOperation` method to use the `_runMiddlewareStack` method.

### DIFF
--- a/packages/core/src/provider/base.ts
+++ b/packages/core/src/provider/base.ts
@@ -324,15 +324,10 @@ export class SmartAccountProvider<
   ) => {
     const uoStruct = await this.UserOperationCallDataToStruct(data);
 
-    const uoStructAfterMiddlewares = await asyncPipe(
-      this.dummyPaymasterDataMiddleware,
-      this.feeDataGetter,
-      this.gasEstimator,
-      this.customMiddleware ?? noOpMiddleware,
-      // This applies the overrides if they've been passed in
-      async (struct) => ({ ...struct, ...overrides }),
-      this.paymasterDataMiddleware
-    )(uoStruct);
+    const uoStructAfterMiddlewares = await this._runMiddlewareStack(
+      uoStruct,
+      overrides
+    );
 
     return resolveProperties(uoStructAfterMiddlewares);
   };
@@ -554,21 +549,17 @@ export class SmartAccountProvider<
   };
 
   private _runMiddlewareStack = async (
-    data: UserOperationRequest,
+    data: Deferrable<UserOperationStruct>,
     overrides?: UserOperationOverrides
   ) => {
-    if (!this.account) {
-      throw new Error("account not connected!");
-    }
-
     const uoStruct = await asyncPipe(
       this.dummyPaymasterDataMiddleware,
       this.feeDataGetter,
       this.gasEstimator,
-      this.paymasterDataMiddleware,
       this.customMiddleware ?? noOpMiddleware,
       // This applies the overrides if they've been passed in
-      async (struct) => ({ ...struct, ...overrides })
+      async (struct) => ({ ...struct, ...overrides }),
+      this.paymasterDataMiddleware
     )(data);
 
     return resolveProperties(uoStruct);
--- a/packages/core/src/provider/base.ts
+++ b/packages/core/src/provider/base.ts
@@ -34,6 +34,7 @@ import {
   getUserOperationHash,
   resolveProperties,
   type Deferrable,
+  bigIntMax,
 } from "../utils.js";
 import type {
   AccountMiddlewareFn,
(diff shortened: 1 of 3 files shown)


packages/core/src/provider/base.ts
```diff
@@ -540,4 +562,33 @@ export class SmartAccountProvider<
...
+    }
+
+    const uoStruct = await asyncPipe(
+      this.dummyPaymasterDataMiddleware,
+      this.feeDataGetter,
+      this.gasEstimator,
+      this.paymasterDataMiddleware,
+      this.customMiddleware ?? noOpMiddleware,
+      // This applies the overrides if they've been passed in
+      async (struct) => ({ ...struct, ...overrides })
```
moldy530, line 581: The paymaster error makes sense, and is something I totally overlooked. These actually need to run before the paymaster middleware runs.

It also means that we can't run the full stack during drop and replace.

It needs to run as:
1. get fees
2. Apply drop and replace logic
3. Run the rest of the stack
4. Send user op

packages/core/src/provider/base.ts
```diff
@@ -540,4 +562,33 @@ export class SmartAccountProvider<
       };
     };
   };
+
+  private _runMiddlewareStack = async (
```
moldy530, line 566: We should use this not just for drop and replace but also in the main function too (_buildUserOp)

packages/core/src/provider/base.ts
```diff
@@ -540,4 +562,33 @@ export class SmartAccountProvider<
...
+      this.paymasterDataMiddleware,
+      this.customMiddleware ?? noOpMiddleware,
+      // This applies the overrides if they've been passed in
+      async (struct) => ({ ...struct, ...overrides })
+    )(data);
+
+    return resolveProperties(uoStruct);
+  };
+
+  private bigIntMax = (...args: bigint[]) => {
```
moldy530, line 587: Nit: can we move this into utils and make sure to export it in index

packages/core/src/provider/base.ts
moldy530: Lemme get back to you on this tomorrow or Thursday, might need to rethink the approach here slightly

packages/core/src/provider/base.ts
therealharpaljadeja: For this, I will have to create a utility `UserOperationCalldata` to `UserOperationStruct`.

I am thinking of keeping it in the `base.ts` file itself since it is specific to provider and not a general utility function.

packages/core/src/provider/base.ts
therealharpaljadeja: Can we say that the paymasterMiddleware should always be at the end?

I am not sure what kind of middlewares devs will use as `customMiddleware` but the process of signature should be at the end right?

therealharpaljadeja:
I was able to make overrides work.

I moved the overrides above the `paymasterDataMiddleware`.

I think the Gas Manager has no UI for dropped userOps, currently it shows expired.

```ts
const uoStructAfterMiddlewares = await asyncPipe(
    this.dummyPaymasterDataMiddleware,
    this.feeDataGetter,
    this.gasEstimator,
    this.customMiddleware ?? noOpMiddleware,
    // This applies the overrides if they've been passed in
    async (struct) => ({ ...struct, ...overrides }),
    this.paymasterDataMiddleware
)(uoStruct);
```

image

UserOperation [1](https://www.jiffyscan.xyz/userOpHash/0x565fc99ede6d12fa3ba5b76d4e322eb10e2e5debd6406fac49a803583cc51a87?network=mumbai), [2](https://www.jiffyscan.xyz/userOpHash/0xdfff2de1ab9fc81877d0617315112fd6ab51cf69f4bb668ac8cc03d7025fb67d?network=mumbai), [3](https://www.jiffyscan.xyz/userOpHash/0x14bb8bb9c53c3919831636503bfd1d2d7386de9c1afe04b5475bcf2e3d6f5cbf?network=mumbai) dropped by [4](https://www.jiffyscan.xyz/userOpHash/0xcb50b6cd5b80035ad68edec5689d2f940df4bf8ac2b56d668b157645012f3026?network=mumbai)


Is this the right way to fix the overrides issue? @moldy530

moldy530:
this is now merged in!

PR #170, opened 2023-10-24
docs: update the magic docs to highlight signer creation is async

moldy530:
### Detailed summary

This PR focuses on updating the `magicSigner` import and usage in the code. The notable changes include:

- Renamed the import from `magicSigner` to `createMagicSigner`
- Updated the `owner` assignment to use `await createMagicSigner()` instead of `magicSigner`
- Added an async function `createMagicSigner` that authenticates the user and creates a `WalletClientSigner` object using the `magic` SDK and `magicClient`

### DIFF
--- a/site/snippets/magic.ts
+++ b/site/snippets/magic.ts
@@ -1,4 +1,4 @@
-import { WalletClientSigner, type SmartAccountSigner } from "@alchemy/aa-core";
+import { WalletClientSigner } from "@alchemy/aa-core";
 import { Magic } from "magic-sdk";
 import { createWalletClient, custom } from "viem";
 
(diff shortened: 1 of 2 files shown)


site/snippets/magic.ts
```diff
@@ -9,14 +9,18 @@ const MAGIC_API_KEY = "pk_test_...";
 export const magic = new Magic(MAGIC_API_KEY);
 
 // a viem wallet client that wraps magic for utility methods
-// NOTE: this isn't necessary since you can just use the `magic.rpcProvider`
-// directly, but this makes things much easier
+// NOTE: because this is async, you will need to put this in a useEffect hook if using React
 export const magicClient = createWalletClient({
-  transport: custom(magic.rpcProvider),
```
denniswon, line 15: this one doesn't work? did not look into magic code, but why is getting rpc provider async here?
hmm, and just using rpcProvider field directly seems to work fine on examples I tried.

denniswon (dismissed):

site/snippets/magic.ts
moldy530: yea I'm not sure why this didn't work for a customer... maybe it was because you have to authenticate with a user first?

lemme take another pass at this and make it even more clear that authentication is required first

moldy530: in the unauthed case `.rpcProvider` was returning undefined

avasisht23: hmm I didn't need to do this if I mapped the magicClient to `any` for the new WalletClientSigner below:

https://github.com/alchemyplatform/aa-simple-dapp/blob/a71b0b78/src/hooks/useMagicSigner.ts#L17

avasisht23 (approved):
unblocking, but yeah curious why it needs to be async

moldy530:
needs to be async because the method to auth with magic is async. if you don't auth first then you get errors when connecting the account and using it. to make it simpler, I think we should just recommend that people auth first before connecting

PR #171, opened 2023-10-26
feat: add zod runtime validation for base provider

avasisht23:
## PR-Codex overview
This PR focuses on adding the `zod` package as a dependency and making changes related to the `SmartAccountProvider` class. 

### Detailed summary
- Added the `zod` package as a dependency
- Removed the `type` exports for `SmartAccountProviderConfig` and `SmartAccountProviderOpts` from `provider/base.js`
- Added `SmartAccountProviderOpts` and `SmartAccountProviderConfig` types to `provider/types.js`
- Updated the `SmartAccountProvider` constructor to use the `SmartAccountProviderConfig` type
- Updated the usage of `SmartAccountProvider` in tests to include a dummy entry point address

> The following files were skipped due to too many changes: `packages/core/src/provider/__tests__/base.test.ts`


avasisht23:
* main
  * **PR #171** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/171" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/171?utm_source=stack-comment).

denniswon:
This task https://app.asana.com/0/1205598840815267/1205773893380310/f will be abstracting away the entrypointaddress input entirely. Do we still need zod like runtime validation for the provider config? Also, https://app.asana.com/0/1205598840815267/1205770922447512/f.

Not too big concern, but seems to me that doing validation and also abstraction are orthogonal efforts? cc: @moldy530

moldy530:
>In alignment to [app.asana.com/0/1205598840815267/1205773893380310/f](https://app.asana.com/0/1205598840815267/1205773893380310/f), wdyt about the overall direction of having a new type that extends Chain from viem and have an additional chain-specific settings like entry point contract address (Address)

I like this idea at first glance, let's get a PR with the proposal up so we can get a feel for. Open question, should people still be able to pass a `viem` Chain? I think yes because aa-core is not opinionated on underlying RPC, chains, contracts, etc. So does this create more confusion than providing factory methods for sane defaults?

let's move this specific convo out of this PR though either way and into Asana or into a PR proposing the fix for abstracting entrypoint away

packages/core/src/provider/validation.ts
```diff
@@ -0,0 +1,53 @@
...
+   */
+  txRetryIntervalMs: z.number().optional(),
+
+  /**
+   * The mulitplier on interval length to wait between retries while waiting for tx receipts (default: 1.5)
+   */
+  txRetryMulitplier: z.number().optional(),
+
+  /**
+   * used when computing the fees for a user operation (default: 100_000_000n)
```
moldy530, line 22: zod supports setting default values in the schema via https://zod.dev/?id=default

packages/core/src/provider/validation.ts
```diff
@@ -0,0 +1,53 @@
...
+   */
+  minPriorityFeePerBid: z.bigint().optional(),
+});
+
+export const SmartAccountProviderConfigSchema = z.object({
+  rpcProvider: z.union([
+    z.string(),
+    z.any().refine((provider) => {
```
moldy530, line 30: we should type the `refinement` so we get accurate types here
```suggestion
    z.any().refine<PublicErc4337Client>((provider) => {
```

packages/core/src/provider/validation.ts
```diff
@@ -0,0 +1,53 @@
...
+      return (
+        typeof provider === "object" &&
+        "request" in provider &&
+        "type" in provider &&
+        "key" in provider &&
+        "name" in provider
+      );
+    }),
+  ]),
+  chain: z.any().refine((chain) => {
```
moldy530, line 40: same comment here 
```suggestion
  chain: z.any().refine<Chain>((chain) => {
```

moldy530 (commented):
We should update tests to include testing for missing or incorrect params as well

packages/core/src/index.ts
```diff
@@ -33,10 +33,6 @@ export {
 } from "./ens/utils.js";
 
 export { SmartAccountProvider, noOpMiddleware } from "./provider/base.js";
-export type {
-  SmartAccountProviderConfig,
-  SmartAccountProviderOpts,
-} from "./provider/base.js";
```
moldy530, line 39: we should still be exporting the new types and/or schemas

packages/core/src/provider/base.ts
```diff
@@ -49,48 +49,19 @@ import type {
...
   SendUserOperationResult,
+  SmartAccountProviderConfig,
 } from "./types.js";
+import { SmartAccountProviderConfigSchema } from "./validation.js";
 
 export const noOpMiddleware: AccountMiddlewareFn = async (
   struct: Deferrable<UserOperationStruct>
 ) => struct;
 
-export interface SmartAccountProviderOpts {
```
moldy530, line 58: we should export these still as export them as `z.infer<typeof SmartAccountProviderOptsSchema>`

packages/core/src/provider/base.ts
```diff
@@ -49,48 +49,19 @@ import type {
...
-   */
-  minPriorityFeePerBid?: bigint;
-}
-
 const minPriorityFeePerBidDefaults = new Map<number, bigint>([
   [arbitrum.id, 10_000_000n],
   [arbitrumGoerli.id, 10_000_000n],
 ]);
 
-export type SmartAccountProviderConfig<
```
moldy530, line 85: same here we should still export this stuff (maybe from `types`)

packages/core/src/provider/validation.ts
moldy530, line 1: super-nit: i'm not crazy about `validation` maybe `schema`?

we should also make sure we export the infered type for the schema from within `types`

packages/core/src/index.ts
avasisht23: these are still exported as part of `export type * from "./provider/types.js";` since i moved these to the types folder

packages/core/src/provider/validation.ts
avasisht23: set them, but it didn't make a difference at compile time re: still needing to have defaults on the base provider's field variables during init() call for constructor.

avasisht23:
>We should update tests to include testing for missing or incorrect params as well

added tests in provider/__tests__/base.test.ts

packages/core/src/provider/base.ts
```diff
@@ -109,12 +80,17 @@ export class SmartAccountProvider<
...
-    opts,
-  }: SmartAccountProviderConfig<TTransport>) {
+  constructor(config: SmartAccountProviderConfig<TTransport>) {
+    try {
+      SmartAccountProviderConfigSchema<TTransport | HttpTransport>().parse(
+        config
+      );
+    } catch (e) {
+      throw new Error("invalid parameters passed to SmartAccountProvider");
+    }
```
moldy530, line 90: either rethrow the zod error or just run it without the try catch block

zod gives detailed messages about what part of schema validation fails which this error message does not capture

packages/core/src/provider/base.ts
```diff
@@ -109,12 +80,17 @@ export class SmartAccountProvider<
...
 
-  constructor({
-    rpcProvider,
-    entryPointAddress,
-    chain,
-    opts,
-  }: SmartAccountProviderConfig<TTransport>) {
+  constructor(config: SmartAccountProviderConfig<TTransport>) {
+    try {
+      SmartAccountProviderConfigSchema<TTransport | HttpTransport>().parse(
```
moldy530, line 85: this should be just `TTransport` right? not `HttpTransport`?

moldy530 (commented):
Ive lost track of the comment on infer, but saw the email

https://stackoverflow.com/a/72781140

I think if we want the default to be populated you want to define the type with `z.input` or `z.output` instead of z.infer

moldy530:
> Ive lost track of the comment on infer, but saw the email
> 
> https://stackoverflow.com/a/72781140
> 
> I think if we want the default to be populated you want to define the type with `z.input` or `z.output` instead of z.infer

I think we want `z.input`. `z.output` is useful if we take in a type that has `string | number` and we refine it into something like `bigint` after the fact.

packages/core/src/provider/base.ts
avasisht23: oh yeah I was copying from the `rpcClient` field on the provider: 

```
  rpcClient:
    | PublicErc4337Client<TTransport>
    | PublicErc4337Client<HttpTransport>;
```

avasisht23: but can remove HTTPTransport and it'll work the same

avasisht23:
> I think we want z.input. z.output is useful if we take in a type that has string | number and we refine it into something like > bigint after the fact. In this case the z.output type would be used by anything downstream of the parse

hmm this doesn't work re: using .input, typescript can't infer the types correctly:
image

going to stick with .infer and that was doing fine. .output also workers.

packages/core/src/provider/__tests__/base.test.ts
```diff
@@ -175,6 +177,59 @@ describe("Base Tests", () => {
     expect(newProvider.testMethod()).toEqual("test");
   });
 
+  it("should correctly do runtime validation", () => {
```
moldy530, line 180: nit:

```suggestion
  it("should correctly do runtime validation when entrypoint is incorrect", () => {
```

packages/core/src/provider/__tests__/base.test.ts
```diff
@@ -175,6 +177,59 @@ describe("Base Tests", () => {
...
+          \\"path\\": [
+            \\"entryPointAddress\\"
+          ],
+          \\"message\\": \\"Expected string, received number\\"
+        }
+      ]"
+    `);
+  });
+
+  it("should correctly do runtime validation 2", () => {
```
moldy530, line 203:
```suggestion
  it("should correctly do runtime validation when chain is incorrect", () => {
```

packages/core/src/provider/schema.ts
```diff
@@ -0,0 +1,66 @@
...
+import type { Chain, Transport } from "viem";
+import z from "zod";
+import type { PublicErc4337Client, SupportedTransports } from "../client/types";
+import { getChain } from "../utils/index.js";
+
+export const SmartAccountProviderOptsSchema = z.object({
+  /**
+   * The maximum number of times to try fetching a transaction receipt before giving up (default: 5)
+   */
+  txMaxRetries: z.number().optional().default(5),
```
moldy530, line 11: ah for the `infer` vs `output` vs `input` i meant it in regard to issues you ran into with defaults? are you still having issues with defaults?

moldy530 (dismissed):

moldy530:
ah wait I approved ,but can you change base branch to development?

packages/core/src/provider/schema.ts
avasisht23: yeah still had issues. can run it by you next week

moldy530 (approved):

PR #181, opened 2023-10-29
feat: add sanity check on provider connect for clearer error message

denniswon:
[app.asana.com/0/1205598840815267/1205773893380310/f](https://app.asana.com/0/1205598840815267/1205773893380310/f)

Per the task linked above, added a sanity check early on on the current `connect()` so that users can get clear error message about the issue.

Follow up PR will made to make entry point contract address param as optional for SmartContractAccount

Also, updated the doc for the missing/added `Other` methods for SCA class.

---

### Detailed summary:
This PR introduces the `getEntryPointAddress` method to the `ISmartContractAccount` interface and the `BaseSmartContractAccount` class. It also adds implementations and documentation for the method in various files. Additionally, it updates the `SmartAccountProvider` class to use the connected account's entry point address or the default entry point address for the chain.

> The following files were skipped due to too many changes: `packages/core/src/provider/__tests__/base.test.ts`

### DIFF
--- a/packages/core/src/provider/__tests__/base.test.ts
+++ b/packages/core/src/provider/__tests__/base.test.ts
@@ -77,6 +77,7 @@ describe("Base Tests", () => {
       rpcClient: providerMock.rpcClient,
       getAddress: async () => "0xMOCK_ADDRESS",
       getFactoryAddress: () => "0xMOCK_FACOTRY_ADDRESS",
+      getEntryPointAddress: () => entryPointAddress,
       getOwner: () => undefined,
     } as any;
 
@@ -100,6 +101,20 @@ describe("Base Tests", () => {
         ],
       ]
     `);
+  });
+
+  it("should throw error if connected account has different entry point address than the provider", async () => {
+    const account = {
+      chain: polygonMumbai,
+      entryPointAddress: "0xOTHER_ENTRY_POINT_ADDRESS",
+      rpcClient: providerMock.rpcClient,
+      getAddress: async () => "0xMOCK_ADDRESS",
+      getFactoryAddress: () => "0xMOCK_FACOTRY_ADDRESS",
+      getEntryPointAddress: () => "0xOTHER_ENTRY_POINT_ADDRESS",
+      getOwner: () => undefined,
+    } as any;
+
+    expect(() => providerMock.connect(() => account)).toThrow();
   });
 
   it("should emit disconnected event on disconnect", async () => {
--- a/packages/core/src/provider/__tests__/base.test.ts
+++ b/packages/core/src/provider/__tests__/base.test.ts
@@ -127,7 +127,11 @@ describe("Base Tests", () => {
       chain,
     });
 
-    expect(() => providerMockWithEntryPoint.connect(() => account)).toThrow();
+    expect(() =>
+      providerMockWithEntryPoint.connect(() => account)
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"Account entryPoint address: ${entryPointAddress} does not match the current provider's entryPoint address: ${dummyEntryPointAddress}"`
+    );
   });
 
   it("should emit disconnected event on disconnect", async () => {
--- a/packages/core/src/provider/__tests__/base.test.ts
+++ b/packages/core/src/provider/__tests__/base.test.ts
@@ -8,8 +8,14 @@ import {
   vi,
   type SpyInstance,
 } from "vitest";
+import { SimpleSmartContractAccount } from "../../account/simple.js";
+import { LocalAccountSigner } from "../../signer/local-account.js";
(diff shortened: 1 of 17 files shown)


packages/core/src/provider/__tests__/base.test.ts
```diff
@@ -102,6 +103,20 @@ describe("Base Tests", () => {
     `);
   });
 
+  it("should throw error if connected account has different entry point address than the provider", async () => {
```
moldy530, line 106: should this maybe throw in this case IFF the provider was initialized explicitly with an entrypoint?

that is if the provider init had undefined entrypoint it would inherit from the connected account

packages/core/src/provider/__tests__/base.test.ts
```diff
@@ -102,6 +103,20 @@ describe("Base Tests", () => {
...
+    const account = {
+      chain: polygonMumbai,
+      entryPointAddress: "0xOTHER_ENTRY_POINT_ADDRESS",
+      rpcClient: providerMock.rpcClient,
+      getAddress: async () => "0xMOCK_ADDRESS",
+      getFactoryAddress: () => "0xMOCK_FACOTRY_ADDRESS",
+      getEntryPointAddress: () => "0xOTHER_ENTRY_POINT_ADDRESS",
+      getOwner: () => undefined,
```
avasisht23, line 114: hmm this may fail with zod being able to check if they are actually addresses at runtime (i.e. hex values, not 0x + english). considering making dummy values?

site/packages/aa-core/accounts/other/getOwner.md
```diff
@@ -0,0 +1,37 @@
...
+
+<<< @/snippets/provider.ts
+
+:::
+
+## Returns
+
+### `SmartAccountSigner | undefined`
+
+The `SmartAccountSigner` object that repensents the current owner for the account
```
avasisht23, line 37:
```suggestion
The `SmartAccountSigner` object that represents the current owner for the account
```

site/packages/aa-core/accounts/other/getOwner.md
```diff
@@ -0,0 +1,37 @@
...
+    - name: description
+      content: Overview of the getOwner method on BaseSmartContractAccount
+  - - meta
+    - property: og:description
+      content: Overview of the getOwner method on BaseSmartContractAccount
+---
+
+# getOwner
+
+Returns the `SmartAccountSigner` that repensents the current owner for the account
```
avasisht23, line 17:
```suggestion
Returns the `SmartAccountSigner` that represents the current owner for the account.
```

avasisht23 (commented):
few comments/suggestions

packages/core/src/provider/__tests__/base.test.ts
moldy530: yea could use SimpleAccount here too and init it with a different address than the provider

moldy530: from a "tests as documentation" perspective using SimpleAccount is probably a better example of how things work versus this

packages/core/src/provider/__tests__/base.test.ts
denniswon: good catch. yes

packages/core/src/provider/base.ts
```diff
@@ -589,6 +601,17 @@ export class SmartAccountProvider<
...
+  /*
+   * Note that the connected account's entryPointAddress always takes the precedence
+   */
+  getEntryPointAddress = (): Address => {
+    return (
+      this.account?.getEntryPointAddress() ??
+      this.entryPointAddress ??
+      getDefaultEntryPointAddress(this.chain)
+    );
+  };
```
avasisht23, line 613: i think it should be this order so we prioritize the current object's entryPoint over its account. it shouldn't matter though if they're the same, but just in case.

```suggestion
  getEntryPointAddress = (): Address => {
    return (
      this.entryPointAddress ??
      this.account?.getEntryPointAddress() ??
      getDefaultEntryPointAddress(this.chain)
    );
  };
```

packages/core/src/provider/base.ts
```diff
@@ -537,6 +536,19 @@ export class SmartAccountProvider<
     ) => TAccount
   ): this & { account: TAccount } => {
     const account = fn(this.rpcClient);
+
+    // sanity check. Note that this check is only performed iff the optional entryPointAddress is given upon initialization.
```
avasisht23, line 540: only because some people might think it's a typo and not the proof notation shorthand 😅 
```suggestion
    // sanity check. Note that this check is only performed if and only if the optional entryPointAddress is given upon initialization.
```

avasisht23 (approved):
this looks good. few nits, then 🚢

PR #190, opened 2023-10-31
feat: biconomy integration

AmanRaj1608:
## PR-Codex overview
This PR focuses on integrating Biconomy functionality into the accounts package. 

### Detailed summary
- Exported several modules and types related to Biconomy.
- Added BiconomyAccountProvider class.
- Added BiconomyFactoryAbi constant.
- Added constants and types related to Biconomy utilities.
- Added preverification.ts file with functions for calculating gas costs.
- Added types related to Biconomy utilities.

> The following files were skipped due to too many changes: `packages/accounts/src/biconomy/utils/types.ts`, `packages/accounts/src/biconomy/abi/SmartAccount.ts`, `packages/accounts/src/biconomy/account.ts`, `yarn.lock`


packages/accounts/src/biconomy/provider.ts
```diff
@@ -0,0 +1,42 @@
...
+  SmartAccountProvider,
+  type BatchUserOperationCallData,
+  type SendUserOperationResult,
+  type UserOperationCallData,
+} from "@alchemy/aa-core";
+import type { Hex, HttpTransport } from "viem";
+import { BiconomySmartAccountV2 } from "./account.js";
+
+export class BiconomyAccountProvider extends SmartAccountProvider<HttpTransport> {
+  sendUserOperation = async (
```
moldy530, line 11: any reason why this can't leverage the SmartAccountProvider's default flows?

Looking at the `account` impl, the gas estimation and paymaster logic could all be defined as `gasEstimationMiddleware` and `paymasterAndDataMiddleware`.

I would recommend looking at how the `AlchemyProvider` provides overrides for the that middleware. 

The benefit of doing it in that way is that it maintains composability. The `account` impl itself should be able to operate outside of a specific bundler impl. The provider is what should specify the specific bundler interactions (like specific fee estimation or gas estimation logic).

The default provider's logic for send user operation is:
1. Build the useroperation based on the CallData. This runs the middleware stack (fee estimation -> gas estimation -> applies overrides -> runs custom middleware (if specified) -> calls paymaster if specified)
2. Validates the useroperation struct and hexlifies the fields
3. asks the account to sign the useroperation struct by calling `account.signMessage(uoHash)` 
4. calls the rpc client to send the UO

The benefit of plugging into this flow is that it should reduce the overall complexity of your account definition and provider definition, and lets you put all of your estimation + paymaster logic into middleware functions.

If you still want to export a provider (which is definitely useful if you want to minimize the dev complexity of composing your middleware), then you can export one that lets you configure the middleware at the time of construction

packages/accounts/src/biconomy/utils/preverificaiton.ts
```diff
@@ -0,0 +1,128 @@
+import type { UserOperation } from "@biconomy/core-types";
```
dphilipson, line 1: Typo in the filename, "preverificaiton"

packages/accounts/src/biconomy/provider.ts
livingrockrises: > The account impl itself should be able to operate outside of a specific bundler impl
I agree. 
our account.ts implementation uses Ibundler.ts which is equivalent to rpcClient (4337PublicClient) so we should test it (rpcClient and gasEstimationMiddleWare) with different bundler urls

livingrockrises: @moldy530 Is there any example or guide for using below

```
readonly customMiddleware?: AccountMiddlewareFn | undefined = undefined;

withCustomMiddleware = (override: AccountMiddlewareFn): this => {
    defineReadOnly(this, "customMiddleware", override);

    return this;
  };
```

avasisht23: yeah, take a look at this doc: https://accountkit.alchemy.com/packages/aa-core/provider/withCustomMiddleware.html#custommiddleware

let us know if it's unclear!

avasisht23:
Hey @AmanRaj1608 closing this in favor of implementing your signer directly into aa-signers, per our discussion on telegram.

PR #197, opened 2023-11-01
feat: add support for overriding the initCode for an account

moldy530:
# Pull Request Checklist

- [x] Did you update relevant docs? (docs are found in the `site` folder, see guidleines below)
- [x] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples? (e.g. `feat!: breaking change`)
- [x] Did you run lint (`yarn lint:check`) and fix any issues? (`yarn lint:fix`)
- [x] Is the base branch you're merging into `development` and not `main`?

---

### Detailed summary
- Added support for optional override of the account address and init code in the `createBaseSmartAccountParamsSchema` function in `packages/core/src/account/schema.ts`.
- Modified the `SimpleSmartContractAccount` constructor in `packages/core/src/account/__tests__/simple.test.ts` to use the account init code override.
- Added the `accountInitCode` property to the `BaseSmartContractAccount` class in `packages/core/src/account/base.ts` to store the account init code.
- Modified the `getAccountInitCode` method in the `BaseSmartContractAccount` class to return the account init code override if available, otherwise call the abstract `getAccountInitCode` method.
- Added the `_getAccountInitCode` and `create6492Signature` methods to the `BaseSmartContractAccount` class to handle the account init code and signature creation logic.

### DIFF
--- a/packages/core/src/account/__tests__/simple.test.ts
+++ b/packages/core/src/account/__tests__/simple.test.ts
@@ -1,5 +1,5 @@
 import type { Address } from "viem";
-import { polygonMumbai, type Chain } from "viem/chains";
+import { polygonMumbai, sepolia, type Chain } from "viem/chains";
 import { describe, it } from "vitest";
 import { getDefaultSimpleAccountFactoryAddress } from "../../index.js";
 import { SmartAccountProvider } from "../../provider/base.js";
@@ -102,29 +102,49 @@ describe("Account Simple Tests", () => {
       ]"
     `);
   });
-});
 
-const givenConnectedProvider = ({
-  owner,
-  chain,
-}: {
-  owner: SmartAccountSigner;
-  chain: Chain;
-}) =>
-  new SmartAccountProvider({
-    rpcProvider: `${chain.rpcUrls.alchemy.http[0]}/${"test"}`,
-    chain,
-  }).connect((provider) => {
+  it("should correctly use the account init code override", async () => {
     const account = new SimpleSmartContractAccount({
-      chain,
-      owner,
-      factoryAddress: getDefaultSimpleAccountFactoryAddress(chain),
-      rpcClient: provider,
+      chain: sepolia,
+      owner: owner,
+      factoryAddress: getDefaultSimpleAccountFactoryAddress(sepolia),
+      rpcClient: `${sepolia.rpcUrls.alchemy.http[0]}/${"test"}`,
+      // override the account address here so we don't have to resolve the address from the entrypoint
+      accountAddress: "0x1234567890123456789012345678901234567890",
+      initCode: "0xdeadbeef",
     });
 
-    account.getAddress = vi.fn(
-      async () => "0xb856DBD4fA1A79a46D426f537455e7d3E79ab7c4"
-    );
+    // @ts-expect-error this object is protected
+    vi.spyOn(account.rpcProvider, "getBytecode").mockImplementation(() => {
+      return Promise.resolve("0x");
+    });
 
-    return account;
+    const initCode = await account.getInitCode();
+    expect(initCode).toMatchInlineSnapshot('"0xdeadbeef"');
   });
+
+  const givenConnectedProvider = ({
+    owner,
+    chain,
+  }: {
+    owner: SmartAccountSigner;
+    chain: Chain;
+  }) =>
+    new SmartAccountProvider({
+      rpcProvider: `${chain.rpcUrls.alchemy.http[0]}/${"test"}`,
+      chain,
+    }).connect((provider) => {
+      const account = new SimpleSmartContractAccount({
+        chain,
+        owner,
+        factoryAddress: getDefaultSimpleAccountFactoryAddress(chain),
+        rpcClient: provider,
+      });
+
+      account.getAddress = vi.fn(
+        async () => "0xb856DBD4fA1A79a46D426f537455e7d3E79ab7c4"
+      );
+
+      return account;
+    });
+});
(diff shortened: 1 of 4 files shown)


packages/core/src/account/__tests__/simple.test.ts
```diff
@@ -102,29 +102,49 @@ describe("Account Simple Tests", () => {
...
-    rpcProvider: `${chain.rpcUrls.alchemy.http[0]}/${"test"}`,
-    chain,
-  }).connect((provider) => {
+  it("should correctly use the account init code override", async () => {
     const account = new SimpleSmartContractAccount({
-      chain,
-      owner,
-      factoryAddress: getDefaultSimpleAccountFactoryAddress(chain),
-      rpcClient: provider,
+      chain: sepolia,
```
avasisht23, line 108: does the chain need to be sepolia as opposed to mumbai?

packages/core/src/account/__tests__/simple.test.ts
```diff
@@ -102,29 +102,49 @@ describe("Account Simple Tests", () => {
...
+      owner: owner,
+      factoryAddress: getDefaultSimpleAccountFactoryAddress(sepolia),
+      rpcClient: `${sepolia.rpcUrls.alchemy.http[0]}/${"test"}`,
+      // override the account address here so we don't have to resolve the address from the entrypoint
+      accountAddress: "0x1234567890123456789012345678901234567890",
```
avasisht23, line 113: does this matter if this test is just evaluating initcode?

packages/core/src/account/base.ts
```diff
@@ -51,8 +52,9 @@ export abstract class BaseSmartContractAccount<
     | PublicErc4337Client<TTransport>
     | PublicErc4337Client<HttpTransport>;
 
-  constructor(params: BaseSmartAccountParams<TTransport>) {
-    createBaseSmartAccountParamsSchema<TTransport>().parse(params);
+  constructor(params_: BaseSmartAccountParams<TTransport>) {
```
avasisht23, line 55: nit[learn]: why underscore after?

denniswon:
@moldy530 

q1: what was the context around needing this feature? This feature is for deployed accounts only or could be also used for non-deployed accounts?

q2: should we add some validations on when these optional parameters should be passed in so that devs can catch when incorrect set of parameters are passed in? (runtime validation using zod) cc: @avasisht23

packages/core/src/account/base.ts
```diff
@@ -316,9 +296,36 @@ export abstract class BaseSmartContractAccount<
...
     [Address, Hex]
   > {
-    const initCode = await this.getAccountInitCode();
+    const initCode = await this._getAccountInitCode();
     const factoryAddress = `0x${initCode.substring(2, 42)}` as Address;
     const factoryCalldata = `0x${initCode.substring(42)}` as Hex;
     return [factoryAddress, factoryCalldata];
   }
+
+  private async _getAccountInitCode(): Promise<Hash> {
```
avasisht23, line 305: why would we not want to use this method publicly, which leverages for the cached override?

avasisht23 (commented):
few qs before the stamp

moldy530:
Q1: context around this is to support the case where a developer wants to snapshot the initCode that generated a given counterfactual. This is useful to protect against the case where the counterfactual might change before a user eventually sends a UO and deploys their account. with this, a dev can make sure that when a UO is finally sent, they deploy the account the user has been using the counterfactual of, even if the default factory address or account impl changes down the line

Q2: I've added validations to Zod around passing this in already here. However, beyond that we can't do much else to enforce that a dev SHOULD pass them. it's opt-in if they want to override these params

packages/core/src/account/base.ts
moldy530: using this means that all of the methods on base actually depend on `getAccountInitCode` uses the override correctly. 

If instead this change was made to `getAccountInitCode` all implementors of `BaseSmartContractAccount` would have to call `super.getAccountInitCode` and check if that returns something. Thats not much different than just updating all sub account impls to check if `this.accountInitCode` is set. 

With this change, all existing implementations get this functionality for free

packages/core/src/account/base.ts
moldy530: oh because I wanted to use the `parsed` params instead

avasisht23: right I guess I meant as opposed to calling the constructor arg `_params`

packages/core/src/account/__tests__/simple.test.ts
moldy530: yea because `getInitCode` calls `getAddress` which uses the `_getAccountInitCode` to call the entrypoint to check the revert for the address. this short circuits that

packages/core/src/account/__tests__/simple.test.ts
moldy530: doesn't matter what the chain is because we don't make any RPC calls (they're stubbed out or short circuited)

moldy530: in the future, we'll want to actually use `localhost` in test with local rundler + anvil running, but we don't support that well yet

packages/core/src/account/base.ts
moldy530: ohhhh, `_params` implies not used

avasisht23 (approved):

avasisht23:
* development
  * **PR #197** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/197" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/197?utm_source=stack-comment).

PR #212, opened 2023-11-05
chore: nit updates to use Promise all and helper methods for readability

denniswon:
### Focus of the PR:
This PR focuses on improving error handling and setting the `maxPriorityFeePerGasBid` value in the SmartAccountProvider class.

### Detailed summary:
- Updated `bigIntMax` function in `bigint.ts` to throw an error if no arguments are provided.
- Updated error message in `base.ts` to include the `uoStruct` parameter.
- Updated `feeDataGetter` function in `base.ts` to use `bigIntMax` function and handle error cases more effectively.

### DIFF
--- a/packages/core/src/provider/base.ts
+++ b/packages/core/src/provider/base.ts
@@ -405,7 +405,7 @@ export class SmartAccountProvider<
       // this pretty prints the uo
       throw new Error(
         `Request is missing parameters. All properties on UserOperationStruct must be set. uo: ${JSON.stringify(
-          request,
+          uoStruct,
           null,
           2
         )}`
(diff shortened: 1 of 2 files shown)


packages/core/src/provider/base.ts
```diff
@@ -405,7 +405,7 @@ export class SmartAccountProvider<
       // this pretty prints the uo
       throw new Error(
         `Request is missing parameters. All properties on UserOperationStruct must be set. uo: ${JSON.stringify(
-          request,
+          uoStruct,
```
avasisht23, line 408: was there a reason not to use `request` here?

avasisht23 (approved):
quick q, otw good catches!

packages/core/src/provider/base.ts
denniswon: thought hex is less readable, because users will identify numbers better.

PR #243, opened 2023-11-13
feat: aa-alchemy ergonomic support for handling gas estimation

denniswon:
https://www.notion.so/alchemotion/2023-08-10-Alchemy-Gas-and-Paymaster-Update-020b41bee84641db98210cb71bca96af?pvs=4

Screenshot 2023-11-13 at 6 08 55 PM

```
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "alchemy_requestGasAndPaymasterAndData",
  "params": [
		{
			"policyId": "string",
			"entryPoint": "string",
			"dummySignature": "string",
			"userOperation": {
				"sender": "string",
				"nonce": "string",
				"initCode": "string",
				"callData": "string"
			},
			// OPTIONAL
			// Each value is either:
			// 1- absolute (hex) or 
			// 2- percentage between 1 -> 1000
      // Note: For backward compatiblity, we will process at feeOverrides 
      // if "overrides" is not provided.
			"overrides": {
				"maxFeePerGas": "hex string" | { "percentage": number },
				"maxPriorityFeePerGas": "hex string" | { "percentage": number },
        "callGasLimit": "hex string" | { "percentage": number },
				"verificationGasLimit": "hex string" | { "percentage": number },
				"preVerificationGas": "hex string" | { "percentage": number }
			}
		}
  ]
}

<!-- start pr-codex -->

---

## PR-Codex overview
This PR focuses on introducing fee options and overrides for user operations in the Alchemy provider. 

### Detailed summary
- Added `UserOperationFeeOptions` type for specifying fee options.
- Added `UserOperationOverrides` type for specifying overrides for user operations.
- Modified `getDefaultUserOperationFeeOptions` function to provide default fee options based on the chain.
- Updated `ClientWithAlchemyMethods` type to include `overrides` property for user operations.
- Updated `ClientWithAlchemyMethods` type to include `RequestGasAndPaymasterAndDataOverrides` type for specifying gas and paymaster overrides.
- Updated `ClientWithAlchemyMethods` type to include `RequestGasAndPaymasterAndDataOverrides` type for specifying gas and paymaster overrides.
- Updated `withAlchemyGasFeeEstimator` function to handle fee options and overrides for estimating gas fees.
- Updated `withAlchemyGasFeeEstimator` function to handle fee options and overrides for estimating gas fees.
- Updated `givenConnectedProvider` function to include `feeOptions` parameter for specifying fee options.

> The following files were skipped due to too many changes: `packages/alchemy/src/provider.ts`, `packages/core/src/provider/base.ts`, `packages/alchemy/src/middleware/gas-manager.ts`, `packages/alchemy/e2e-tests/simple-account.test.ts`

<!-- end pr-codex -->

### DIFF
--- a/packages/alchemy/src/middleware/client.ts
+++ b/packages/alchemy/src/middleware/client.ts
@@ -4,6 +4,7 @@ import {
 } from "@alchemy/aa-core";
 import type { Address, Hex } from "viem";
 import type {
+  RequestGasAndPaymasterAndDataOverrides,
   SimulateUserOperationAssetChangesRequest,
   SimulateUserOperationAssetChangesResponse,
 } from "./types/index.js";
@@ -30,10 +31,7 @@ export type ClientWithAlchemyMethods = PublicErc4337Client & {
             entryPoint: Address;
             userOperation: UserOperationRequest;
             dummySignature: Hex;
-            feeOverride?: {
-              maxFeePerGas: Hex;
-              maxPriorityFeePerGas: Hex;
-            };
+            overrides?: RequestGasAndPaymasterAndDataOverrides;
           }
         ];
       }): Promise<{
--- a/packages/alchemy/src/middleware/gas-manager.ts
+++ b/packages/alchemy/src/middleware/gas-manager.ts
@@ -1,11 +1,18 @@
 import {
   deepHexlify,
+  filterUndefined,
+  isPercentage,
   resolveProperties,
   type AccountMiddlewareFn,
+  type Hex,
+  type Percentage,
+  type UserOperationFeeOptions,
   type UserOperationRequest,
 } from "@alchemy/aa-core";
+import { fromHex } from "viem";
 import type { AlchemyProvider } from "../provider.js";
 import type { ClientWithAlchemyMethods } from "./client.js";
+import type { RequestGasAndPaymasterAndDataOverrides } from "./types/index.js";
 
 export interface AlchemyGasManagerConfig {
   policyId: string;
@@ -47,7 +54,7 @@ export const withAlchemyGasManager = <P extends AlchemyProvider>(
         .withGasEstimator(async (struct, overrides) => {
           // but if user is bypassing paymaster to fallback to having the account to pay the gas (one-off override),
           // we cannot delegate gas estimation to the bundler because paymaster middleware will not be called
-          if (overrides?.paymasterAndData !== undefined) {
+          if (overrides?.paymasterAndData != null) {
             const result = await fallbackGasEstimator(struct, overrides);
             return {
               callGasLimit: (await result.callGasLimit) ?? 0n,
@@ -69,7 +76,7 @@ export const withAlchemyGasManager = <P extends AlchemyProvider>(
 
           // but if user is bypassing paymaster to fallback to having the account to pay the gas (one-off override),
           // we cannot delegate gas estimation to the bundler because paymaster middleware will not be called
-          if (overrides?.paymasterAndData !== undefined) {
+          if (overrides?.paymasterAndData != null) {
             const result = await fallbackFeeDataGetter(struct, overrides);
             maxFeePerGas = (await result.maxFeePerGas) ?? maxFeePerGas;
             maxPriorityFeePerGas =
@@ -145,19 +152,41 @@ const withAlchemyPaymasterAndDataMiddleware = <P extends AlchemyProvider>(
 const withAlchemyGasAndPaymasterAndDataMiddleware = <P extends AlchemyProvider>(
   provider: P,
   config: AlchemyGasManagerConfig
-): Parameters<P["withPaymasterMiddleware"]>["0"] => ({
-  paymasterDataMiddleware: async (struct) => {
+): Parameters<AlchemyProvider["withPaymasterMiddleware"]>["0"] => ({
+  paymasterDataMiddleware: async (struct, overrides) => {
     const userOperation: UserOperationRequest = deepHexlify(
       await resolveProperties(struct)
     );
 
-    let feeOverride = undefined;
-    if (userOperation.maxFeePerGas && BigInt(userOperation.maxFeePerGas) > 0n) {
-      feeOverride = {
-        maxFeePerGas: userOperation.maxFeePerGas,
-        maxPriorityFeePerGas: userOperation.maxPriorityFeePerGas,
-      };
-    }
+    const overrideField = (
+      field: keyof UserOperationFeeOptions
+    ): Hex | Percentage | undefined => {
+      // one-off absolute override
+      if (overrides?.[field] != null) {
+        return deepHexlify(overrides[field]);
+      }
+
+      // provider level fee options with percentage
+      if (isPercentage(provider.feeOptions?.[field])) {
+        return {
+          percentage: 100 + Number(provider.feeOptions?.[field]?.percentage),
+        };
+      }
+
+      if (fromHex(userOperation[field], "bigint") > 0n) {
+        return userOperation[field];
+      }
+
+      return undefined;
+    };
+
+    const _overrides: RequestGasAndPaymasterAndDataOverrides = filterUndefined({
+      maxFeePerGas: overrideField("maxFeePerGas"),
+      maxPriorityFeePerGas: overrideField("maxPriorityFeePerGas"),
+      callGasLimit: overrideField("callGasLimit"),
+      verificationGasLimit: overrideField("verificationGasLimit"),
+      preVerificationGas: overrideField("preVerificationGas"),
+    });
 
     const result = await (
       provider.rpcClient as ClientWithAlchemyMethods
@@ -169,7 +198,8 @@ const withAlchemyGasAndPaymasterAndDataMiddleware = <P extends AlchemyProvider>(
           entryPoint: provider.getEntryPointAddress(),
           userOperation: userOperation,
           dummySignature: userOperation.signature,
-          feeOverride: feeOverride,
+          overrides:
+            Object.keys(_overrides).length > 0 ? _overrides : undefined,
         },
       ],
     });
--- a/packages/alchemy/src/middleware/gas-manager.ts
+++ b/packages/alchemy/src/middleware/gas-manager.ts
@@ -51,11 +51,15 @@ export const withAlchemyGasManager = <P extends AlchemyProvider>(
   return gasEstimationOptions.estimateGas
     ? provider
         // no-op gas estimator
-        .withGasEstimator(async (struct, overrides) => {
+        .withGasEstimator(async (struct, overrides, feeOptions) => {
           // but if user is bypassing paymaster to fallback to having the account to pay the gas (one-off override),
           // we cannot delegate gas estimation to the bundler because paymaster middleware will not be called
           if (overrides?.paymasterAndData != null) {
-            const result = await fallbackGasEstimator(struct, overrides);
+            const result = await fallbackGasEstimator(
+              struct,
+              overrides,
+              feeOptions
+            );
             return {
               callGasLimit: (await result.callGasLimit) ?? 0n,
               preVerificationGas: (await result.preVerificationGas) ?? 0n,
@@ -70,14 +74,18 @@ export const withAlchemyGasManager = <P extends AlchemyProvider>(
           }
         })
         // no-op fee because the alchemy api will do it
-        .withFeeDataGetter(async (struct, overrides) => {
+        .withFeeDataGetter(async (struct, overrides, feeOptions) => {
           let maxFeePerGas = (await struct.maxFeePerGas) ?? 0n;
           let maxPriorityFeePerGas = (await struct.maxPriorityFeePerGas) ?? 0n;
 
           // but if user is bypassing paymaster to fallback to having the account to pay the gas (one-off override),
           // we cannot delegate gas estimation to the bundler because paymaster middleware will not be called
           if (overrides?.paymasterAndData != null) {
-            const result = await fallbackFeeDataGetter(struct, overrides);
+            const result = await fallbackFeeDataGetter(
+              struct,
+              overrides,
+              feeOptions
+            );
             maxFeePerGas = (await result.maxFeePerGas) ?? maxFeePerGas;
             maxPriorityFeePerGas =
               (await result.maxPriorityFeePerGas) ?? maxPriorityFeePerGas;
@@ -153,7 +161,7 @@ const withAlchemyGasAndPaymasterAndDataMiddleware = <P extends AlchemyProvider>(
   provider: P,
   config: AlchemyGasManagerConfig
 ): Parameters<AlchemyProvider["withPaymasterMiddleware"]>["0"] => ({
-  paymasterDataMiddleware: async (struct, overrides) => {
+  paymasterDataMiddleware: async (struct, overrides, feeOptions) => {
     const userOperation: UserOperationRequest = deepHexlify(
       await resolveProperties(struct)
     );
@@ -167,9 +175,9 @@ const withAlchemyGasAndPaymasterAndDataMiddleware = <P extends AlchemyProvider>(
       }
 
       // provider level fee options with percentage
-      if (isPercentage(provider.feeOptions?.[field])) {
+      if (isPercentage(feeOptions?.[field])) {
         return {
-          percentage: 100 + Number(provider.feeOptions?.[field]?.percentage),
+          percentage: 100 + Number(feeOptions?.[field]?.percentage),
         };
       }
 
--- a/packages/core/src/provider/base.ts
+++ b/packages/core/src/provider/base.ts
@@ -541,24 +541,40 @@ export class SmartAccountProvider<
       feeOptions
     );
 
-    const request = deepHexlify(await resolveProperties(struct));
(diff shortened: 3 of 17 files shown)

### CODE REVIEW

File: packages/alchemy/src/middleware/client.ts [hunk 1]
```diff
@@ -26,9 +27,12 @@ export type ClientWithAlchemyMethods = PublicErc4337Client & {
             entryPoint: Address;
             userOperation: UserOperationRequest;
             dummySignature: Hex;
-            feeOverride?: {
-              maxFeePerGas: Hex;
-              maxPriorityFeePerGas: Hex;
+            overrides?: {
```
moldy530, line 30: are all of the `override` fields required if the object exists? or is it possible to only override individual items?

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -112,13 +112,16 @@ const withAlchemyGasAndPaymasterAndDataMiddleware = (
...
+    const overrides =
+      userOperation.maxFeePerGas && BigInt(userOperation.maxFeePerGas) > 0n
+        ? {
+            maxFeePerGas: userOperation.maxFeePerGas,
+            maxPriorityFeePerGas: userOperation.maxPriorityFeePerGas,
+            callGasLimit: userOperation.callGasLimit,
+            verificationGasLimit: userOperation.verificationGasLimit,
+            prevVerificationGas: userOperation.preVerificationGas,
+          }
+        : undefined;
```
moldy530, line 124: if in the above, it's possible to override individual items, then we need to check each field individually to make sure they're greater > 0 because the alchemy gas manager middleware sets all the gas + fee fields to 0.

If someone overrides just the fee fields, then we'd be setting the gas limits to 0 and the request will fail

packages/core/src/utils/types.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type NumericRange<
+  Arr extends number[],
+  End extends number,
+  Acc extends number = never
+> = Arr["length"] extends End
+  ? Acc | End
+  : NumericRange<[...Arr, 1], End, Acc | Arr["length"]>;
+
+export type Percentage = { percentage: NumericRange<ComputeRange<1>, 1000> };
```
moldy530, line 14: should percentage go up to `1000` or should that be `100`?

packages/alchemy/src/middleware/client.ts
```diff
@@ -26,9 +27,12 @@ export type ClientWithAlchemyMethods = PublicErc4337Client & {
...
+              maxFeePerGas: Hex | Percentage;
```
moldy530, line 31: oh interesting,  is this `percentage` type the way the alchemy RPC method expects it? I thought it took a `Hex` string or `number`

packages/alchemy/src/middleware/client.ts
```diff
@@ -1,4 +1,5 @@
 import {
+  type Percentage,
```
avasisht23, line 2: nit: might need to lint this file. 
image

packages/core/src/types.ts
```diff
@@ -19,12 +20,13 @@ export interface UserOperationCallData {
 
 export type BatchUserOperationCallData = UserOperationCallData[];
 
-export type UserOperationOverrides = Partial<
-  Pick<
-    UserOperationStruct,
-    "maxFeePerGas" | "maxPriorityFeePerGas" | "paymasterAndData"
-  >
->;
+export type UserOperationOverrides = Partial<{
```
avasisht23, line 23: can we avoid dupes with `UserOperationStruct` by making this an `Omit` or `Pick` on `UserOperationStruct` again?

packages/core/src/provider/base.ts
```diff
@@ -482,7 +482,7 @@ export class SmartAccountProvider<
     const maxFeePerGasBid =
       BigInt(feeData.maxFeePerGas) -
       BigInt(feeData.maxPriorityFeePerGas) +
-      maxPriorityFeePerGasBid;
+      maxPriorityFeePerGasBid!;
```
avasisht23, line 485: nit: can't this be undefined? how does the math work here if undefined? should it be:
```suggestion
      maxPriorityFeePerGasBid ?? 0n;
```

avasisht23 (commented):
few q's

packages/alchemy/src/middleware/client.ts
denniswon: @moldy530 this is a draft. don't review yet. will give ping for asking review soon.

denniswon:
@avasisht23 @moldy530 this PR is now ready for review. Requires some close look and expecting discussion.

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -112,13 +112,16 @@ const withAlchemyGasAndPaymasterAndDataMiddleware = (
...
     );
 
-    let feeOverride = undefined;
-    if (userOperation.maxFeePerGas && BigInt(userOperation.maxFeePerGas) > 0n) {
-      feeOverride = {
-        maxFeePerGas: userOperation.maxFeePerGas,
-        maxPriorityFeePerGas: userOperation.maxPriorityFeePerGas,
-      };
-    }
+    const overrides =
```
moldy530, line 115: Shouldn't this use applyOverrides? Right now it doesn't account for the cases I brought up in the first review

denniswon: ah right. forgot to update here. will update

packages/core/src/utils/schema.ts
```diff
@@ -18,3 +18,12 @@ export const ChainSchema = z.custom<Chain>((chain) => {
...
   }
 });
+
+export const PercentageSchema = z.object({
+  /**
+   * Percent value between 1 and 1000 inclusive for maxPriorityFeePerGas estimate added buffer.
+   * maxPriorityFeePerGasBid is set to the max between the buffer "added" priority fee estimate and
+   * the minPriorityFeePerBid (default: 33)
+   */
+  percentage: z.number().min(1).max(1000).default(33),
```
dancoombs, line 28: Does this require a default? Is this required for backwards compatibility? This schema is used for more fields than the fees now and a 33% default doesn't make sense for each value.

packages/core/src/types.ts
```diff
@@ -19,12 +22,18 @@ export interface UserOperationCallData {
...
```
dancoombs, line 25: This is starting to leak alchemy specific schemas into `aa-core`. The way that we structured the overrides in `alchemy_requestGasAndPaymasterAndData` is non-standard. Is there a way to structure this such that our `aa-core` interfaces stay general, and we can apply this specific schema only in the alchemy provider?

moldy530: yea maybe we should make the methods that accept `overrides` fields in `aa-core` be generic on the overrides like:

`methodWithOverrides<O extends BaseOverrides>(uo: UoThingy, overrides: O)`

then in aa-alchemy we can make it like `methodWithOverrides<O extends BaseOverrides = AlchemyOverrides>(...)` wonder if it's possible to even make the signature be `methodWithOverrides(uo: UoThingy, overrides: AlchemyOverrides)` in aa-alchemy and have typescript infer the rest

packages/core/src/utils/schema.ts
moldy530: yea agree with @dancoombs here, let's not add a default on the percentage schema. instead places that use the schema can specify the default themselves

also same question as earlier (i may have missed your comment), but why is the `max(1000)`? i imagine it should either be `100` or unbounded

packages/alchemy/src/middleware/gas-manager.ts
denniswon: oh so throughout the pipeline, we should not apply the percentage overrides and just pass through until here right?

packages/core/src/provider/base.ts
denniswon: I don't think this can be undefined right?

packages/core/src/utils/schema.ts
denniswon: @moldy530 per https://www.notion.so/alchemotion/2023-08-10-Alchemy-Gas-and-Paymaster-Update-020b41bee84641db98210cb71bca96af?pvs=4#b791899b4f7242c998cf5522fa734bb3

packages/alchemy/src/middleware/gas-manager.ts
denniswon: So when the override is percentage, then we should NOT call the apis for estimating gas fee, etc but just pass through with the percentage correct? @moldy530

denniswon: Currently for percentage, the middleware pipeline just applies the percentage on the estimates set from the feeDataGetter and gasEstimator, not just for alchemy provider but also for base provider. So TLDR, we would want to only make override params accept percentage types only for alchemy provider while the base provider experience stays the same right?

packages/core/src/utils/schema.ts
denniswon: synced offline with @moldy530 . To avoid not making the aa-sdk providers code not too messy with overused generics, will leave it as introducing these percentage types at the base level, at least for now. cc: @dancoombs @moldy530

packages/alchemy/src/middleware/gas-manager.ts
moldy530: I think we chatted some offline about this, but I think we should apply overrides as percent or constants during the typical override location (right before paymaster and data) in the base provider. 

Having a percent override might actually make sense even in non alchemy cases if they're using default fee estimator and they just want to tweak its output with a buffer.

The alchemy provider will be fine with this logic because it only accepts overrides if the value is >0 otherwise it doesn't pass it as an override to the server (or shouldn't)

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -112,13 +112,16 @@ const withAlchemyGasAndPaymasterAndDataMiddleware = (
...
+      userOperation.maxFeePerGas && BigInt(userOperation.maxFeePerGas) > 0n
+        ? {
+            maxFeePerGas: userOperation.maxFeePerGas,
+            maxPriorityFeePerGas: userOperation.maxPriorityFeePerGas,
+            callGasLimit: userOperation.callGasLimit,
+            verificationGasLimit: userOperation.verificationGasLimit,
+            preVerificationGas: userOperation.preVerificationGas,
```
moldy530, line 122: overrides should be applied on a per field basis using similar logic: if the UO has the field set to a value > 0 then override the field otherwise leave as undefined

packages/core/src/utils/schema.ts
```diff
@@ -18,3 +18,52 @@ export const ChainSchema = z.custom<Chain>((chain) => {
...
+     * Defaults 5% on Arbitrum and Optimism, 0% elsewhere
+     *
+     * This is only useful on Arbitrum and Optimism, where the preVerificationGas is
+     * dependent on the gas fee during the time of estimation. To improve chances of
+     * the UserOperation being mined, users can increase the preVerificationGas by
+     * a buffer. This buffer will always be charged, regardless of price at time of mine.
+     *
+     * NOTE: this is only applied if the defualt gas estimator is used.
+     */
+    preVerificationGasBufferPercent: z.bigint().optional(),
```
moldy530, line 61: were these part of the object before? I thought these were provider configs?

packages/core/src/provider/base.ts
```diff
@@ -461,6 +468,36 @@ export class SmartAccountProvider<
...
+      overrides?.callGasLimit
+    );
+    struct.verificationGasLimit = applyOverride(
+      _struct.verificationGasLimit,
+      overrides?.verificationGasLimit
+    );
+    struct.preVerificationGas = applyOverride(
+      _struct.preVerificationGas,
+      overrides?.preVerificationGas
+    );
```
moldy530, line 496: any way we can make this more future proof? maybe using `Object.entries().forEach` or something?

otherwise each time a new override is added we have to make sure to deal with it here too

packages/core/src/types.ts
moldy530: chatted offline some as well, but thought about it some more and we have had requests for "how do I add a buffer?" the current proposal is very annoying (you have to get a reference to the old feeEstimatorMiddleware and then write a new one that calls the old one + applies percent buffer). Introducing the percentage as an option in the base simplifies this.

With that it in mind @denniswon can you update the docs where we talk about adding a buffer (cc @avarobinson did we write this doc actually, or was that just a convo you had offline with a customer?)

packages/alchemy/src/middleware/gas-manager.ts
denniswon: Agree @moldy530 I think it makes sense to have UserOperationOverrides type in aa-core, then apply the overrides from the core smart account provider, not just for alchemy provider.

@moldy530 but what we can also do is that for alchemy provider, we don't apply the percentage override but instead just passthrough just like the absolute value override because the alchemy gas and paymaster request data api do accept the percentage type as is.

cc: @dancoombs

packages/alchemy/src/middleware/gas-manager.ts
denniswon: these values here should be already overrides applied because we apply the overrides before the paymaster and data

packages/core/src/provider/base.ts
denniswon: sure. will make the change. That would be okay because the user operation overrides are subset fields of user operations.

packages/core/src/utils/schema.ts
denniswon: these changes are reverted. was just a wip commit. sorry for the confusion.

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -21,27 +28,54 @@ export interface AlchemyGasManagerConfig {
...
  * @param config - the alchemy gas manager configuration
- * @param estimateGas - if true, this will use `alchemy_requestGasAndPaymasterAndData` else will use `alchemy_requestPaymasterAndData`
+ * @param deleteGasEstimation - if true, this will use `alchemy_requestGasAndPaymasterAndData` else will use `alchemy_requestPaymasterAndData`
  * @returns the provider augmented to use the alchemy gas manager
  */
 export const withAlchemyGasManager = (
   provider: AlchemyProvider,
   config: AlchemyGasManagerConfig,
-  estimateGas: boolean = true
+  deleteGasEstimation: boolean
```
moldy530, line 37: nit: can we call this something like `withGasEstimation`? delete gas estimation is odd

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -21,27 +28,54 @@ export interface AlchemyGasManagerConfig {
...
         // no-op gas estimator
-        .withGasEstimator(async () => ({
-          callGasLimit: 0n,
-          preVerificationGas: 0n,
-          verificationGasLimit: 0n,
-        }))
+        .withGasEstimator(async (struct, overrides) => {
+          // but if user is bypassing paymaster to fallback to having the account to pay the gas (one-off override),
+          // we cannot delegate gas estimation to the bundler because paymaster middleware will not be called
+          if (overrides?.paymasterAndData !== undefined) {
```
moldy530, line 45:
```suggestion
          if (overrides?.paymasterAndData != null) {
```

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -107,18 +141,40 @@ const withAlchemyGasAndPaymasterAndDataMiddleware = (
...
-      feeOverride = {
-        maxFeePerGas: userOperation.maxFeePerGas,
-        maxPriorityFeePerGas: userOperation.maxPriorityFeePerGas,
-      };
-    }
+    const overrideField = (
+      field: keyof UserOperationFeeOptions
+    ): Hex | Percentage | undefined => {
+      // one-off absolute override
+      if (overrides?.[field] !== undefined) {
```
moldy530, line 153:
```suggestion
      if (overrides?.[field] != null) {
```

packages/alchemy/src/middleware/gas-manager.ts
moldy530: oh it's called `delegateGasEstimation` further down; let's call it that here too

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -21,27 +28,54 @@ export interface AlchemyGasManagerConfig {
...
-          maxFeePerGas: (await struct.maxFeePerGas) ?? 0n,
-          maxPriorityFeePerGas: (await struct.maxPriorityFeePerGas) ?? 0n,
-        }))
+        .withFeeDataGetter(async (struct, overrides) => {
+          let maxFeePerGas = (await struct.maxFeePerGas) ?? 0n;
+          let maxPriorityFeePerGas = (await struct.maxPriorityFeePerGas) ?? 0n;
+
+          // but if user is bypassing paymaster to fallback to having the account to pay the gas (one-off override),
+          // we cannot delegate gas estimation to the bundler because paymaster middleware will not be called
+          if (overrides?.paymasterAndData !== undefined) {
```
moldy530, line 67:
```suggestion
          if (overrides?.paymasterAndData != null) {
```

PR #229, opened 2023-11-08
feat: add magic signer to aa-signers

avasisht23:
# Pull Request Checklist

- [x] Did you add new tests and confirm existing tests pass? (`yarn test`)
- [x] Did you update relevant docs? (docs are found in the `site` folder, see guidleines below)
- [x] Do your commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [x] Does your PR title also follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [x] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples? (e.g. `feat!: breaking change`)
- [x] Did you run lint (`yarn lint:check`) and fix any issues? (`yarn lint:fix`)
- [x] Is the base branch you're merging into `development` and not `main`?

---

### Detailed summary
This PR introduces the `MagicSigner` class in the `@alchemy/aa-signers` package. It includes the following changes:

- Deleted `packages/signers/src/__tests__/signer.test.ts`
- Added `export { MagicSigner } from "./signer.js"` and `export type * from "./types.js"` in `packages/signers/src/index.ts`
- Added `"magic-sdk": "^21.3.0"` as an optional dependency in `packages/signers/package.json`
- Added new files `packages/signers/src/magic/types.ts` and `packages/signers/src/magic/signer.ts`
- Added new methods `getAddress`, `signMessage`, `signTypedData`, `authenticate`, and `getAuthDetails` in `packages/signers/src/magic/signer.ts`
- Added new files `site/packages/aa-signers/magic/getAddress.md`, `site/packages/aa-signers/magic/signMessage.md`, `site/packages/aa-signers/magic/getAuthDetails.md`, `site/packages/aa-signers/magic/constructor.md`, and `site/packages/aa-signers/magic/signTypedData.md`
- Updated `site/.vitepress/config.ts` to include documentation links for `Magic Signer`

> The following files were skipped due to too many changes: `site/packages/aa-signers/magic/signTypedData.md`, `yarn.lock`, `site/packages/aa-signers/magic/authenticate.md`, `site/packages/aa-signers/magic/introduction.md`, `packages/signers/src/magic/__tests__/signer.test.ts`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #228, opened 2023-11-08
feat: add aa-signers package

avasisht23:
# Pull Request Checklist

- [x] Did you add new tests and confirm existing tests pass? (`yarn test`)
- [x] Did you update relevant docs? (docs are found in the `site` folder, see guidleines below)
- [x] Do your commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [x] Does your PR title also follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [x] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples? (e.g. `feat!: breaking change`)
- [x] Did you run lint (`yarn lint:check`) and fix any issues? (`yarn lint:fix`)
- [x] Is the base branch you're merging into `development` and not `main`?

---

### Focus of this PR:
This PR focuses on upgrading the Node version, updating dependencies, adding new files, and making changes to existing files in the `@alchemy/aa-signers` package.

### Detailed summary:
- Upgraded Node version from 18.10 to 18.16
- Added `tsconfig.json` file
- Updated `deploy-site.yml` and `publish-package.yml` workflows to use Node version 18.16
- Added `tsconfig.build.json` file with updated compiler options
- Updated content and links in `index.md` and `contributing.md` files
- Added new files: `signer.test.ts`, `vitest.config.ts`, `vitest.config.e2e.ts`, `getting-started.md`, `index.md`, `config.ts`, `package.json`, `package-overview.md`
- Updated content in `README.md` and `package.json` files

> The following files were skipped due to too many changes: `site/overview/package-overview.md`, `packages/signers/CHANGELOG.md`, `yarn.lock`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #247, opened 2023-11-13
test: add tests for web3auth

avasisht23:
# Pull Request Checklist

- [ ] Did you update relevant docs? (docs are found in the `site` folder, see guidleines below)
- [ ] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples? (e.g. `feat!: breaking change`)
- [ ] Did you run lint (`yarn lint:check`) and fix any issues? (`yarn lint:fix`)
- [ ] Is the base branch you're merging into `development` and not `main`?


avasisht23:
* development
  * **PR #227** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/227?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #228** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/228?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #229** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #247** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/247?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/247?utm_source=stack-comment).

packages/signers/src/web3auth/__tests__/signer.test.ts
```diff
@@ -0,0 +1,125 @@
...
+          ]) as R;
+        case "personal_sign":
+          return Promise.resolve("0xtest") as R;
+        case "eth_signTypedData_v4":
+          return Promise.resolve("0xtest") as R;
+        default:
+          return Promise.reject(new Error("Method not found"));
+      }
+    },
+  } as any);
```
avasisht23, line 113: just because I didn't want to mock the other 15 methods on here.

packages/signers/src/web3auth/__tests__/signer.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    verifier: "verifier",
+    verifierId: "verifierId",
+    typeOfLogin: "typeOfLogin",
+    dappShare: "dappShare",
+    idToken: "idToken",
+    oAuthIdToken: "oAuthIdToken",
+    oAuthAccessToken: "oAuthAccessToken",
+  });
+
+  vi.spyOn(inner, "provider", "get").mockReturnValue({
```
avasisht23, line 98: needed to spy vs. setting the return a mock b/c of [this](https://github.com/Web3Auth/web3auth-web/blob/7bb3cfc0d492da4e05a1a217e5784e1385862376/packages/no-modal/src/noModal.ts#L124) lol #typescriptissecure

packages/signers/src/web3auth/__tests__/signer.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    const signMessage = signer.signMessage("test");
+    await expect(signMessage).rejects.toThrowErrorMatchingInlineSnapshot(
+      '"Not authenticated"'
+    );
+  });
+
+  it("should correctly authenticate", async () => {
+    const spy = vi.spyOn(AASignersModule, "Web3AuthSigner");
+    givenSigner();
+    expect(spy.mock.calls).toMatchInlineSnapshot("[]");
```
moldy530, line 19: shouldn't this vaidate something specific to authenticated state?

moldy530: looking at your other test cases, I think it'd be more valuable to add test cases for all of the methods that require authentication and validate that they correctly throw. 

Right now the tests below only test happy paths

packages/signers/src/web3auth/signer.ts
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  signTypedData = (params: SignTypedDataParams) => {
+    if (!this.signer) throw new Error("Not authenticated");
+
+    return this.signer.signTypedData(params);
+  };
+
+  authenticate = async (params: Web3AuthAuthenticationParams) => {
+    await params.initModal();
```
moldy530, line 54: nit: I think this is usually done when the web3auth sdk is initialized because it will inject the iframe into the page

packages/signers/src/web3auth/signer.ts
```diff
@@ -0,0 +1,74 @@
...
+    SmartAccountAuthenticator<
+      Web3AuthAuthenticationParams,
+      Web3AuthUserInfo,
+      Web3Auth
+    >
+{
+  inner: Web3Auth;
+  private signer: WalletClientSigner | undefined;
+
+  constructor({ inner }: { inner: Web3Auth }) {
```
moldy530, line 24: I think we can continue to improve the devx, by making this takes the params for the SDK instead.

if we want to support someone passing a web3auth instance we can make that an option as well. Maybe this should be like `Web3AuthSDKParams | { inner: Web3Auth}` or we can add a static method like `Web3AuthSigner.fromSDKInstance(inner: Web3Auth)`

packages/signers/src/web3auth/signer.ts
avasisht23: sorry is the recommendation then to remove this and have the developer define it? I'm just copying it from the example: https://accountkit.alchemy.com/smart-accounts/signers/web3auth.html

packages/signers/package.json
```diff
@@ -63,6 +65,8 @@
     "viem": "^1.16.2"
   },
   "peerDependencies": {
+    "@web3auth/base": "^7.1.0",
+    "@web3auth/modal": "^7.1.1",
     "magic-sdk": "^21.2.0"
```
denniswon, line 70: this will make our signers have multiple peer Dependencies that are actually not peer dependencies. Would there be any better way to manage this? Devs with this flow will have to ignore the other peer deps that are not relevant for them.

wdyt? @moldy530 @avasisht23

avasisht23: could you say more? why would signers need multiple peer dependencies?

It's up to the developer to install in their project the extra dependency needed for the specific signer they choose. and if they swap out, they would need to install/uninstall as necessary.

moldy530: yea I think this is fine, they would only need to install the peer deps they want. It's possible we can get away with having these as deps instead and then leave it to developer's tree-shaking set up to drop the packages they don't need, but that comes at the cost of increased `npm install` times in dev

packages/signers/src/web3auth/signer.ts
```diff
@@ -0,0 +1,79 @@
...
+      this.inner = params.inner;
+      return;
+    }
+
+    this.inner = new Web3Auth(params);
+  }
+
+  get signerType() {
+    return "web3auth";
+  }
```
moldy530, line 35: why this? instead of just having
```
readonly signerType = "web3auth"
```

as a property. if we want to disable setting, we should probably have a `set signerType() { throw new Error("nope.") }`

packages/signers/src/web3auth/signer.ts
moldy530: nm yea this is fine

packages/signers/src/web3auth/types.ts
```diff
@@ -0,0 +1,8 @@
+import type { UserInfo } from "@web3auth/base";
+
+export interface Web3AuthAuthenticationParams {
+  initModal: () => Promise<void>;
```
moldy530, line 4: nit: `init` instead because someone might be using web3auth core

packages/signers/package.json
avasisht23: not: making these optionalDependencies so it's not required to install. Think I might have conflated peer with optional and led to this confusion.

packages/signers/src/web3auth/signer.ts
```diff
@@ -0,0 +1,84 @@
+import {
+  WalletClientSigner,
+  type SignTypedDataParams,
+  type SmartAccountAuthenticator,
+} from "@alchemy/aa-core";
+import { Web3Auth, type Web3AuthOptions } from "@web3auth/modal";
```
moldy530, line 6: why do we require modal here? ideally this should work with `core` no?

site/snippets/web3auth.ts
```diff
@@ -1,29 +1,21 @@
...
+  });
 
-await web3auth.initModal();
+  await web3AuthSigner.authenticate({
+    init: async () => {
+      await web3AuthSigner.inner.initModal();
+    },
+    connect: async () => {
+      await web3AuthSigner.inner.connect();
+    },
```
moldy530, line 17: if we require `modal` to be installed can we provide these as the default params for `authenticate`?

site/smart-accounts/signers/web3auth.md
```diff
@@ -69,7 +73,7 @@ const provider = new AlchemyProvider({
   (rpcClient) =>
     new LightSmartContractAccount({
       chain,
-      owner: web3authSigner,
+      owner: new createWeb3AuthSigner(),
```
moldy530, line 76: don't think you want `new` here

moldy530: you do want to `await` though

moldy530 (approved):
there's a small bug in the docs, but LGTM otherwise

PR #248, opened 2023-11-14
feat: add initial skeleton for 6900 account support

moldy530:
# Pull Request Checklist

- [x] Did you update relevant docs? (docs are found in the `site` folder, see guidleines below)
- [x] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples? (e.g. `feat!: breaking change`)
- [x] Did you run lint (`yarn lint:check`) and fix any issues? (`yarn lint:fix`)
- [x] Is the base branch you're merging into `development` and not `main`?

---

## PR-Codex overview
This PR focuses on adding plugin configuration and management functionality to the smart contract account system. 

### Detailed summary
- Added plugin configuration files for multi-owner, session key, and token receiver plugins.
- Added plugin generation configurations for multi-owner, session key, and token receiver plugins.
- Updated the provider schema to include connection configuration options.
- Updated the account base class to include an `extend` method for extending the account functionality.
- Added types for plugin hooks and injected hooks information.
- Updated the wagmi.config.ts file to generate plugin files based on the plugin configurations.
- Updated the Alchemy schema to include the connection configuration schema.
- Updated the provider interface to include a generic type parameter for provider decorators.
- Updated the uninstallPlugin function to include a generic type parameter for the smart account provider.
- Added the IStandardExecutorAbi to the account base class.

> The following files were skipped due to too many changes: `packages/accounts/src/msca/plugin-manager/uninstallPlugin.ts`, `packages/accounts/src/msca/utils.ts`, `packages/core/src/provider/base.ts`, `packages/accounts/src/msca/account-loupe/decorator.ts`, `packages/accounts/src/msca/plugins/types.ts`, `packages/accounts/src/msca/plugin-manager/installPlugin.ts`, `packages/accounts/src/index.ts`, `examples/aa-simple-dapp/src/hooks/useAlchemyProvider.ts`, `packages/accounts/src/msca/abis/IAccountLoupe.ts`, `packages/accounts/src/msca/account-loupe/types.ts`, `packages/core/src/account/types.ts`, `packages/accounts/src/msca/abis/MultiOwnerMSCAFactory.ts`, `packages/accounts/src/msca/abis/IPluginManager.ts`, `packages/accounts/src/msca/abis/MultiOwnerTokenReceiverMSCAFactory.ts`, `packages/accounts/src/msca/multi-owner-account.ts`, `packages/accounts/src/msca/plugins/session-key/plugin.ts`, `packages/accounts/scripts/plugingen.ts`, `packages/accounts/src/msca/plugins/multi-owner/plugin.ts`, `packages/accounts/src/msca/builder.ts`, `packages/accounts/src/msca/plugins/token-receiver/plugin.ts`, `packages/accounts/src/msca/abis/IPlugin.ts`, `packages/accounts/plugindefs/token-receiver/abi.ts`, `packages/accounts/plugindefs/session-key/abi.ts`, `packages/accounts/plugindefs/multi-owner/abi.ts`, `yarn.lock`


packages/accounts/src/msca/plugins/multi-owner/signer.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.plugin = new MultiOwnerPlugin(pluginAddress);
+  }
+
+  // This returns the address of the current owner, not all owners
+  getAddress: () => Promise<`0x${string}`> = () => {
+    return this.inner.getAddress();
+  };
+
+  // NOTE: signing Userops like this will not work because UOs are signed the old way...
+  // TODO(jay + moldy): work with howy to do this consistently for UOs and 1271
```
moldy530, line 25: could have a special method here called `signMessagefor1271` here and use that in the account as well. signMessage can be used to sign UOs but devs would have to call `signMessageFor1271` manually for all other cases

packages/accounts/src/msca/plugins/multi-owner/signer.ts
```diff
@@ -0,0 +1,46 @@
...
+export class MultiOwnerPluginSigner
+  implements SmartAccountSigner<SmartAccountSigner>
+{
+  signerType: string;
+  inner: SmartAccountSigner<any>;
+  plugin: MultiOwnerPlugin;
+
+  // TODO(moldy) -- use a struct instead of args array
+  constructor(inner: SmartAccountSigner<any>, pluginAddress: string) {
+    this.signerType = inner.signerType;
```
moldy530, line 14: i think this should actually be something like `multi-owner:${inner.signerType}`

packages/accounts/src/msca/account.ts
```diff
@@ -0,0 +1,102 @@
...
+      createModularSmartContractAccountSchema<TTransport>().parse(params_);
+
+    super(params);
+
+    this.owner = params.owner;
+    this.index = params.index;
+  }
+
+  getDummySignature(): Hex {
+    throw new Error("Method not implemented.");
```
moldy530, line 51: for MSCA, I think this should actually be defined on the signer itself... which raises a question for @avasisht23 
Should we add `getDummySignature` as a method on the SmartAccountSigner impls? we could make it an optional lambda property for now or something and intro a breaking change later

but the dummy signature is something that is dependent on the type of signature validation the account does

packages/accounts/src/msca/plugins/multi-owner/plugin.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.contract = getContract({
+      address: this.pluginAddress,
+      publicClient: this.client as PublicClient<HttpTransport>,
+      abi: MultiOwnerPluginAbi,
+    });
+  }
+
+  async getDomainSeparator(): Promise<TypedDataDomain> {
+    const [, name, version, chainId, verifyingContract, salt] =
+      await this.contract.read.eip712Domain();
```
moldy530, line 46: @jaypaik looking at the contract code, I have a feeling I'm doing this wrong... the calling context to this method needs to be the SCA right? so that the verifying contract returned is correct?

packages/accounts/src/msca/plugins/multi-owner/signer.ts
```diff
@@ -0,0 +1,81 @@
...
+        message: {
+          ERC6900Message: {
+            message: typeof msg === "string" ? toBytes(msg) : msg,
+          },
+        },
+        primaryType: "ERC6900Message",
+      });
+    };
+
+  signTypedData: (params: SignTypedDataParams) => Promise<`0x${string}`> = (
```
moldy530, line 76: @jaypaik can you confirm this one too? I'm pretty sure this is wrong too. 712 signatures also need to be hashed and wrapped like the message above right? The multiowner plugin will take the digest being verified and encodes it, so I guess here that means we need to sign the result of `wrapWithEnvelope(hashTypedData(params))` where `wrapWithEnvelope` is what the code above is doing

packages/accounts/src/msca/base.ts
```diff
@@ -0,0 +1,59 @@
...
+>() =>
+  createBaseSmartAccountParamsSchema<TTransport>().extend({
+    owner: SignerSchema,
+  });
+
+export type ModularSmartContractAccountParams = z.input<
+  ReturnType<typeof createModularSmartContractAccountSchema>
+>;
+
+export abstract class BaseModularSmartContractAccount<
```
moldy530, line 29: rohan's gonna kill me....

I think we should explore some kind of Builder Pattern for MSCA that lets you generate new classes by doing something like:

```
type MSCA = ISmartContractAccount & MSCASpecificMethods // like withPlugin?

export const createMultiOwnerMSCA = <R extends ISmartContractAccount>({owner}: MultiOwnerMSCAParams): R => { 
  return ModularSmartContractAccountBuilder()
    .withStandardExecutor()
    .withPlugin(MultiOwnerPlugin)
    .withSignMethods(account => ({
      signMessage: ...
      signTypedData: ...
      signUserOperationHash: ...
    }))
}
```

moldy530: Actually after talking to @adam-alchemy and @howydev, we will probably want something like the above. Right now, the standard executor function is only callable in context of the root owner plugin. So for session key plugins or alternative signers, they will have their own methods to call. So that means that `executeBatch` is not appropriate the way it is defined here.

We'll want to introduce some kind of MSCA class builder that returns an MSCA class that can then be further extended with decorators (from plugin gen)

moldy530: prototyped this some here: https://github.com/alchemyplatform/aa-sdk/pull/264

packages/accounts/src/msca/plugins/types.ts
```diff
@@ -0,0 +1,6 @@
+import type { BaseSmartContractAccount } from "@alchemy/aa-core";
+
+export interface Plugin<D> {
+  meta: { name: string; version: string };
+  accountDecorators: (a: BaseSmartContractAccount) => D;
```
moldy530, line 5: we should expose a method or the `rpcProvider` property on `ISmartContractAccount` so we can use that here instead of `BaseSmartContractAccount`

packages/accounts/plugindefs/index.ts
```diff
@@ -0,0 +1 @@
+export { MultiOwnerPluginGenConfig } from "./multi-owner/config.js";
```
moldy530, line 1: we can probably do better here by just reading all the `config.js` files found within this folder and exporting their contents. we might need to do default exports for the configs though

denniswon: hmm but will take break the vscode module navigation ? or you mean reading all config.ts files and export their contents by writing into a new file for exports in `out` or something

packages/accounts/src/msca/multi-owner-account.ts
```diff
@@ -0,0 +1,124 @@
...
+            ERC6900Message: [{ name: "message", type: "bytes" }],
+          },
+          message: {
+            ERC6900Message: {
+              message: toBytes(msg),
+            },
+          },
+          primaryType: "ERC6900Message",
+        });
+      };
```
denniswon, line 83: this is signing with 1271?
can you point to the reference where the contract implements `function isValidSignature`? Wondering how you came up with the primaryType `ERC6900Message`. Is that degined in the `eip712Domain` function of `MultiOwnerPlugin`?

packages/accounts/src/msca/multi-owner-account.ts
```diff
@@ -0,0 +1,124 @@
...
+      return concatHex([
+        acct.getFactoryAddress(),
+        encodeFunctionData({
+          abi: MultiOwnerMSCAFactoryAbi,
+          functionName: "createAccount",
+          // TODO: this needs to support creating accounts with multiple owners
+          args: [params.index, [ownerAddress]],
+        }),
+      ]);
+    })
```
denniswon, line 55: so `MultiOwnerMSCAFactory` is always the factory that create MSCAs? There isn't a separate `MSCAFactory` right?

packages/accounts/src/msca/multi-owner-account.ts
```diff
@@ -0,0 +1,124 @@
...
+import { MultiOwnerMSCAFactoryAbi } from "./abis/MultiOwnerMSCAFactory.js";
+import { MSCABuilder, StandardExecutor } from "./builder.js";
+import { MultiOwnerPlugin } from "./plugins/multi-owner.js";
+
+export const createMultiOwnerMSCASchema = <
+  TTransport extends SupportedTransports = Transport
+>() =>
+  createBaseSmartAccountParamsSchema<TTransport>().extend({
+    owner: SignerSchema,
+    index: z.bigint().optional().default(0n),
```
denniswon, line 29: what does this `index` specify?

packages/accounts/src/msca/builder.ts
```diff
@@ -0,0 +1,140 @@
...
+) => Pick<ISmartContractAccount, "encodeExecute" | "encodeBatchExecute">;
+
+export type SignerMethods = <A extends MSCA>(
+  acct: A
+) => Pick<
+  ISmartContractAccount,
+  | "signMessage"
+  | "signTypedData"
+  | "signUserOperationHash"
+  | "getDummySignature"
```
denniswon, line 34: these are abstraction functions except the execute functions from SmartContractAccount correct?

Then why also `signUserOperationHash`? Don't we just use `signMessage` for signing uos?

packages/accounts/plugindefs/types.ts
```diff
@@ -0,0 +1,10 @@
+import type { Abi, Address, Chain } from "viem";
+
+export type PluginGenConfig = {
+  abi: Abi;
+  address: Address;
+  chain: Chain;
+  name: string;
+  // TODO: need to make this configurable to run in CI without requiring this
+  rpcUrl?: string;
```
denniswon, line 9: what do you mean configurable here?

packages/accounts/src/msca/multi-owner-account.ts
moldy530: https://github.com/OMGWINNING/account-contracts/blob/master/src/plugins/owner/MultiOwnerPlugin.sol#L136

All signatures for 1271 validation in the MSCA (and LightAccount) require this wrapping

packages/accounts/src/msca/multi-owner-account.ts
moldy530: There can be any number of factories for an MSCA which is why I added this `withFactory` concept. The other factory that exists already is the MultiOwnerWithTokenReceiver factory

packages/accounts/src/msca/multi-owner-account.ts
moldy530: index is used as a salt, if you want to create many accounts with the same set of initial owners but have them be at other addresses use can use the salt to differentiate them

packages/accounts/src/msca/builder.ts
moldy530: for 1271 validation, we sign messages differently that we do for UO hashes. The MSCA (and LightAccount) expects the UO to be personal sign of the hash. 1271 expects a sign typed data following the envelope defined above

packages/accounts/plugindefs/types.ts
moldy530: I think I forgot to remove the TODO because it is optional now

moldy530: the idea is that we shouldn't have to hardcode an RPC URL here. This is useful for testing locally if you want to hit a local anvil instance

.vscode/settings.json
avasisht23, line 1: is this file necessary?

packages/accounts/package.json
```diff
@@ -29,6 +29,8 @@
     "./package.json": "./package.json"
   },
   "scripts": {
+    "generate": "npx wagmi generate",
```
avasisht23, line 32: learn: have you worked with the wagmi cli? if not / maybe if you remember when you first did, did you just look up a way to generate abis and some stackoverflow pointed you to it? 

otherwise, curious how you arrived at this solution.

packages/accounts/src/msca/plugins/types.ts
avasisht23: by having made it readonly you can now, right?

packages/accounts/scripts/plugingen.ts
avasisht23, line 1: should the file be "plugin-gen.ts" to be kebab case?

packages/accounts/plugindefs/types.ts
avasisht23: is there a better way to write this type so it's dependent on environment to see if rpcUrl is defined?

packages/accounts/wagmi.config.ts
```diff
@@ -0,0 +1,25 @@
...
+import dotenv from "dotenv";
+import * as configs from "./plugindefs/index.js";
+import { plugingen } from "./scripts/plugingen.js";
+dotenv.config();
+
+const pluginConfigs = Object.values(configs);
+const pluginRegEx: RegExp = /[pP]lugin/g;
+
+export default defineConfig(
+  pluginConfigs.map((config) => ({
```
avasisht23, line 12: nit: is there a way to check if there's no diff to the generated file? this re-writes every plugin auto-gen file each time the yarn script is run, right?

packages/accounts/plugindefs/multi-owner/config.ts
```diff
@@ -0,0 +1,10 @@
+import { sepolia } from "viem/chains";
+import type { PluginGenConfig } from "../types";
+import { MultiOwnerPluginAbi } from "./abi.js";
+
+export const MultiOwnerPluginGenConfig: PluginGenConfig = {
+  name: "MultiOwnerPlugin",
+  abi: MultiOwnerPluginAbi,
+  address: "0x56bC629F342821FBe91C5273880792dFECBE7920",
```
avasisht23, line 8: nit: should we make this a util method i.e. `getDefaultMSCAFactoryAddress(chain)`

packages/accounts/scripts/plugingen.ts
```diff
@@ -0,0 +1,151 @@
...
+  http,
+  type AbiItem,
+  type Chain,
+} from "viem";
+import { IPluginAbi } from "../src/msca/abis/IPlugin.js";
+
+type RequiredBy<TType, TKeys extends keyof TType> = Required<
+  Pick<TType, TKeys>
+> &
+  Omit<TType, TKeys>;
```
avasisht23, line 17: seems useful, wanna put it in utils somewhere?

packages/accounts/src/msca/abis/IPlugin.ts
avasisht23, line 1: nit: want to include a link to the repo + etherscan / some way to validate the abi?

packages/accounts/src/msca/abis/IStandardExecutor.ts
avasisht23, line 1: nit: want to include a link to the repo + etherscan / some way to validate the abi?

packages/accounts/src/msca/abis/MultiOwnerMSCAFactory.ts
avasisht23, line 1: nit: want to include a link to the repo + etherscan / some way to validate the abi?

also do we want to call this MultiOwnerMSCA if this is the only MSCA we'll implement? will we have SingleOwner? I thought that was out of scope.

packages/accounts/plugindefs/index.ts
avasisht23, line 1: nit: should this folder be `plugin-defs`?

avasisht23 (commented):
would like to review the builder + multi-owner-account files live as a team, but left comments elsewhere

packages/accounts/package.json
moldy530: I used it before, but then looked at their codebase for how they define plugins so I could write my own

packages/accounts/plugindefs/index.ts
moldy530: yea I think that's a fair change

packages/accounts/plugindefs/index.ts
moldy530: yea the way I was thinking would break module navigation good call... we could make the plugin gen script overwrite this file though that's a good idea

packages/accounts/plugindefs/multi-owner/config.ts
moldy530: this is a different address, it's the plugin address and we will only ever use one chain here. Also until this is in production we shouldn't introduce utils for getting default addresses

packages/accounts/src/msca/abis/IPlugin.ts
moldy530: replied elsewhere too, but let's keep those out of here for now since none of these things are in production yet

packages/accounts/src/msca/abis/MultiOwnerMSCAFactory.ts
moldy530: There's also MultiOwnerWithTokenReceiverFactory, others may add their own MSCA factories here potentially and we might as well (like maybe one that includes the session key plugin later). We should just match the names of the contracts these ABIs belong to

packages/accounts/src/msca/plugins/types.ts
moldy530: I did but not on the ISmartContractAccount, I can update the ISmartContractAccount interface here if we're cool with that change

packages/accounts/wagmi.config.ts
moldy530: yea it does and we probs can, but we can probs leverage `nx` and `lerna` to do that by adding an `nx` config for `generate`

avasisht23:
* development
  * **PR #248** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/248?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/248?utm_source=stack-comment).

packages/accounts/scripts/plugingen.ts
denniswon: pushed the change

packages/accounts/src/msca/base.ts
moldy530, line 1: TODO: we need to delete this file because we don't use it

denniswon (approved):

denniswon (approved):

PR #272, opened 2023-11-21
feat: allow passing raw call data to sendUserOperation

moldy530:
# Pull Request Checklist

- [ ] Did you update relevant docs? (docs are found in the `site` folder, see guidleines below)
- [ ] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples? (e.g. `feat!: breaking change`)
- [ ] Did you run lint (`yarn lint:check`) and fix any issues? (`yarn lint:fix`)
- [ ] Is the base branch you're merging into `development` and not `main`?

---

### Focus of the PR:
This PR focuses on modifying the `UserOperationCallData` type to allow for an additional property `Hex`, and updating related code to accommodate this change.

### Detailed summary:
- Modified `UserOperationCallData` type to include an additional property `Hex`
- Updated `KernelUserOperationCallData` type to exclude `Hex`
- Updated `KernelBatchUserOperationCallData` type to exclude `Hex`
- Updated `BatchUserOperationCallData` type to exclude `Hex`
- Updated `sendUserOperation.md` guide to include `Hex` as a possible parameter type

### DIFF
--- a/packages/accounts/src/kernel-zerodev/types.ts
+++ b/packages/accounts/src/kernel-zerodev/types.ts
@@ -1,7 +1,11 @@
 import type { UserOperationCallData } from "@alchemy/aa-core";
+import type { Hex } from "viem";
 
-export interface KernelUserOperationCallData extends UserOperationCallData {
+export type KernelUserOperationCallData = Exclude<
+  UserOperationCallData,
+  Hex
+> & {
   delegateCall?: boolean;
-}
+};
 
 export type KernelBatchUserOperationCallData = KernelUserOperationCallData[];
--- a/site/packages/aa-core/provider/sendUserOperation.md
+++ b/site/packages/aa-core/provider/sendUserOperation.md
@@ -65,7 +65,9 @@ A Promise containing the hash of the user operation and the request that was sen
 
 ## Parameters
 
-### `UserOperationCallData | UserOperationCallData[]`
+### `UserOperationCallData | UserOperationCallData[] | Hex`
+
+`UserOperationCallData` is an object with the following properties:
 
 - `target: Address` - the target of the call (equivalent to `to` in a transaction)
 - `data: Hex` - can be either `0x` or a call data string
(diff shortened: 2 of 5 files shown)


denniswon (commented):
we should update the sendUO doc

moldy530:
@denniswon docs are updated

packages/accounts/src/kernel-zerodev/types.ts
```diff
@@ -1,7 +1,11 @@
...
+import type { Hex } from "viem";
 
-export interface KernelUserOperationCallData extends UserOperationCallData {
+export type KernelUserOperationCallData = Exclude<
+  UserOperationCallData,
+  Hex
+> & {
   delegateCall?: boolean;
-}
+};
```
avasisht23, line 9: oh so does Exclude only work when `  UserOperationCallData` is the `Hex` type alias, vs. the calldata object of target, data, value?

site/packages/aa-core/provider/sendUserOperation.md
```diff
@@ -65,7 +65,9 @@ A Promise containing the hash of the user operation and the request that was sen
 
 ## Parameters
 
-### `UserOperationCallData | UserOperationCallData[]`
+### `UserOperationCallData | UserOperationCallData[] | Hex`
+
+`UserOperationCallData` is an object with the following properties:
```
avasisht23, line 70:
```suggestion
`UserOperationCallData` is either a `Hex`value, or an object with the following properties:
```

avasisht23 (approved):
two nits, lgtm

site/packages/aa-core/provider/sendUserOperation.md
moldy530: Ah, I didn't want to confuse that here actually... I add the `| Hex` explicitly already above to make that clearer. because otherwise it might be the case that you can do `Hex[]` as batch (which is not the case)

packages/accounts/src/kernel-zerodev/types.ts
moldy530: Other way around, the use of `Exclude` here means take all the union types of `UserOperationCallData` and omit the `Hex` one. Essentially: `Exclude<a | b, b> = a`

PR #276, opened 2023-11-24
feat: aa-core smart account provider fee options to handle more systematic fee options for userops

denniswon:
per https://app.asana.com/0/1205598840815267/1205909484815093/f

this task is a ground work for supporting more advanced ways to enable/manage gas estimates during the user operation creation from the sdk.

This PR specifically defines and implements the smart account provider behavior when populating the gas fields of the user operation, enabling more stress-free and less error-prone experience for developers.

Tested with e2e tests that used to be more flaky, but fixed to be rather consistent on the PR https://github.com/alchemyplatform/aa-sdk/pull/243

---

## PR-Codex overview
This PR focuses on updating the gas estimation and fee options in the Alchemy codebase. 

### Detailed summary:
- Renamed `estimateGas` to `delegateGasEstimation` in `withAlchemyGasManager.md`
- Removed `FeeOpts` type in `type.ts`
- Added `RequestGasAndPaymasterAndDataOverrides` type in `index.ts`
- Updated `ClientWithAlchemyMethods` type in `client.ts`
- Updated log level in `logger.ts`
- Added utility functions in `index.ts`
- Added new types in `types/index.ts`
- Updated `AccountMiddlewareFn` and related types in `provider/types.ts`
- Updated schema in `schema.ts`
- Added utility function in `bigint.ts`

> The following files were skipped due to too many changes: `packages/core/src/utils/bigint.ts`, `packages/alchemy/src/middleware/gas-fees.ts`, `packages/core/src/utils/index.ts`, `packages/core/src/provider/schema.ts`, `packages/accounts/src/light-account/e2e-tests/light-account.test.ts`, `packages/core/e2e-tests/simple-account.test.ts`, `packages/alchemy/src/provider.ts`, `packages/alchemy/src/middleware/gas-manager.ts`, `packages/alchemy/e2e-tests/simple-account.test.ts`, `packages/core/src/provider/base.ts`


packages/core/src/provider/types.ts
```diff
@@ -78,9 +81,28 @@ export type PaymasterAndDataMiddleware = AccountMiddlewareOverrideFn<
...
+>;
+
 export type FeeDataMiddleware = AccountMiddlewareOverrideFn<
   "maxFeePerGas" | "maxPriorityFeePerGas"
 >;
+export type FeeDataFeeOptions = Partial<
+  Pick<UserOperationFeeOptions, "maxFeePerGas" | "maxPriorityFeePerGas">
+>;
+
+export type FeeOptionsMiddleware = AccountMiddlewareOverrideFn<
```
moldy530, line 98: how is this middleware used? feels like it combines the fee estimator and gas estimator middlewares? Do we really need it? I'm concerned this becomes confusing for devs

packages/core/src/schema.ts
moldy530, line 1: shouldn't this be with `provider/schema`?

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -13,35 +13,65 @@ export interface AlchemyGasManagerConfig {
...
         // no-op gas estimator
-        .withGasEstimator(async () => ({
-          callGasLimit: 0n,
-          preVerificationGas: 0n,
-          verificationGasLimit: 0n,
-        }))
+        .withGasEstimator(async (struct, overrides) => {
+          // but if user is bypassing paymaster to fallback to having the account to pay the gas (one-off override),
+          // we cannot delegate gas estimation to the bundler because paymaster middleware will not be called
+          if (overrides?.paymasterAndData !== undefined) {
```
moldy530, line 41: I think we shouldn't allow any `paymasterAndData` override here and only allow someone to set it to `0x` (as in skip paymaster for this UO) if the paymaster and data expects a specific UO hash, then reestimating gas will result in a broken UO.

packages/core/src/provider/base.ts
```diff
@@ -508,36 +523,66 @@ export class SmartAccountProvider<
...
+      (await estimateMaxFeePerGas(struct.maxPriorityFeePerGas));
+
+    return struct;
+  };
 
-    const maxFeePerGasBid =
-      BigInt(feeData.maxFeePerGas) -
-      BigInt(feeData.maxPriorityFeePerGas) +
-      maxPriorityFeePerGasBid;
+  readonly feeOptionsMiddleware: AccountMiddlewareFn = async (
```
moldy530, line 560: fee options set on the provider were initially only meant to be used with the default gas estimator and default fee estimator. This was to allow someone to configure the bounds we use without having to write your own estimators. I'm not sure we need to support customizing them via a middleware though. IF someone overrides feeDataGetter or gasEstimator, then they would add their own custom logic and we shouldn't tightly couple overriding `feeDataGetter` and `feeOptionsMiddleware`. I think this creates a weird experience. EG:

Dev overrides feeDataGetter with their custom logic. Dev doesn't override feeOptionsMiddleware. Dev writes test for their override and expects values based on their implementation. Dev gets different values because the `feeOptionsMiddleware` affects those values

packages/alchemy/src/middleware/gas-manager.ts
avasisht23, line 41:
```suggestion
          if (overrides?.paymasterAndData != null) {
```

packages/core/src/provider/types.ts
denniswon: discussed offline

packages/alchemy/src/middleware/gas-manager.ts
denniswon: discussed offline

packages/core/src/utils/index.ts
```diff
@@ -89,6 +97,24 @@ export function deepHexlify(obj: any): any {
   );
 }
 
+export function applyFeeOption(
+  value: BigNumberish | undefined,
+  feeOption: UserOperationFeeOptionsField | undefined
+): BigNumberish {
+  if (feeOption === undefined) {
+    return value ?? 0n;
+  }
```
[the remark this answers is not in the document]
moldy530: +1 to this suggestion

packages/core/src/utils/schema.ts
```diff
@@ -18,3 +18,25 @@ export const ChainSchema = z.custom<Chain>((chain) => {
     return false;
   }
 });
+
+export const HexSchema = z.custom<`0x${string}` | "0x">((val) => {
```
moldy530, line 22: nit: we've been calling these `Schema` everywhere but the idiomatic zod way of defining these is as `z*` so here would be `zHex`

packages/core/src/provider/base.ts
```diff
@@ -309,9 +313,13 @@ export class SmartAccountProvider<
       await new Promise((resolve) =>
         setTimeout(resolve, txRetryIntervalWithJitterMs)
       );
-      const receipt = await this.getUserOperationReceipt(hash as `0x${string}`)
-        // TODO: should maybe log the error?
-        .catch(() => null);
+      const receipt = await this.getUserOperationReceipt(
+        hash as `0x${string}`
+      ).catch((e) => {
+        Logger.debug(
```
moldy530, line 319:
```suggestion
        Logger.error(
```

packages/core/src/provider/base.ts
```diff
@@ -487,59 +495,149 @@ export class SmartAccountProvider<
...
+      feeOptions
+    );
+
+    let { callGasLimit, verificationGasLimit, preVerificationGas } =
+      overrides ?? {};
+
+    if (
+      callGasLimit === undefined ||
+      verificationGasLimit === undefined ||
+      preVerificationGas === undefined
```
moldy530, line 561: nit, we should use `== null` when doing these checks. There's some peeps that don't use typescript so there's chance they might set these incorrectly

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -1,47 +1,101 @@
...
-          maxFeePerGas: (await struct.maxFeePerGas) ?? 0n,
-          maxPriorityFeePerGas: (await struct.maxPriorityFeePerGas) ?? 0n,
-        }))
+        .withFeeDataGetter(async (struct, overrides, feeOptions) => {
+          let maxFeePerGas = (await struct.maxFeePerGas) ?? 0n;
+          let maxPriorityFeePerGas = (await struct.maxPriorityFeePerGas) ?? 0n;
+
+          // but if user is bypassing paymaster to fallback to having the account to pay the gas (one-off override),
+          // we cannot delegate gas estimation to the bundler because paymaster middleware will not be called
+          if (overrides?.paymasterAndData != null) {
```
moldy530, line 83:
```suggestion
          if (overrides?.paymasterAndData === "0x") {
```

packages/core/src/provider/schema.ts
```diff
@@ -3,35 +3,32 @@ import type { Transport } from "viem";
...
-   * between the buffer "added" priority fee estimate and the minPriorityFeePerBid (default: 33)
-   */
-  maxPriorityFeePerGasEstimateBuffer: z.number().min(0).optional(),
-});
+    /**
+     * Optional user operation fee options to be set globally at the provider level
+     */
+    feeOptions: UserOperationFeeOptionsSchema.optional(),
+  })
+  .strict();
```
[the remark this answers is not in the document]
denniswon: If there are any unknown keys in the input, Zod will throw an error. This is intended. And I do not intend to catch the error. For `is<...>` methods I use safeParse to see if the it is strictly that type

packages/core/src/provider/base.ts
```diff
@@ -405,10 +409,10 @@ export class SmartAccountProvider<
         BigInt(maxPriorityFeePerGas ?? 0n),
         bigIntPercent(uoToDrop.maxPriorityFeePerGas, 110n)
       ),
-      paymasterAndData: uoToDrop.paymasterAndData,
     };
 
     const uoToSend = await this._runMiddlewareStack(uoToSubmit, _overrides);
+
     return this._sendUserOperation(uoToSend);
```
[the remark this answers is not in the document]
denniswon: yes

packages/core/src/utils/schema.ts
denniswon: to do it to all schemas on a separate PR

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -1,47 +1,101 @@
...
         // no-op gas estimator
-        .withGasEstimator(async () => ({
-          callGasLimit: 0n,
-          preVerificationGas: 0n,
-          verificationGasLimit: 0n,
-        }))
+        .withGasEstimator(async (struct, overrides, feeOptions) => {
+          // but if user is bypassing paymaster to fallback to having the account to pay the gas (one-off override),
+          // we cannot delegate gas estimation to the bundler because paymaster middleware will not be called
+          if (overrides?.paymasterAndData != null) {
```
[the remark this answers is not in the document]
denniswon: hmmm is this necessary? if so, then we should just make the `paymasterAndData` field as `0x` only. not Hex. We can do this change on a separate PR.

packages/alchemy/src/middleware/gas-manager.ts
denniswon: ditto

packages/core/src/utils/schema.ts
moldy530: sounds good!

packages/alchemy/src/middleware/gas-manager.ts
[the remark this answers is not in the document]
moldy530: I think we should handle this as part of this PR. overriding to something other than `0x` doesn't work in all cases. Most notably if the paymaster and data is based on a specific UO hash. This isn't the case for all paymasters probably, but for our paymaster it is. So for our gas manager client side impl we should enforce that the only override we support is `0x`

packages/alchemy/src/middleware/gas-manager.ts
moldy530: see above

packages/alchemy/src/middleware/gas-fees.ts
```diff
@@ -1,33 +1,44 @@
...
+      overrides?.maxPriorityFeePerGas != null
+        ? overrides?.maxPriorityFeePerGas
+        : // it's a fair assumption that if someone is using this Alchemy Middleware, then they are using Alchemy RPC
+          applyFeeOption(
+            await (provider.rpcClient as ClientWithAlchemyMethods).request({
+              method: "rundler_maxPriorityFeePerGas",
+              params: [],
+            }),
+            feeOptions?.maxPriorityFeePerGas
+          );
```
avasisht23, line 19: seeing this confuses me a bit (and I imagine it would be confusing to a new dev). 

why do both `overrides` and `feeOptions` need to have a value for something like `maxPriorityFeePerGas`? how would I know without looking at the code that the value in `overrides` takes priority over the one in `feeOptions`?

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -106,19 +160,41 @@ const withAlchemyPaymasterAndDataMiddleware = <P extends AlchemyProvider>(
 const withAlchemyGasAndPaymasterAndDataMiddleware = <P extends AlchemyProvider>(
   provider: P,
   config: AlchemyGasManagerConfig
-): Parameters<P["withPaymasterMiddleware"]>["0"] => ({
-  paymasterDataMiddleware: async (struct) => {
+): Parameters<AlchemyProvider["withPaymasterMiddleware"]>["0"] => ({
```
avasisht23, line 163: this can't be `P`?

avasisht23 (commented):
some qs/nits

packages/alchemy/src/middleware/gas-manager.ts
denniswon: ah no should be just P. will update.

packages/alchemy/src/middleware/gas-fees.ts
denniswon: we should indicate on the documentation but overrides is a one-time overrides, while feeOptions is at a class level globally. So makes sense for overrides to be override the feeoptions on the provider.

packages/alchemy/src/middleware/gas-fees.ts
```diff
@@ -1,33 +1,44 @@
...
 
-    const baseFeeIncrease =
-      (baseFeePerGas * (100n + baseFeeBufferPercent)) / 100n;
-    const prioFeeIncrease =
-      (priorityFeePerGas * (100n + maxPriorityFeeBufferPercent)) / 100n;
+    const estimateMaxFeePerGas = async (priorityFeePerGas: BigNumberish) => {
```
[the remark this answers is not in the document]
denniswon: more readable

packages/alchemy/e2e-tests/simple-account.test.ts
```diff
@@ -231,43 +242,44 @@ describe("Simple Account Tests", () => {
...
+  }, 100000);
 
   it("should correctly simulate asset changes for the user operation", async () => {
     const provider = givenConnectedProvider({
       owner,
       chain,
     });
 
-    const simulatedAssetChanges =
+    const simulatedAssetChanges: SimulateUserOperationAssetChangesResponse =
```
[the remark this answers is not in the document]
denniswon: not necessary. will revert.

packages/alchemy/e2e-tests/simple-account.test.ts
```diff
@@ -284,33 +296,33 @@ describe("Simple Account Tests", () => {
...
       txMaxRetries: 10,
+      feeOptions: {
+        ...feeOptions,
+        maxFeePerGas: { percentage: 50 },
+        maxPriorityFeePerGas: { percentage: 5 },
+      },
     },
   }).connect(
-    (provider) =>
+    (provider: PublicErc4337Client) =>
```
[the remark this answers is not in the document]
denniswon: nope. will revert.

packages/core/src/utils/index.ts
```diff
@@ -26,11 +34,11 @@ export const getChain = (chainId: number): chains.Chain => {
  * @returns result of the pipe
  */
 export const asyncPipe =
-  <T>(...fns: ((x: T) => Promise<T>)[]) =>
-  async (x: T) => {
-    let result = x;
+  <S, O, F>(...fns: ((s: S, o?: O, f?: F) => Promise<S>)[]) =>
```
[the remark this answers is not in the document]
denniswon: on a separate PR

packages/alchemy/e2e-tests/simple-account.test.ts
```diff
@@ -81,10 +87,14 @@ describe("Simple Account Tests", () => {
...
     await expect(txnHash).resolves.not.toThrowError();
-  }, 50000);
+  }, 100000);
 
   it("should successfully override fees in alchemy paymaster", async () => {
-    const provider = givenConnectedProvider({ owner, chain })
+    const provider = givenConnectedProvider({
+      owner,
+      chain,
+      feeOptions: { maxFeePerGas: undefined, maxPriorityFeePerGas: undefined },
```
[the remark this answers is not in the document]
denniswon: yes so that the default feeOptions that we set are removed by being overridden with undefined

packages/accounts/src/light-account/e2e-tests/light-account.test.ts
```diff
@@ -165,17 +164,6 @@ describe("Light Account Tests", () => {
       chain,
     });
 
-    // fund the throwaway address
```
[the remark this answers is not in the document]
denniswon: the e2e tests fail without this. we can add it back when it works.

packages/alchemy/src/middleware/gas-manager.ts
denniswon: @moldy530 got it. updated it.

packages/alchemy/src/middleware/gas-manager.ts
```diff
@@ -1,47 +1,101 @@
...
+ * If `delegateGasEstimation` is true, it will use `alchemy_requestGasAndPaymasterAndData` to get all of the gas estimates + paymaster data
  * in one RPC call.
  *
  * Otherwise, it will use `alchemy_requestPaymasterAndData` to get only paymaster data, allowing you
  * to customize the gas and fee estimation middleware.
  *
  * @param provider - the smart account provider to override to use the alchemy gas manager
  * @param config - the alchemy gas manager configuration
- * @param estimateGas - if true, this will use `alchemy_requestGasAndPaymasterAndData` else will use `alchemy_requestPaymasterAndData`
+ * @param delegateGasEstimation - if true, this will use `alchemy_requestGasAndPaymasterAndData` else will use `alchemy_requestPaymasterAndData`
```
avasisht23, line 38: wait did this change to an object called `gasEstimationOptions`?

packages/alchemy/src/middleware/gas-fees.ts
avasisht23: okay with that in mind, I actually recommend:
1. having a boolean like `applyProviderFeeOptions` instead of passing around the object itself
2. make the provider's `feeOptions` field readonly instead of protected so it's accessible

where I get confused a dev is why the class-level fields can't just be read from the provider directly if we have its context in any of these middleware decorators. passing them in again and again kind of defeats the point of having them listed as a class field in the first place.

The boolean captures the necessary intent of the developer more succinctly and clearly IMO. wdyt? cc: @moldy530

packages/alchemy/src/middleware/gas-manager.ts
denniswon: updated

packages/core/src/utils/index.ts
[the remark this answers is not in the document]
avasisht23: bump? doesn't look like this changed

packages/alchemy/e2e-tests/simple-account.test.ts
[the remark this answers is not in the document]
avasisht23: bump

packages/alchemy/e2e-tests/simple-account.test.ts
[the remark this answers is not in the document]
avasisht23: bump

packages/core/src/provider/types.ts
```diff
@@ -325,7 +329,8 @@ export interface ISmartAccountProvider<
    * Overrides the feeDataGetter middleware which is used for setting the fee fields on the UserOperation
    * prior to execution.
    *
-   * @param override - a function for overriding the default feeDataGetter middleware
+   * @param override   - a function for overriding the default feeDataGetter middleware
```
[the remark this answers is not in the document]
avasisht23: bump?

packages/core/src/provider/types.ts
```diff
@@ -325,7 +329,8 @@ export interface ISmartAccountProvider<
...
+   * @param feeOptions - optional FeeDataFeeOptions to set at the global level of the provider.
```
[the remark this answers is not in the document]
avasisht23: bump?

packages/alchemy/src/middleware/gas-manager.ts
avasisht23: bump?

avasisht23 (commented):
few leftover comments to address?

packages/alchemy/src/middleware/gas-manager.ts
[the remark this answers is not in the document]
moldy530: bump

packages/alchemy/src/middleware/gas-manager.ts
moldy530: bump?

packages/alchemy/src/middleware/types/index.ts
```diff
@@ -1 +1,11 @@
+import type { Percentage, UserOperationRequest } from "@alchemy/aa-core";
+
 export type * from "./simulate-uo/index.js";
+
+export type RequestGasAndPaymasterAndDataOverrides = Partial<{
+  maxFeePerGas: UserOperationRequest["maxFeePerGas"] | Percentage;
+  maxPriorityFeePerGas: UserOperationRequest["maxFeePerGas"] | Percentage;
+  callGasLimit: UserOperationRequest["maxFeePerGas"] | Percentage;
+  preVerificationGas: UserOperationRequest["maxFeePerGas"] | Percentage;
+  verificationGasLimit: UserOperationRequest["maxFeePerGas"] | Percentage;
```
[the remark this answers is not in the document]
moldy530: good catch!

packages/core/src/provider/base.ts
moldy530: bump

packages/core/src/provider/base.ts
moldy530: bump

packages/core/src/utils/bigint.ts
```diff
@@ -8,12 +8,58 @@ import type { BigNumberish } from "../types";
  */
 export const bigIntMax = (...args: bigint[]) => {
   if (!args.length) {
-    throw new Error("bigIntMax requires at least one argument");
+    return undefined;
```
moldy530, line 11: why not throw here?

packages/core/src/utils/index.ts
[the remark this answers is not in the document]
moldy530: oh just saw this... this tightly couples the use of async pipe with its use in the codebase now

Is there any way to do this with a `...args`? maybe it needs to be something like:
```suggestion
  <S>(...fns: ((s: S, ...args: any) => Promise<S>)[]) =>
```

packages/core/src/provider/base.ts
denniswon: so weird. I thought I already changed all these... hmm

packages/alchemy/src/middleware/gas-manager.ts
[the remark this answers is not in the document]
denniswon: updated

moldy530 (approved):
LGTM but there's still one unaddressed comment from @avasisht23 here

packages/core/src/provider/base.ts
```diff
@@ -221,15 +217,21 @@ export class SmartAccountProvider<
...
-        overrides?.maxPriorityFeePerGas ?? request.maxPriorityFeePerGas;
-    }
-    _overrides.paymasterAndData = overrides?.paymasterAndData;
+    const _overrides: UserOperationOverrides = {
+      maxFeePerGas:
+        overrides?.maxFeePerGas != null
+          ? overrides?.maxFeePerGas
+          : request.maxFeePerGas
+          ? fromHex(request.maxFeePerGas, "bigint")
+          : undefined,
```
avasisht23, line 226: could you make a ticket to follow up and generalize this value prioritization ternary statement against request, overrides, and maxFeePerGas into a util method? Priority: P2 :)

I can imagine we might need this elsewhere in the future (like below).

avasisht23 (approved):
lgtm, one last comment to address + 1 request for follow-up :) nice work!

packages/core/src/provider/base.ts
denniswon: https://app.asana.com/0/1205598840815267/1206061917337242/f

PR #286, opened 2023-11-28
fix: fix the fee override logic in smart account provider dropAndReplace

denniswon:
### Focus of the PR:
This PR focuses on improving the logging functionality and making adjustments to the account and provider classes.

### Detailed summary:
- Updated the default log level in the Logger class to `INFO`.
- Added a new `verbose` method to the Logger class for logging verbose messages.
- Modified the `shouldLog` method in the Logger class to check if a message should be logged based on the log level and filter.
- Updated the `maxFeePerGas` and `maxPriorityFeePerGas` values in the `UserOperationOverrides` object in the provider class.
- Adjusted the calculation of `baseFee` in the provider class.
- Replaced `Logger.debug` with `Logger.verbose` in the account class for logging debug messages.
- Updated the log messages in the account class for better clarity.

### DIFF
--- a/packages/core/src/provider/base.ts
+++ b/packages/core/src/provider/base.ts
@@ -396,15 +396,19 @@ export class SmartAccountProvider<
     const { maxFeePerGas, maxPriorityFeePerGas } =
       await this._runMiddlewareStack(uoToSubmit, overrides);
 
+    const _maxPriorityFeePerGas = bigIntMax(
+      BigInt(maxPriorityFeePerGas ?? 0n),
+      bigIntPercent(uoToDrop.maxPriorityFeePerGas, 110n)
+    );
     const _overrides: UserOperationOverrides = {
       maxFeePerGas: bigIntMax(
         BigInt(maxFeePerGas ?? 0n),
-        bigIntPercent(uoToDrop.maxFeePerGas, 110n)
+        bigIntPercent(
+          BigInt(uoToDrop.maxFeePerGas) - _maxPriorityFeePerGas,
+          110n
+        ) + _maxPriorityFeePerGas
       ),
-      maxPriorityFeePerGas: bigIntMax(
-        BigInt(maxPriorityFeePerGas ?? 0n),
-        bigIntPercent(uoToDrop.maxPriorityFeePerGas, 110n)
-      ),
+      maxPriorityFeePerGas: _maxPriorityFeePerGas,
       paymasterAndData: uoToDrop.paymasterAndData,
     };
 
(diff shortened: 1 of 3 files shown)


packages/core/src/provider/base.ts
```diff
@@ -396,15 +396,19 @@ export class SmartAccountProvider<
...
         BigInt(maxFeePerGas ?? 0n),
-        bigIntPercent(uoToDrop.maxFeePerGas, 110n)
-      ),
-      maxPriorityFeePerGas: bigIntMax(
-        BigInt(maxPriorityFeePerGas ?? 0n),
-        bigIntPercent(uoToDrop.maxPriorityFeePerGas, 110n)
+        bigIntPercent(
+          BigInt(uoToDrop.maxFeePerGas) - _maxPriorityFeePerGas,
+          110n
+        ) + _maxPriorityFeePerGas
```
avasisht23, line 409: since, `maxFeePerGas = baseFeePerGas + maxPriorityFeePerGas`, yeah this makes sense. this way, we don't double count the 10% bump.

however, shouldn't we just use the current pending block's baseFee?

denniswon: my understanding was the bump should be applied to the base fee, as otherwise, we would be double bumping the maxFeePerGas overall.

packages/core/src/provider/base.ts
```diff
@@ -396,15 +396,19 @@ export class SmartAccountProvider<
...
     const _overrides: UserOperationOverrides = {
       maxFeePerGas: bigIntMax(
         BigInt(maxFeePerGas ?? 0n),
-        bigIntPercent(uoToDrop.maxFeePerGas, 110n)
-      ),
-      maxPriorityFeePerGas: bigIntMax(
-        BigInt(maxPriorityFeePerGas ?? 0n),
-        bigIntPercent(uoToDrop.maxPriorityFeePerGas, 110n)
+        bigIntPercent(
+          BigInt(uoToDrop.maxFeePerGas) - _maxPriorityFeePerGas,
```
avasisht23, line 407: I think we need this:
```suggestion
          BigInt(uoToDrop.maxFeePerGas) - uoToDrop.maxPriorityFeePerGas,
```

That way, we correctly get:
```
const _maxPriorityFeePerGas = max (1.1 * uoToDrop.maxPriorityFeePerGas, maxPriorityFeePerGas)

const new_baseFeePerGas = 1.1 * (old_maxFeePerGas - old_maxPriorityFeePerGas)

const new_maxFeePerGas = new_baseFeePerGas + _maxPriorityFeePerGas
```

denniswon: thanks @avasisht23

denniswon:
per https://alchemyinsights.slack.com/archives/C04RHSDSP9U/p1701280202925489, closing this pr

PR #314, opened 2023-12-11
feat: msca plugingen to accept multichain address format

denniswon:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on updating the codebase to support multiple addresses for contracts. 

### Detailed summary
- Updated `types.ts` to import `Address` and `Hash` from "viem"
- Added `addresses` field to `PluginGenConfig` type in `plugindefs/types.ts`
- Updated `config.ts` to use `addresses` instead of `address` in `MultiOwnerPlugin` config
- Updated `wagmi.config.ts` to use `address` instead of `address` in `config`
- Updated `type.ts` to import `ConnectionConfigSchema` from "schema.js"
- Added `ConnectionConfig` type to `provider/types.ts`
- Updated `multi-owner.ts` to import `Address` from "viem"
- Updated `types.ts` to import `Address` from "@alchemy/aa-core"
- Updated `schema.ts` to include `ConnectionConfigSchema`
- Updated `plugingen.ts` to import `ConnectionConfig` from "@alchemy/aa-core" and use it for connection configuration


packages/accounts/wagmi.config.ts
```diff
@@ -20,6 +20,11 @@ export default defineConfig(
         address: config.address,
       },
     ],
-    plugins: [plugingen({ chain: config.chain, rpcUrl: config.rpcUrl })],
+    plugins: [
+      plugingen({
+        chain: config.chain,
+        connectionConfig: { rpcUrl: config.rpcUrl! },
```
moldy530, line 26: isn't rpcUrl optional?


```suggestion
        connectionConfig: { rpcUrl: config.rpcUrl },
```

denniswon: no, it's using ConnectionConfig like in AlchemyProvider

packages/accounts/wagmi.config.ts
```diff
@@ -20,6 +20,11 @@ export default defineConfig(
         address: config.address,
       },
     ],
-    plugins: [plugingen({ chain: config.chain, rpcUrl: config.rpcUrl })],
+    plugins: [
+      plugingen({
+        chain: config.chain,
```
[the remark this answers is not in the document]
denniswon: No I feel like creating one client and pulling contract data for one chain of reference is good. The plugin output is the same for different chains (under assumption that the contract is the same on multiple deployed chains)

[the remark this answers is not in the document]
denniswon: trying to avoid creating multiple viem public client and fetching the contract for each chain in a loop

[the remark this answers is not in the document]
moldy530: yea that's fine, we don't have to do it in a loop. We can just do `Object.entries(contract.address)[0]` to get the reference impl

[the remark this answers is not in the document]
moldy530: oh just rereviewed, yea this is fine

packages/accounts/plugindefs/types.ts
```diff
@@ -1,8 +1,9 @@
-import type { Abi, Address, Chain } from "viem";
+import type { ContractConfig } from "@wagmi/cli";
+import type { Abi, Chain } from "viem";
 
 export type PluginGenConfig = {
   abi: Abi;
-  address: Address;
+  address: ContractConfig["address"];
```
moldy530, line 6: doesn't this allow it to be either string or object now?

packages/accounts/scripts/plugingen.ts
```diff
@@ -135,6 +155,11 @@ export function plugingen({
                 meta: {
                     name: "${name}",
                     version: "${version}",
+                    address: {${Object.entries(address).reduce(
```
moldy530, line 158:
```suggestion
                    addresses: {${Object.entries(address).reduce(
```

packages/accounts/plugindefs/types.ts
moldy530: thinking about this, we should be prescriptive about this and only allow Record<number, Address>

moldy530: also let's just call this field `addresses` it's a bit weird to see it called `address` and then a list of addresses

packages/accounts/wagmi.config.ts
moldy530: yea but here https://github.com/alchemyplatform/aa-sdk/pull/314/files#diff-29103509588947d27bb0a950af40aa9d8eace89db6ef63d3927cee22e466dab7R31-R36 you still have to check for this null case directly

moldy530: is `!` just something javascript ignores? we don't set `config.rpcUrl` anywhere in our plugin configs. I think here is where should handle that case with something like:

```suggestion
        connectionConfig: config.rpcUrl ? { rpcUrl: config.rpcUrl } : {apiKey: process.env.API_KEY },
```

packages/accounts/plugindefs/multi-owner/config.ts
```diff
@@ -5,6 +5,8 @@ import { MultiOwnerPluginAbi } from "./abi.js";
 export const MultiOwnerPluginGenConfig: PluginGenConfig = {
   name: "MultiOwnerPlugin",
   abi: MultiOwnerPluginAbi,
-  address: "0x56bC629F342821FBe91C5273880792dFECBE7920",
+  address: {
```
moldy530, line 8:
```suggestion
  addresses: {
```

moldy530 (commented):
LGTM other than the one missed update

packages/accounts/plugindefs/multi-owner/config.ts
denniswon: hmm this is already updated

moldy530 (approved):
sweeet

PR #326, opened 2023-12-13
feat: plugin manager and session key support added to aa simple dapp

denniswon:
Demo videos:
- https://drive.google.com/file/d/1fuOMWelSP_2L-NYuzalw3nzOLtCDABgy/view?usp=drive_link
- https://drive.google.com/file/d/1IxitGxJq87ggF6pgMzPPrfIxSxSr2BcC/view?usp=drive_link

# Pull Request Checklist

---

### Focus of this PR:
This PR focuses on making changes to the code diff that include renaming a token contract address, updating import statements, adding new functions, and modifying existing functions.

### Detailed summary:
- Renamed `tokenContractAddress` to `alchTokenContractAddress` in `token-contract.ts`.
- Added `Content` component in `Root.tsx`.
- Imported `alchTokenContractAddress` in `Footer.tsx`.
- Added new functions and imports in `util.ts`.
- Modified `Navbar.tsx` to update a link and add new elements.
- Modified functions in `base.ts`.
- Added new hook `useAsyncEffect` in `useAsyncEffect.ts`.
- Updated imports and removed unused code in `useAlchemyProvider.ts`.
- Modified `Hero.tsx` by removing unused imports and variables.

> The following files were skipped due to too many changes: `examples/aa-simple-dapp/src/components/Hero.tsx`, `examples/aa-simple-dapp/src/hooks/useSessionKey.ts`, `examples/aa-simple-dapp/src/hooks/usePluginManager.ts`, `examples/aa-simple-dapp/src/context/wallet/index.tsx`, `examples/aa-simple-dapp/src/config/alch-nft.ts`, `examples/aa-simple-dapp/src/components/Content.tsx`


examples/aa-simple-dapp/src/hooks/useAsyncEffect.ts
```diff
@@ -0,0 +1,79 @@
...
+  const [result, setResult] = useState<any>();
+
+  useEffect(() => {
+    isMounted.current = true;
+    return () => {
+      isMounted.current = false;
+    };
+  }, []);
+
+  useEffect(() => {
```
moldy530, line 24: isn't this effect going to run every time the `deps` array updates? Should this no-op when `isMounted` is true?

moldy530: I ask because it seems unintended based on the description of the effect:
`Hook to run an async effect on mount and another on unmount`

moldy530 (approved):
LGTM, would be good to add a video demo to the PR of it working

examples/aa-simple-dapp/src/hooks/useAsyncEffect.ts
denniswon: yes when deps update, it is intended to be run again, just like useeffect.

examples/aa-simple-dapp/src/components/Content.tsx
```diff
@@ -0,0 +1,326 @@
...
+              .extend(SessionKeyPlugin.providerMethods)
+              .executeWithSessionKey(
+                {
+                  args: [
+                    [{ ...userOpData, value: userOpData.value || BigInt(0) }],
+                    sessionKey,
+                  ],
+                },
+                {
+                  signer: sessionKeySigner,
```
moldy530, line 84: yea I'm still not seeing the use case for this... there's a method in the `WalletContext` (i think?) called `connectProviderToAccount` which takes in a signer and an address. Why not use that method like so:

```ts
const { provider, connectProviderToAccount } = useWalletContext();
const { sessionKeySigner } = useSessionKey();

if (sessionKeySigner && provider.isConnected<MSCA>()) {
  connectProviderToAccount(sessionKeySigner, await provider.getAddress());
}
```

Now you don't need to override the signer ever

moldy530: you could probably have this `connectProviderToAccount` logic in the WalletContext itself (maybe in the `login` method or method that checks magic's existing state)

denniswon: I don't think this is a better approach.

1. Then the provider will only be connected to session key signer which is not an MSCA. then typescript for MSCA breaks because provider is not `Provider & { account?: MSCA }` (maybe it won't, haven't checked)
2. we might want to add a fallback to original account with signing in case session key signing somehow goes wrong. 

Also, I want to discuss about making `Plugin` as an object so that we can do `account.plugins.sessionkey.<some plugin method>` ultimately instead of using Static calls because currently devs have to manage individual plugin code all separately. Gonna be end up writing react hooks for every single plugin being used in their projects.

denniswon: 3. we also need to support all other overrides still for session key enabled accounts
4. we still want the original account to be connected to the provider to not let users bypass gas policy if using paymaster. We can still do that but then devs have to set the accountAddress for the smart account being connected, and that is quite hidden for devs. I can see devs often just copying pasting the original connect code here for session keys and accidentally have the session key signer being connected to another account address. There is no technical good way to catch these mistakes.

moldy530: One way to solve 2 (which will solve 4 as well) is to define the concept of a `SessionKeySigner` which can handle the fallback logic. That way when you connect an account you can connect it with the SessionKeySigner off the bat.

By consolidating logic into one class for signing, this should keep it in line with the rest of the flows we have. I'm weary of how many overrides we have now and it's not the best pattern for everything.

examples/aa-simple-dapp/src/components/Content.tsx
```diff
@@ -0,0 +1,326 @@
...
+          provider.isConnected<MSCA>() && sessionKeyEnabled && sessionKey;
+        console.log("executing with session key", executeWithSessionKey);
+        const uoHash = executeWithSessionKey
+          ? await provider
+              .extend(SessionKeyPlugin.providerMethods)
+              .executeWithSessionKey(
+                {
+                  args: [
+                    [{ ...userOpData, value: userOpData.value || BigInt(0) }],
+                    sessionKey,
```
moldy530, line 80: isn't this the same as `sessionKeySigner.getAddress()`?

It feels weird that the `executeWithSessionKey` takes in just the session key address but not the session key to use as well? Ah but that's because of how it's autogenned

having to call `executeWithSessionKey` maybe isn't the cleanest devex IMO. Can we leverage the builder pattern to have a method like `createAccountWithSessionKey` or something that makes this even simpler? That could return a built out MSCA using `withExecutor` which handles fallback logic of either encoding `execute` or `executeWithSessionKey`.

moldy530 (commented):
Ok yea this is more complex than I imagined thanks for putting all this together!

Looking at this, I think we might want to consider adding a layer of abstraction on top of using SessionKeys

Ideally something like a `SessionKeySigner` which can handle falling back to another signer at a minimum. To take it one step further we can also introduce a new builder function or `Executor` type which takes in a `SessionKeySigner` and for `encodeExecute` chooses the correct method based on the state of the session key on the SessionKeySigner.

Having `SessionKeySigner` which has a `fallbackSigner` property in its construction should tackle the edge cases you called out

denniswon:
replaced by this pr https://github.com/alchemyplatform/aa-sdk/pull/373

PR #389, opened 2024-01-29
feat: add rundler-js testing utils

moldy530:
WIP: still fleshing this out, but the attempt here is to move off using public rpc endpoints for e2e tests and executing our tests against anvil directly.
Doing this also lets us pin anvil to specific blocks which means we should get highly reproducible tests and hopefully much less flakiness

# Pull Request Checklist

---

### Detailed summary
- Added `createRundler` function in `rundlerjs/src/createRundler.ts` to create a bundler instance.
- Added `Rundler` type and `RundlerOptionsSchema` in `rundlerjs/src/types.ts` to define the bundler options and instance type.
- Updated `rundlerjs/test.mjs` to import and use the `createRundler` function.
- Updated `rundlerjs/package.json` to include `@alch/rundler-js` as a dev dependency.
- Updated `.github/workflows/publish-package.yml` and `.github/workflows/on-pull-request.yml` to use `actions/checkout@v4` and `actions/setup-node@v4`.
- Updated `rundlerjs/tsconfig.build.json` to exclude certain files and include the `src` folder.
- Updated `packages/core/package.json` to include `@alch/rundler-js` as a dev dependency.
- Updated `packages/core/src/account/__tests__/simple.test.ts` to import and use the `createRundler` function and `Rundler` type.

> The following files were skipped due to too many changes: `packages/core/src/account/__tests__/simple.test.ts`, `yarn.lock`

### DIFF
--- a/packages/core/src/account/__tests__/simple.test.ts
+++ b/packages/core/src/account/__tests__/simple.test.ts
@@ -1,14 +1,16 @@
-import { polygonMumbai, sepolia } from "@alchemy/aa-core";
+import { createRundler } from "@alch/rundler-js";
 import {
   createPublicClient,
   custom,
   http,
+  testActions,
   type Address,
   type Chain,
 } from "viem";
 import { describe, it } from "vitest";
+import { sepolia } from "../../chains/index.js";
 import { createBundlerClientFromExisting } from "../../client/bundlerClient.js";
-import { createSmartAccountClient } from "../../client/smartAccountClient.js";
+import { createSmartAccountClientFromExisting } from "../../client/smartAccountClient.js";
 import { LocalAccountSigner } from "../../signer/local-account.js";
 import { type SmartAccountSigner } from "../../signer/types.js";
 import type { BatchUserOperationCallData } from "../../types.js";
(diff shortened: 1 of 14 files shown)


rundlerjs/src/createRundler.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const stderr = new Writable({
+    write(chunk, _, callback) {
+      // TODO: send this somewhere else
+      console.error(chunk.toString());
+      callback();
+    },
+  });
+
+  const start = async () => {
```
moldy530, line 40: this method needs some better lifecycle handling

rundlerjs/src/createRundler.ts
```diff
@@ -0,0 +1,89 @@
...
+        "--entry_points",
+        rundlerOptions.entry_points.join(","),
+        "--chain_id",
+        rundlerOptions.chain_id.toString(),
+        "--node_http",
+        `http://${anvil.host}:${anvil.port}`,
+        "--max_verification_gas",
+        rundlerOptions.max_verification_gas.toString(),
+        "--builder.private_key",
+        rundlerOptions.builder.private_key,
```
moldy530, line 58: we should build these dynamically instead so we can expand the customizability more easily

packages/core/src/account/__tests__/simple.test.ts
```diff
@@ -10,23 +11,59 @@ import { getDefaultSimpleAccountFactoryAddress } from "../../utils/index.js";
 import { createSimpleSmartAccount } from "../simple.js";
 
 describe("Account Simple Tests", async () => {
+  const chain = sepolia;
   const dummyMnemonic =
     "test test test test test test test test test test test test";
   const owner: SmartAccountSigner =
     LocalAccountSigner.mnemonicToAccountSigner(dummyMnemonic);
+  const rundler = createRundler({
```
moldy530, line 19: in practice this is probably not what we want to use. we will want to use a pool like in https://github.com/wevm/anvil.js/blob/b09bd0ec8013a674dcc5dbe6885e12425bd1f823/examples/example-vitest/tests/utils.ts

moldy530:
once anvil is starting correctly, the next challenge to this is that the library needs to ship with binaries for different platforms. For this we might need to do something like: https://blog.orhun.dev/packaging-rust-for-npm/

moldy530:
I ran into some roadblocks here... I'll revisit at another time

PR #425, opened 2024-02-03
fix(aa-core): add back eip-1193 method handling to the client

moldy530:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on updating the `IncompatibleClientError` class in the codebase. The notable changes include:
- Adding the `client` parameter to the `IncompatibleClientError` constructor.
- Updating the error message to include the type of the client that caused the error.
- Adding a new `smartAccountClientMethodKeys` constant to check for supported RPC methods in a smart account client.

> The following files were skipped due to too many changes: `packages/core/src/client/smartAccountClient.ts`

### DIFF
--- a/packages/core/src/client/isSmartAccountClient.ts
+++ b/packages/core/src/client/isSmartAccountClient.ts
@@ -1,5 +1,6 @@
 import type { Chain, Client, Transport } from "viem";
 import type { SmartContractAccount } from "../account/smartContractAccount";
+import { smartAccountClientMethodKeys } from "./decorators/smartAccountClient.js";
 import type {
   BaseSmartAccountClient,
   SmartAccountClient,
@@ -22,9 +23,13 @@ export function isSmartAccountClient<
 >(
   client: Client<TTransport, TChain, TAccount>
 ): client is SmartAccountClient<TTransport, TChain, TAccount> {
-  return (
-    client && "middleware" in client && client.type === "SmartAccountClient"
-  );
+  for (const key of smartAccountClientMethodKeys) {
+    if (!(key in client)) {
+      return false;
+    }
+  }
+
+  return client && "middleware" in client;
 }
 
 /**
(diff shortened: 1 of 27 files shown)


packages/core/src/client/isSmartAccountClient.ts
```diff
@@ -22,9 +30,13 @@ export function isSmartAccountClient<
 >(
   client: Client<TTransport, TChain, TAccount>
 ): client is SmartAccountClient<TTransport, TChain, TAccount> {
-  return (
-    client && "middleware" in client && client.type === "SmartAccountClient"
```
moldy530, line 26: this check of the `type` was causing an E2E alchemy test to fail. So had to change this. The goal of the test is really to test what's going on on the right anyways: does this client support the actions we expect it to

packages/core/src/client/smartAccountClient.ts
```diff
@@ -106,25 +115,95 @@ export function createSmartAccountClient(
...
-    // we should support those here as well
-    transport,
-  });
+    // TODO: this needs to be tested
+    transport: (opts) => {
+      const rpcTransport = transport(opts);
 
-  return client
+      return custom({
+        async request({ method, params }) {
```
moldy530, line 135: adding these handlers back because I forgot to handle them after the refactor

moldy530:
I also caught a broken e2e test here.... we really need the rundler-js bit in here so we can move our e2e tests as part of the PR pipeline

packages/core/src/client/isSmartAccountClient.ts
denniswon: huh why was it failing? what was `client.type` referring to?

denniswon (approved):

packages/core/src/client/isSmartAccountClient.ts
moldy530: ah because the AlchemyClient sets the type to `AlchemySmartAccountClient` which results in this not passing in the core methods. So instead I changed this to make sure the methods on it match the expectation instead

PR #441, opened 2024-02-10
feat: add alchemy signer to aa-alchemy

moldy530:
# Pull Request Checklist

### DIFF
--- a/packages/alchemy/package.json
+++ b/packages/alchemy/package.json
@@ -47,6 +47,10 @@
   },
   "dependencies": {
     "@alchemy/aa-core": "^3.0.0-alpha.5",
+    "@turnkey/http": "^2.6.2",
+    "@turnkey/iframe-stamper": "^1.0.0",
+    "@turnkey/viem": "^0.4.8",
+    "@turnkey/webauthn-stamper": "^0.4.3",
     "viem": "^2.5.0"
   },
   "publishConfig": {
--- a/packages/alchemy/src/signer/session/manager.ts
+++ b/packages/alchemy/src/signer/session/manager.ts
@@ -1,0 +1,115 @@
+import { z } from "zod";
+import type { AlchemySignerClient } from "../client";
+import type { User } from "../client/types";
+import type { Session } from "./types";
+
+export const DEFAULT_SESSION_MS = 15 * 60 * 1000; // 15 minutes
+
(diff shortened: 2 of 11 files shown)


denniswon:
is this same as account-sdk repo src?

moldy530:
yup the signer definition will live in here, not account-sdk repo anymore

packages/alchemy/package.json
```diff
@@ -47,6 +47,10 @@
   },
   "dependencies": {
     "@alchemy/aa-core": "^3.0.0-alpha.5",
+    "@turnkey/http": "^2.6.2",
```
denniswon, line 50: can we add nohoist config for these so that other 3rd  party libs don't get hoisted to the root workspace?

packages/alchemy/src/signer/session/manager.ts
```diff
@@ -0,0 +1,115 @@
...
+  sessionKey: z.string().default("alchemy-signer-session"),
+  storage: z.enum(["localStorage", "sessionStorage"]).default("localStorage"),
+  expirationTimeMs: z
+    .number()
+    .default(DEFAULT_SESSION_MS)
+    .describe(
+      "The time in milliseconds that a session should last before expiring [default: 15 minutes]"
+    ),
+  client: z.custom<AlchemySignerClient>(),
+});
```
denniswon, line 18: thought we were going to remove zod? then should we keep adding new zod types?

denniswon (commented):
overall LGTM.

packages/alchemy/src/signer/session/manager.ts
moldy530: Ah yeah so I started writing these before the V3 refactor so this is now "legacy". In the interest of unblocking royal, I'll come back to this later. 

But yeah, in general let's not add more zod stuff because it gets hairy when generics are introduced.

packages/alchemy/package.json
moldy530: Not familiar with that, do you have a link to what that looks like / does?

moldy530:
gonna merge this in and come back to no hoist since I'm not convinced it's needed here. turnkey SDK has worked in this setting (aa-signers as example) and we build from root, so would rather keep the benefits of hoisting for now

PR #448, opened 2024-02-13
feat(amoy): add amoy from viem update

0xfourzerofour:
Add polygon amoy and update viem as there is a new chain definition in it.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -52,7 +52,7 @@
   "dependencies": {
     "abitype": "^0.8.3",
     "eventemitter3": "^5.0.1",
-    "viem": "^2.5.0",
+    "viem": "^2.7.8",
     "zod": "^3.22.4"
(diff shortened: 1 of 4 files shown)


packages/core/package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "abitype": "^0.8.3",
     "eventemitter3": "^5.0.1",
-    "viem": "^2.5.0",
+    "viem": "^2.7.8",
```
denniswon, line 55: viem versions should be updated all packages together not just aa-core. let's revert this for now unless not required.

denniswon (changes requested):

moldy530 (approved):
built just fine locally now that I've updated viem to 2.7.8 across the board

denniswon (approved):

PR #465, opened 2024-02-19
fix: respect plugin address override param in plugin management actions

denniswon:
# Pull Request Checklist

### DIFF
--- a/packages/accounts/src/msca/plugins/session-key/plugin.ts
+++ b/packages/accounts/src/msca/plugins/session-key/plugin.ts
@@ -331,7 +331,8 @@ export const sessionKeyPluginActions: <
 
     const dependencies = params.dependencyOverrides ?? [
       (() => {
-        const pluginAddress = MultiOwnerPlugin.meta.addresses[chain.id];
+        const pluginAddress =
+          params.pluginAddress ?? MultiOwnerPlugin.meta.addresses[chain.id];
         if (!pluginAddress) {
           throw new Error(
             "missing MultiOwnerPlugin address for chain " + chain.name
@@ -342,7 +343,8 @@ export const sessionKeyPluginActions: <
       })(),
 
       (() => {
-        const pluginAddress = MultiOwnerPlugin.meta.addresses[chain.id];
+        const pluginAddress =
+          params.pluginAddress ?? MultiOwnerPlugin.meta.addresses[chain.id];
         if (!pluginAddress) {
           throw new Error(
             "missing MultiOwnerPlugin address for chain " + chain.name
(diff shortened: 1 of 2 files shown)


packages/accounts/src/msca/plugins/session-key/plugin.ts
```diff
@@ -342,7 +343,8 @@ export const sessionKeyPluginActions: <
       })(),
 
       (() => {
-        const pluginAddress = MultiOwnerPlugin.meta.addresses[chain.id];
+        const pluginAddress =
+          params.pluginAddress ?? MultiOwnerPlugin.meta.addresses[chain.id];
```
moldy530, line 347: this doesn't seem right actually. if you pass in a pluginAddress override (as in a SessionKeyPlugin address) it will result in using that address as the override for MultiOwner. The right way to override the MultiOwnerPlugin address in this context is by using the `dependencyOverrides` field

denniswon: ah got it. but then shouldn't I do `[pluginname].meta.addresses[chain.id];` in the plugingen phase? feel like we should still use the plugin address override, and mention in the doc.

moldy530: >[pluginname].meta.addresses[chain.id]

I'm not following. That is what we do in the plugingen phase: https://github.com/alchemyplatform/aa-sdk/pull/465/files#diff-22631202035580cb943d00ec6863bb65fe01d9d81e2602ede833778d66080751L36

I think it's ok for the default dependency definitions to be the ones that come from `PluginGenConfig`. It's not that much of a deviation from how we use the plugin address. the plugin address is defined in the `PluginGenConfig` as defaults, but then a user can override. For dependencies same thing: default dependencies are defined in `PluginGenConfig` and then consumers can override them if they choose

moldy530: basically saying, if we want to allow peeps to override the dependency plugin addresses. then we either:
1) leave as is
2) add a new field to the `install*Plugin` method that allows for dependency address overrides. This will have to be a `Map<string, Address>` but the `string key actually needs to be a string const where the only values allowed are the unique names of the dependent Plugins. eg. `Map<"MultiOwnerPlugin", Address>`

moldy530: for `2` I think that does complicate things a bit, because now someone can pass in either `dependencyOverrides` or `dependencyPluginOverride`, so we would need to omit `dependencyOverrides` from the InstallParams

denniswon: ah makes sense. will leave as is. sorry got confused.

PR #486, opened 2024-02-25
feat: sign user op and send raw user op methods added to smart account client

denniswon:
# Pull Request Checklist

---

## PR-Codex overview
This PR introduces new types and functions related to user operation handling in the smart account client. 

### Detailed summary
- Added `UserOperationCallData`, `SendUserOperationResult`, and `UpgradeToData` types
- Updated functions to include `UserOperationOverrides` parameter
- Refactored internal functions for user operation handling

> The following files were skipped due to too many changes: `packages/core/src/actions/smartAccount/signUserOperation.ts`, `site/packages/aa-core/smart-account-client/actions/checkGasSponsorshipEligibility.md`, `packages/core/src/actions/smartAccount/types.ts`, `site/packages/aa-core/smart-account-client/actions/signUserOperation.md`, `site/packages/aa-core/smart-account-client/actions/sendUserOperation.md`, `packages/core/src/client/decorators/smartAccountClient.ts`, `site/packages/aa-core/smart-account-client/actions/sendTransaction.md`, `site/packages/aa-core/smart-account-client/actions/sendTransactions.md`, `site/using-smart-accounts/send-user-operations.md`, `site/packages/aa-core/smart-account-client/actions/buildUserOperationFromTx.md`, `site/packages/aa-core/smart-account-client/actions/buildUserOperation.md`, `site/packages/aa-core/smart-account-client/actions/buildUserOperationFromTxs.md`


packages/core/src/actions/smartAccount/buildUserOperationFromTxs.ts
```diff
@@ -20,10 +18,7 @@ export const buildUserOperationFromTxs: <
 >(
   client: Client<TTransport, TChain, TAccount>,
   args: SendTransactionsParameters<TAccount>
-) => Promise<BuildUserOperationFromTransactionsResult> = async (
```
moldy530, line 23: why not just update the type definition of `BuildUserOperationFromTransactionsResult`?

packages/core/src/actions/smartAccount/sendRawUserOperation.ts
```diff
@@ -0,0 +1,52 @@
...
+} from "viem";
+import type { SmartContractAccount } from "../../account/smartContractAccount";
+import { isBaseSmartAccountClient } from "../../client/isSmartAccountClient.js";
+import { AccountNotFoundError } from "../../errors/account.js";
+import { IncompatibleClientError } from "../../errors/client.js";
+import { InvalidUserOperationError } from "../../errors/useroperation.js";
+import { isValidRequest } from "../../utils/index.js";
+import type { SendRawUserOperationParameters } from "./types";
+
+export const sendRawUserOperation: <
```
moldy530, line 16: why redefine this? it's already being added on the smartAccountClient: https://github.com/alchemyplatform/aa-sdk/blob/main/packages/core/src/client/smartAccountClient.ts#L122C38-L122C57

packages/core/src/actions/smartAccount/buildUserOperationFromTxs.ts
```diff
@@ -77,8 +72,9 @@ export const buildUserOperationFromTxs: <
   };
   filterUndefined(_overrides);
 
-  return {
-    batch,
+  return buildUserOperation(client, {
+    uo: batch,
     overrides: _overrides,
-  };
```
moldy530, line 83: this is a breaking change....

packages/core/src/actions/smartAccount/types.ts
```diff
@@ -26,6 +27,18 @@ export type SendUserOperationParameters<
...
+  TAccount extends SmartContractAccount | undefined
+> = {
+  uoStruct: UserOperationStruct;
+} & GetAccountParameter<TAccount>;
+
+export type SendRawUserOperationParameters<
+  TAccount extends SmartContractAccount | undefined
+> = {
+  request: UserOperationRequest;
+} & GetAccountParameter<TAccount>;
```
moldy530, line 40: this is already defined on the `bundlerClient` as part of `bundlerActions` and already added to the smartAccountClient
```suggestion

```

packages/core/src/actions/smartAccount/types.ts
```diff
@@ -43,8 +56,3 @@ export type DropAndReplaceUserOperationParameters<
 export type WaitForUserOperationTxParameters = {
   hash: Hex;
 };
-
-export type BuildUserOperationFromTransactionsResult = {
-  batch: BatchUserOperationCallData;
-  overrides: UserOperationOverrides;
-};
```
moldy530, line 50: this is a breaking change

packages/core/src/client/decorators/smartAccountClient.ts
```diff
@@ -74,6 +81,9 @@ export type BaseSmartAccountClientActions<
   sendUserOperation: (
     args: SendUserOperationParameters<TAccount>
   ) => Promise<SendUserOperationResult>;
+  sendRawUserOperation: (
+    args: SendRawUserOperationParameters<TAccount>
+  ) => Promise<Hash>;
```
moldy530, line 86:
```suggestion
```

packages/core/src/client/decorators/smartAccountClient.ts
```diff
@@ -114,11 +124,13 @@ export const smartAccountClientActions: <
   buildUserOperationFromTxs: (args) => buildUserOperationFromTxs(client, args),
   checkGasSponsorshipEligibility: (args) =>
     checkGasSponsorshipEligibility(client, args),
+  signUserOperation: (args) => signUserOperation(client, args),
   dropAndReplaceUserOperation: (args) =>
     dropAndReplaceUserOperation(client, args),
   sendTransaction: (args) => sendTransaction(client, args),
   sendTransactions: (args) => sendTransactions(client, args),
   sendUserOperation: (args) => sendUserOperation(client, args),
+  sendRawUserOperation: (args) => sendRawUserOperation(client, args),
```
moldy530, line 133:
```suggestion
```

site/packages/aa-core/smart-account-client/actions/buildUserOperationFromTx.md
```diff
@@ -14,7 +14,7 @@ head:
 
 # buildUserOperationFromTx
 
-Converts a traditional Ethereum transaction and builds an _unsigned_ `UserOperation` (UO) struct with the all of the middleware run on it through the middleware pipeline.
+A Promise containing the _unsigned_ UO struct converted from the input transaction with all the middleware run on the resulting UO
```
moldy530, line 17: This sounds like it describes the return type.. but this section is the description of the method. I think it was ok before, but with a fix:
 ```suggestion
Converts a traditional Ethereum transaction and builds an _unsigned_ `UserOperation` (UO) struct with all of the middleware run on it through the middleware pipeline.
```

File: site/packages/aa-core/smart-account-client/actions/buildUserOperationFromTx.md [hunk 9]
```diff
@@ -24,7 +24,7 @@ The order of the middlewares is:
 4.  `paymasterMiddleware` -- used to set paymasterAndData. (default: "0x")
 5.  `customMiddleware` -- allows you to override any of the results returned by previous middlewares
 
-Note that `to` field of transaction is required, and among other fields of transaction, only `data`, `value`, `maxFeePerGas`, `maxPriorityFeePerGas` fields are considered and optional.
+Note that among all the transaction request object fields available, only `to`, `data`, `value`, `maxFeePerGas`, and `maxPriorityFeePerGas` fields are used to extract the user operation calldata from the input transaction and build the corresponding user operation struct.
```
moldy530, line 27:
```suggestion
Note that of all of the transaction request object fields available, only `to`, `data`, `value`, `maxFeePerGas`, and `maxPriorityFeePerGas` fields are used to extract the user operation calldata from the input transaction and build the corresponding user operation struct.
```

site/packages/aa-core/smart-account-client/actions/buildUserOperationFromTxs.md
```diff
@@ -0,0 +1,71 @@
...
+    - name: description
+      content: Overview of the buildUserOperationFromTxs method on SmartAccountClient
+  - - meta
+    - property: og:description
+      content: Overview of the buildUserOperationFromTxs method on SmartAccountClient
+---
+
+# buildUserOperationFromTxs
+
+A Promise containing the _unsigned_, batched UO struct converted from the input transactions with all the middleware run on the resulting UO
```
moldy530, line 17: This also sounds like a description of the return type not the method

site/packages/aa-core/smart-account-client/actions/sendRawUserOperation.md
```diff
@@ -0,0 +1,51 @@
...
+      content: sendRawUserOperation
+  - - meta
+    - name: description
+      content: Overview of the sendRawUserOperation method on SmartAccountClient
+  - - meta
+    - property: og:description
+      content: Overview of the sendRawUserOperation method on SmartAccountClient
+---
+
+# sendRawUserOperation
```
moldy530, line 15: this doc already exists https://accountkit.alchemy.com/packages/aa-core/bundler-client/actions/sendRawUserOperation.html

site/packages/aa-core/smart-account-client/actions/sendRawUserOperation.md
```diff
@@ -0,0 +1,51 @@
...
+
+Sends a user operation or batch of user operations using the connected account.
+
+Before executing, sendUserOperation will run the user operation through the middleware pipeline. The order of the middlewares is:
+
+1.  `dummyPaymasterDataMiddleware` -- populates a dummy paymaster data to use in estimation (default: "0x")
+2.  `feeDataGetter` -- sets maxfeePerGas and maxPriorityFeePerGas
+3.  `gasEstimator` -- calls eth_estimateUserOperationGas
+4.  `paymasterMiddleware` -- used to set paymasterAndData. (default: "0x")
+5.  `customMiddleware` -- allows you to override any of the results returned by previous middlewares
```
moldy530, line 25: this isn't true based on the impl above. it doesn't run middleware

packages/core/src/actions/smartAccount/buildUserOperationFromTxs.ts
denniswon: wasn't buildUserOperationFromTxs always returning uo struct with middleware pipeline run?
I just saw that now buildUserOperationFromTxs is just returning the uo batch for the txs, without calling buildUO function inside. so this change was intentional? I thought this was a mistake.

packages/core/src/actions/smartAccount/sendRawUserOperation.ts
denniswon: oh I gotta use a different name, but this method is different from what sendRawUserOperation of bundler client does (just call rpc)

packages/core/src/actions/smartAccount/types.ts
denniswon: ditto: my comment above. will change the name of this added func.

packages/core/src/actions/smartAccount/sendRawUserOperation.ts
denniswon: it's just that the Bundler client's `sendRawUserOperation` requires account.getEntryPoint().address to be passed in as a param. So wanted to make the function formats consistent with other smart account client actions.

packages/core/src/actions/smartAccount/buildUserOperationFromTxs.ts
moldy530: https://github.com/alchemyplatform/aa-sdk/blob/v2.x.x/packages/core/src/provider/base.ts#L296 nope, it was always returning this object

packages/core/src/actions/smartAccount/sendTransaction.ts
```diff
@@ -1,35 +1,30 @@
...
   TAccount extends SmartContractAccount | undefined =
     | SmartContractAccount
-    | undefined,
-  TChainOverride extends Chain | undefined = Chain | undefined
+    | undefined
 >(
-  client: Client<Transport, TChain, TAccount>,
-  args: SendTransactionParameters<TChain, TAccount, TChainOverride>
+  client: Client<TTransport, TChain, TAccount>,
+  args: SendTransactionParameters<TAccount>
```
moldy530, line 20: does. this still build? this was the `viem` type before because it caused typing issues without it (because there was a conflict between the public client action of the same name

packages/core/src/actions/smartAccount/sendRawUserOperation.ts
moldy530: hmmm good catch, but there's still the issue that the rest of bundler actions don't follow that same standard. We'd have to fix the inconsistency across the board. I'd rather keep the bundler actions and the smart account client actions consistent amongst themselves for now. We should fix ALL of the bundler client actions to match the object param style of the other clients, but we can't change that API right now

moldy530: I think we should not duplicate the actions here, and track that we need to refactor the bundler client action signatures next time we intro a breaking change (likely v4 or if need a breaking change in the entrypoint 0.7 support)

packages/core/src/actions/smartAccount/buildUserOperationFromTxs.ts
denniswon: wouldn't that be a breaking change?

denniswon:
@moldy530 PR is ready for review.

packages/core/src/actions/smartAccount/sendTransactions.ts
```diff
@@ -31,20 +31,16 @@ export const sendTransactions: <
...
+  const { uoStruct } = await buildUserOperationFromTxs(client, {
+    requests,
+    overrides,
     account,
   });
 
+  const { hash } = await _sendUserOperation(client, {
+    account: account as SmartContractAccount,
+    uoStruct,
+  });
```
moldy530, line 43: this is gonna run the middleware stack again, so you need to pass the overrides from above through to here.

I don't think `buildUserOperationFromTx` was ever meant to actually call `buildUserOperation`... which I admit makes the name super confusing

packages/core/src/client/decorators/smartAccountClient.ts
```diff
@@ -54,19 +60,24 @@ export type BaseSmartAccountClientActions<
...
+  signUserOperation: (
+    args: SignUserOperationParameters<TAccount>
+  ) => Promise<UserOperationRequest>;
   dropAndReplaceUserOperation: (
     args: DropAndReplaceUserOperationParameters<TAccount>
   ) => Promise<SendUserOperationResult>;
   sendTransaction: <TChainOverride extends Chain | undefined = undefined>(
-    args: SendTransactionParameters<TChain, TAccount, TChainOverride>
+    args: SendTransactionParameters<TChain, TAccount, TChainOverride>,
+    overrides?: UserOperationOverrides
```
moldy530, line 80: ah yea good catch here

packages/core/src/client/decorators/smartAccountClient.ts
```diff
@@ -54,19 +60,24 @@ export type BaseSmartAccountClientActions<
     args: SendUserOperationParameters<TAccount>
   ) => Promise<UserOperationStruct>;
   buildUserOperationFromTx: (
-    args: SendTransactionParameters<TChain, TAccount>
+    args: SendTransactionParameters<TChain, TAccount>,
+    overrides?: UserOperationOverrides
   ) => Promise<UserOperationStruct>;
   buildUserOperationFromTxs: (
     args: SendTransactionsParameters<TAccount>
   ) => Promise<BuildUserOperationFromTransactionsResult>;
```
moldy530, line 68: I think we should leave these two as is (I know it's confusing), but they were initially just meant to be used to convert a TX into a UO type that can be passed in to `sendUserOperation`. The thinking was that if a dev wanted to send a TX object as a UO, they weren't limited to just `sendTransaction`, but they could do `sendUserOperation()` with the result of `buildUserOperationFromTx(s)`

packages/core/src/actions/smartAccount/sendTransactions.ts
denniswon: this doesn't run middleware pipeline https://github.com/alchemyplatform/aa-sdk/blob/d08ac65a7e9d3ee1d82c803d25911804afdb4a05/packages/core/src/actions/smartAccount/internal/sendUserOperation.ts#L14 this is the internal send user op func

packages/core/src/actions/smartAccount/buildUserOperationFromTxs.ts
```diff
@@ -77,7 +78,15 @@ export const buildUserOperationFromTxs: <
   };
   filterUndefined(_overrides);
 
+  const uoStruct = await buildUserOperation(client, {
+    uo: batch,
+    overrides: _overrides,
+    account: account as SmartContractAccount,
+  });
```
moldy530, line 85: after reviewing, I think we should leave this method as is.

I think a todo for v4 would be:
1. rename this to `convertTxToUserOperation`
2. add `buildUserOperationFromTx` which calls `buildUserOperation` as you have it here

For now, if someone wants to actually `buildUserOperationFromTx` it would look like:

```ts
const { batch, overrides } = client.buildUserOperationFromTx(...);
const result = client.buildUserOperation({ uo: batch, overrides });
```

moldy530 (commented):
~Overall LGTM, with the caveat of leaving the `buildTx` methods as is right now, which I know isn't ideal since what they really do is `convertTxToUo`. Then we can follow up in v4 with the improvements you make in this PR because what you have here makes more sense. It's just that in the current form, `sendTransaction` is made more inefficient because it runs the middleware stack twice. This adds complexity and latency~

never mind! just saw what `buildUserOperationFromTx` was doing. I'm ok with this change. Are we able to merge the `args` and `overrides` below so the signature matches the rest of the methods?

packages/core/src/actions/smartAccount/sendTransactions.ts
moldy530: oh wait you're right

moldy530: should we then update the `buildUserOperationFromTx` as well?

moldy530: oh it already does this... gotcha so it's just `buildUserOperationFromTxs` that's weird

packages/core/src/client/decorators/smartAccountClient.ts
```diff
@@ -54,19 +60,24 @@ export type BaseSmartAccountClientActions<
     args: SendUserOperationParameters<TAccount>
   ) => Promise<UserOperationStruct>;
   buildUserOperationFromTx: (
-    args: SendTransactionParameters<TChain, TAccount>
+    args: SendTransactionParameters<TChain, TAccount>,
+    overrides?: UserOperationOverrides
```
moldy530, line 64: ah wait, should this be like instead:

```suggestion
    args: SendTransactionParameters<TChain, TAccount> & {overrides?: UserOperationOverrides}
```

packages/core/src/client/decorators/smartAccountClient.ts
moldy530, line 80: and same here:

and same here::

```suggestion
    args: SendTransactionParameters<TChain, TAccount> & {overrides?: UserOperationOverrides}
```

packages/core/src/actions/smartAccount/buildUserOperationFromTxs.ts
moldy530: ah wait nevermind... it's just the `Txs` one that wasn't calling `buildUo` never mind

packages/core/src/actions/smartAccount/sendTransactions.ts
denniswon: lol

packages/core/src/client/decorators/smartAccountClient.ts
denniswon: @moldy530 I can't.

viem public client blocked overriding the param types of some of PublicActions protected methods. 

Screenshot 2024-02-28 at 6 00 05 PM

```
export type Client<
  transport extends Transport = Transport,
  chain extends Chain | undefined = Chain | undefined,
  account extends Account | undefined = Account | undefined,
  rpcSchema extends RpcSchema | undefined = undefined,
  extended extends Extended | undefined = Extended | undefined,
> = Client_Base<transport, chain, account, rpcSchema> &
  (extended extends Extended ? extended : unknown) & {
    extend: <
      const client extends Extended &
        Partial<ExtendableProtectedActions<transport, chain, account>>,
    >(
      fn: (
        client: Client<transport, chain, account, rpcSchema, extended>,
      ) => client,
    ) => Client<
      transport,
      chain,
      account,
      rpcSchema,
      Prettify<client> & (extended extends Extended ? extended : unknown)
    >
  }
```

https://github.com/alchemyplatform/aa-sdk/blob/denniswon/sign-send-raw-user-op/packages/core/node_modules/viem/clients/createClient.ts#L59
```
// Actions that are used internally by other Actions (ie. `call` is used by `readContract`).
// They are allowed to be extended, but must conform to their parameter & return type interfaces.
// Example: an extended `call` action must accept `CallParameters` as parameters,
// and conform to the `CallReturnType` return type.
type ExtendableProtectedActions<
  transport extends Transport = Transport,
  chain extends Chain | undefined = Chain | undefined,
  account extends Account | undefined = Account | undefined,
> = Pick<
  PublicActions<transport, chain, account>,
  | 'call'
  | 'createContractEventFilter'
  | 'createEventFilter'
  | 'estimateContractGas'
  | 'estimateGas'
  | 'getBlock'
  | 'getBlockNumber'
  | 'getChainId'
  | 'getContractEvents'
  | 'getEnsText'
  | 'getFilterChanges'
  | 'getGasPrice'
  | 'getLogs'
  | 'getTransaction'
  | 'getTransactionCount'
  | 'getTransactionReceipt'
  | 'prepareTransactionRequest'
  | 'readContract'
  | 'sendRawTransaction'
  | 'simulateContract'
  | 'uninstallFilter'
  | 'watchBlockNumber'
  | 'watchContractEvent'
> &
  Pick<WalletActions<chain, account>, 'sendTransaction' | 'writeContract'>
```

packages/core/src/client/decorators/smartAccountClient.ts
denniswon: refer to my comment below. @moldy530

packages/core/src/client/decorators/smartAccountClient.ts
moldy530: ah gotcha, ok yea that checks out

moldy530 (approved):

PR #532, opened 2024-03-27
fix: dummy paymaster and data addresses updates

avarobinson:
# Pull Request Checklist

---

## PR-Codex overview
The focus of this PR is to update Alchemy paymaster addresses based on specific chains and IDs.

### Detailed summary
- Added `polygonMumbai` and `sepolia` to gas manager
- Updated Alchemy paymaster addresses for different chains and IDs

### DIFF
--- a/packages/alchemy/src/gas-manager.ts
+++ b/packages/alchemy/src/gas-manager.ts
@@ -6,40 +6,50 @@ import {
   baseSepolia,
   fraxtal,
   fraxtalTestnet,
+  goerli,
   mainnet,
   optimism,
   optimismSepolia,
   polygon,
   polygonAmoy,
+  polygonMumbai,
+  sepolia,
   zora,
   zoraSepolia,
 } from "viem/chains";
 
-export const NewAlchemyPaymasterAddress =
+export const AlchemyPaymasterAddressV2 =
   "0x4Fd9098af9ddcB41DA48A1d78F91F1398965addc";
 export const OldAlchemyPaymasterAddress =
   "0xc03aac639bb21233e0139381970328db8bceeb67";
 export const ArbSepoliaPaymasterAddress =
   "0x0804Afe6EEFb73ce7F93CD0d5e7079a5a8068592";
+export const AlcehmyPaymasterAddressV3 =
+  "0x4f84a207A80c39E9e8BaE717c1F25bA7AD1fB08F";
 
 export const getAlchemyPaymasterAddress = (chain: Chain): Address => {
   switch (chain) {
     case mainnet:
+    case arbitrum:
     case optimism:
     case polygon:
-    case arbitrum:
+    case base:
+      return AlchemyPaymasterAddressV2;
     case polygonAmoy:
     case optimismSepolia:
-    case base:
     case baseSepolia:
     case zora:
     case zoraSepolia:
     case fraxtal:
     case fraxtalTestnet:
-      return NewAlchemyPaymasterAddress;
+      return AlcehmyPaymasterAddressV3;
     case arbitrumSepolia:
+      return ArbSepoliaPaymasterAddress;
+    case sepolia:
+    case goerli:
+    case polygonMumbai:
       return OldAlchemyPaymasterAddress;
     default:
-      return ArbSepoliaPaymasterAddress;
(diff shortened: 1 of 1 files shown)


packages/alchemy/src/gas-manager.ts
```diff
@@ -6,40 +6,50 @@ import {
...
+      return AlcehmyPaymasterAddressV3;
     case arbitrumSepolia:
+      return ArbSepoliaPaymasterAddress;
+    case sepolia:
+    case goerli:
+    case polygonMumbai:
       return OldAlchemyPaymasterAddress;
     default:
-      return ArbSepoliaPaymasterAddress;
+      return OldAlchemyPaymasterAddress;
```
rthomare, line 53: I think at this point we should throw... we shouldn't put a default value in the enum.

packages/alchemy/src/gas-manager.ts
```diff
@@ -6,40 +6,50 @@ import {
...
     case zora:
     case zoraSepolia:
     case fraxtal:
     case fraxtalTestnet:
-      return NewAlchemyPaymasterAddress;
+      return AlcehmyPaymasterAddressV3;
     case arbitrumSepolia:
+      return ArbSepoliaPaymasterAddress;
+    case sepolia:
+    case goerli:
```
avasisht23, line 49: we shouldn't support eth goerli anymore right? we deprecated support

packages/alchemy/src/gas-manager.ts
```diff
@@ -6,40 +6,50 @@ import {
...
+  polygonMumbai,
+  sepolia,
   zora,
   zoraSepolia,
 } from "viem/chains";
 
-export const NewAlchemyPaymasterAddress =
+export const AlchemyPaymasterAddressV2 =
   "0x4Fd9098af9ddcB41DA48A1d78F91F1398965addc";
 export const OldAlchemyPaymasterAddress =
```
avasisht23, line 23: call this v1?

packages/alchemy/src/gas-manager.ts
rthomare: If it's a chain we don't support no use in putting a random address in there.

packages/alchemy/src/gas-manager.ts
```diff
@@ -6,40 +6,50 @@ import {
...
   "0xc03aac639bb21233e0139381970328db8bceeb67";
 export const ArbSepoliaPaymasterAddress =
   "0x0804Afe6EEFb73ce7F93CD0d5e7079a5a8068592";
+export const AlcehmyPaymasterAddressV3 =
```
avasisht23, line 27:
```suggestion
export const AlchemyPaymasterAddressV3 =
```

packages/alchemy/src/gas-manager.ts
rthomare: Also it's probably better to match on chainId than just the whole chain object. That way we can support customized chains.

rthomare (commented):
Not going to block on my comments but we should make some updates here.

packages/alchemy/src/gas-manager.ts
avasisht23: also nit: can we order this list in descending order: v3, v2, v1

packages/alchemy/src/gas-manager.ts
```diff
@@ -6,40 +6,50 @@ import {
...
     case fraxtalTestnet:
-      return NewAlchemyPaymasterAddress;
+      return AlcehmyPaymasterAddressV3;
     case arbitrumSepolia:
+      return ArbSepoliaPaymasterAddress;
+    case sepolia:
+    case goerli:
+    case polygonMumbai:
       return OldAlchemyPaymasterAddress;
     default:
```
avasisht23, line 52: don't need to have default different from the above cases

packages/alchemy/src/gas-manager.ts
moldy530: we need to update these middleware factories to also allow for paymaster address and dummy paymaster and data overrides. that way we're not blocking devs on new pushes

moldy530 (approved):

PR #514, opened 2024-03-16
feat(entrypoint-0.7): base - all changes in regards to the Entrypoint v6 & v7 support

denniswon:
# Pull Request Checklist

- [ ] Did you add new tests and confirm existing tests pass? (`yarn test`)
- [ ] Did you update relevant docs? (docs are found in the `site` folder, and guidelines for updating/adding docs can be found in the [contribution guide](https://github.com/alchemyplatform/aa-sdk/blob/main/CONTRIBUTING.md))
- [ ] Do your commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [ ] Does your PR title also follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [ ] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples)? (e.g. `feat!: breaking change`)
- [ ] Did you run lint (`yarn lint:check`) and fix any issues? (`yarn lint:write`)
- [ ] Did you follow the [contribution guidelines](https://github.com/alchemyplatform/aa-sdk/blob/main/CONTRIBUTING.md)?

---

## PR-Codex overview
This PR updates various files across packages to introduce versioning changes related to entry points, smart accounts, and APIs.

### Detailed summary
- Renamed entry point ABI files to include version suffix
- Updated `HexSchema` to use strict hex validation
- Added `API_KEY_STAGING` constant in test constants
- Modified functions to use `Promise<NullAddress | Address>`
- Updated function signatures to include versioning information
- Introduced new error classes for entry point and user operation
- Refactored middleware functions to handle different entry point versions

[...]

PR #539, opened 2024-03-30
feat: add alchemy accounts context

moldy530:
# Pull Request Checklist

---

## PR-Codex overview
This PR introduces new React hooks and config options, enhances error handling, and adds client-side properties. 

### Detailed summary
- Added `AWAITING_EMAIL_AUTH` status in `signer/types.ts`
- Created `config` with `rpcUrl` and `chain` in `site/snippets/react/config.ts`
- Implemented `ComponentWithSignerStatus` and `ComponentWithUser` in `site/snippets/react/useSignerStatus.tsx` and `site/snippets/react/useUser.tsx`
- Added `NoAlchemyAccountContextError` in `packages/alchemy/src/react/errors.ts`
- Updated `getBundlerClient`, `getUser`, `getSigner`, `getSignerStatus` in `packages/alchemy/src/config/actions/`
- Modified `tsconfig.json` for `moduleResolution` and `module`
- Enhanced session management in `packages/alchemy/src/signer/session/manager.ts`
- Added `ClientOnlyPropertyError` in `packages/alchemy/src/config/errors.ts`
- Implemented `ComponentWithSigner` and `ComponentWithBundlerClient` in `site/snippets/react/useSigner.tsx` and `site/snippets/react/useBundlerClient.tsx`
- Created `Login` component in `site/snippets/react/login.tsx`
- Introduced `watchUser`, `watchSigner`, and `watchSignerStatus` in `packages/alchemy/src/config/actions/`
- Updated `useUser` and `useBundlerClient` hooks in `packages/alchemy/src/react/hooks/`

> The following files were skipped due to too many changes: `packages/alchemy/src/client/smartAccountClient.ts`, `site/snippets/react/useAccount.tsx`, `site/snippets/react/useSmartAccountClient.tsx`, `packages/core/src/account/smartContractAccount.ts`, `packages/alchemy/src/react/hooks/useSigner.ts`, `site/react/useUser.md`, `packages/alchemy/src/config/actions/watchAccount.ts`, `packages/alchemy/src/config/actions/getAccount.ts`, `packages/alchemy/src/config/store/core.ts`, `packages/alchemy/src/config/index.ts`, `site/react/useBundlerClient.md`, `packages/alchemy/src/react/hooks/useSignerStatus.ts`, `site/react/useSigner.md`, `packages/alchemy/src/react/index.ts`, `packages/alchemy/package.json`, `packages/alchemy/src/react/hooks/useAuthenticate.ts`, `site/react/useAuthenticate.md`, `packages/alchemy/src/config/createConfig.ts`, `packages/alchemy/src/config/store/types.ts`, `packages/alchemy/src/react/context.ts`, `site/react/useSignerStatus.md`, `packages/alchemy/src/config/types.ts`, `packages/alchemy/src/react/hooks/useAccount.ts`, `site/react/createConfig.md`, `yarn.lock`, `site/react/overview.md`, `site/react/useAccount.md`, `packages/alchemy/src/config/actions/createAccount.ts`, `site/react/useSmartAccountClient.md`, `packages/alchemy/src/react/hooks/useSmartAccountClient.ts`, `packages/alchemy/src/config/store/client.ts`


moldy530:
example usage can be found in https://github.com/alchemyplatform/embedded-accounts-demo/pull/4

moldy530:
TODOS
- [ ] address TODOs in code
- [ ] add a summary of the changes and the proposal for how this works

packages/alchemy/src/react/hooks/useAuthenticate.ts
```diff
@@ -0,0 +1,37 @@
...
+        if (!signer) {
+          throw new ClientOnlyPropertyError("signer");
+        }
+
+        const user = await signer.authenticate(authParams);
+
+        return user;
+      },
+    },
+    queryClient
```
moldy530, line 29: I had to do this because for w/e reason react-query exports hooks under `modern` or under `legacy`. `aa-alchemy` was using legacy and the demo app was using modern. doing it this way ensures a query client is always present

packages/alchemy/src/react/hooks/useAccount.ts
```diff
@@ -0,0 +1,59 @@
...
+  useEffect(() => {
+    if (
+      !params.skipCreate &&
+      status === AlchemySignerStatus.CONNECTED &&
+      !account &&
+      !isPending
+    ) {
+      mutate();
+    }
+  }, [account, isPending, mutate, params.skipCreate, status]);
```
moldy530, line 49: I'm gonna try condensing this into a `useQuery` call instead of this.

The reason I had to do this instead of the `getAccount` method was because if I set state in there, then I'd end up calling setState during the render loop. This was causing an error. Having it in a `useEffect` guarantees that `setState` is not called during the render loop

packages/alchemy/src/config/actions/createAccount.ts
```diff
@@ -0,0 +1,96 @@
...
+  SupportedAccounts,
+} from "../types";
+import { getSignerStatus } from "./getSignerStatus.js";
+
+export type AccountConfig<TAccount extends SupportedAccountTypes> =
+  TAccount extends "LightAccount"
+    ? Omit<CreateLightAccountParams, "signer" | "transport" | "chain">
+    : Omit<
+        CreateMultiOwnerModularAccountParams,
+        "signer" | "transport" | "chain"
```
moldy530, line 22: actually, we shouldn't omit the `signer` here and instead make it optional. that would allow someone to create a session key signer based account

packages/alchemy/src/config/store/types.ts
```diff
@@ -0,0 +1,34 @@
+import type { Mutate, StoreApi } from "zustand";
+import type { ClientWithAlchemyMethods } from "../../client/types";
+import type { AlchemySigner, AlchemySignerStatus, User } from "../../signer";
+import type { SupportedAccount, SupportedAccountTypes } from "../types";
+
+export type AccountState<TAccount extends SupportedAccountTypes> =
```
moldy530, line 6: nit: maybe add one more here called `disconnected` which has account as undefined

packages/alchemy/src/config/store/types.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+  | {
+      status: "READY";
+      account: SupportedAccount<TAccount>;
+    };
+
+export type ClientState = {
+  signer: AlchemySigner;
+  user?: User;
+  signerStatus: AlchemySignerStatus;
```
moldy530, line 19: todo: change this to an object similar to:
```
{
  isInitializing: boolean;
  isAuthenticating: boolean;
  ...
  status: AlchemySignerStatus;
}
```

packages/alchemy/src/config/types.ts
```diff
@@ -0,0 +1,46 @@
...
+  bundlerClient: ClientWithAlchemyMethods;
+  signer: AlchemySigner;
+  coreStore: CoreStore;
+  clientStore: ClientStore;
+};
+
+export type CreateConfigProps = ConnectionConfig & {
+  chain: Chain;
+  sessionConfig?: AlchemySignerParams["sessionConfig"];
+  signerConnection?: ConnectionConfig;
```
moldy530, line 39: this can be used as an override if you have different rpc urls. This is most useful if devs are doing what they shouldn't be doing and putting their API keys in the front end

packages/alchemy/src/react/context.ts
```diff
@@ -0,0 +1,46 @@
...
+  return context;
+};
+
+export const AlchemyAccountProvider = ({
+  config,
+  queryClient,
+  children,
+}: React.PropsWithChildren<AlchemyAccountsProviderProps>) => {
+  // Note: we don't use .tsx because we don't wanna use rollup or similar to bundle this package.
+  // This lets us continue to use TSC for building the packages which preserves the "use client" above
```
moldy530, line 40: to expand a bit more, when I first prototyped this outside of this repo, I was initially using rollup to build. This was fine except that it wouldn't preserve the `use client` directive so some hooks weren't working well in nextjs. 

We'll need to figure this out when we add UI components, but for hooks I think this is fine for now

packages/alchemy/src/react/hooks/useSmartAccountClient.ts
```diff
@@ -0,0 +1,83 @@
...
+      Transport,
+      Chain,
+      SupportedAccount<TAccount>
+    >,
+    "rpcUrl" | "chain" | "apiKey" | "jwt" | "account"
+  > & { account: SupportedAccount<TAccount> };
+
+export type ClientActions<TAccount extends SupportedAccountTypes> =
+  TAccount extends "LightAccount"
+    ? BaseAlchemyActions<Chain, LightAccount<AlchemySigner>> &
```
moldy530, line 38: this isn't necessary in @adam-alchemy's multisig branch because we simplify it a bit to not require passing in the `BaseActions`

moldy530: so will need to update here once that's merged in (or fix in his branch if this is merged in first)

packages/alchemy/src/react/hooks/useUser.ts
```diff
@@ -0,0 +1,15 @@
...
+import { useSyncExternalStore } from "react";
+import { watchUser } from "../../config/actions/watchUser.js";
+import { useAlchemyAccountContext } from "../context.js";
+
+export const useUser = () => {
+  const { config } = useAlchemyAccountContext();
+
+  return useSyncExternalStore(
+    watchUser(config),
+    () => config.clientStore?.getState().user,
```
moldy530, line 12: TODO: need to throw client only error above

moldy530, line 12: nit: move this into a `getUser` action

packages/alchemy/src/signer/client/index.ts
```diff
@@ -80,11 +80,11 @@ export class AlchemySignerClient {
     this.rootOrg = rootOrgId;
   }
 
-  private get user() {
+  public get user() {
     return this._user;
   }
 
-  private set user(user: User | undefined) {
+  public set user(user: User | undefined) {
```
moldy530, line 87: undo this

packages/alchemy/src/signer/signer.ts
```diff
@@ -276,13 +276,14 @@ export class AlchemySigner
...
       this.sessionManager.setTemporarySession({ orgId });
+      this.store.setState({ status: AlchemySignerStatus.AWAITING_EMAIL_AUTH });
 
       // We wait for the session manager to emit a connected event if
       // cross tab sessions are permitted
       return new Promise<User>((resolve) => {
         const removeListener = this.sessionManager.on(
           "connected",
-          (session) => {
+          async (session) => {
```
moldy530, line 286: undo

packages/alchemy/src/config/actions/createAccount.ts
moldy530: not gonna do this now actually as it requires customizing signer in a lot of other places. opting for simplicity over flexibility / power in this first approach

packages/alchemy/src/react/hooks/useSmartAccountClient.ts
```diff
@@ -0,0 +1,97 @@
...
+        account,
+        ...params,
+      })
+        .extend(multiOwnerPluginActions)
+        .extend(pluginManagerActions)
+        .extend(accountLoupeActions);
+    default:
+      throw new Error("Unsupported account type");
+  }
+}
```
moldy530, line 97: Before this impl, I wasn't using an overloaded function with the super general generics and I was having to do some typecasting of the `account` when creating the client so that typescript wouldn't yell at me. And even then, I was still getting type errors.

Adding this function overload where the generics are union types of the possible inputs / outputs allows you to define a typesafe api for the consumer while also maintaining type safety in the implementation without having to typecast. 

We do this in places like `createSmartAccountClient` too, but I don't think it really clicked until I needed to return to vastly different types like I do here. We similarly do that in https://github.com/alchemyplatform/aa-sdk/pull/514 which allows us to handle various EP versions here: https://github.com/alchemyplatform/aa-sdk/pull/544

packages/alchemy/src/config/actions/createAccount.ts
moldy530: if we want session keys, we can always add a hook for that later

packages/alchemy/src/config/actions/createAccount.ts
```diff
@@ -0,0 +1,95 @@
...
+    accounts: {
+      ...state.accounts,
+      [type]: {
+        status: "INITIALIZING",
+        account: accountPromise,
+      },
+    },
+  }));
+
+  const account = await accountPromise;
```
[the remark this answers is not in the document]
moldy530: oh yea good catch

packages/alchemy/src/config/actions/watchAccount.ts
```diff
@@ -0,0 +1,30 @@
...
+      ({ accounts }) => accounts[type],
+      onChange,
+      {
+        fireImmediately: true,
+        equalityFn(a, b) {
+          return (
+            a?.status === b?.status ||
+            (a?.status === "READY" &&
+              b?.status === "READY" &&
+              a.account.address === b.account.address)
```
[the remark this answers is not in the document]
moldy530: oops yea this is wrong.

The comparison is meant to be for status === ready also compare their addresses otherwise only compare their statuses... which means I need to swap the order here

packages/alchemy/src/react/hooks/useSmartAccountClient.ts
```diff
@@ -0,0 +1,97 @@
...
+import { useBundlerClient } from "./useBundlerClient.js";
+
+export type UseSmartAccountClientProps<
+  TTransport extends Transport = Transport,
+  TChain extends Chain | undefined = Chain | undefined,
+  TAccount extends SupportedAccounts = SupportedAccounts
+> = Omit<
+  AlchemySmartAccountClientConfig<TTransport, TChain, TAccount>,
+  "rpcUrl" | "chain" | "apiKey" | "jwt" | "account"
+> & { account: TAccount };
```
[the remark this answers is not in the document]
moldy530: that does mean that the result of `useSmartAccountClient` can return a nullable client, or maybe I can have it return a client that accepts either type of account. the tricky part is adding the actions to the generic client... if no account should that return a client which has both LA and MA actions attached?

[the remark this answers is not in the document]
moldy530: I went with returning a nullable account and its usage is here:
https://github.com/alchemyplatform/embedded-accounts-demo/pull/4/files#diff-36b09641ed3ec22dbde5def2ae2d515f826edc07f78919d2cc16024701f8a785R33-R36

I kinda like this because it means you can use the `useAccount` hook in the same component as the `useSmartAccountClient` hook. It also means that the developer can await the result of `useAccount` before using the client.

maybe we should bring some of the functionality of `useAccount` into `useSmartAccountClient`. Instead of calling it with an account instance you call it with an account type like with `useAccount` and associated params. That would turn this hook into one line for usage rather than two lines since you always need to `useAccount` to `useSmartAccountClient`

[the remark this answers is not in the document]
moldy530: ok I ended up doing the last bit here instead. This hook now takes in the same params as `useAccount` + configs for the `smartAccountClient`. devex reflected here: https://github.com/alchemyplatform/embedded-accounts-demo/pull/4/files#diff-36b09641ed3ec22dbde5def2ae2d515f826edc07f78919d2cc16024701f8a785R32-R34

feels much better, but lemme know if you have any reservations with this approach

packages/alchemy/src/signer/session/manager.ts
```diff
@@ -232,5 +232,14 @@ export class SessionManager {
 
       this.setSession({ type: "passkey", user });
     });
+
+    window.addEventListener("storage", (e: StorageEvent) => {
```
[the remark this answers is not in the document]
moldy530: oh speaking of worth of words! would be good to understand from your lens as a potential consumer of this what we should focus on supporting next for your use case. can be a good way to dogfood this; ideally, we can replace most of your aa-sdk usage with hooks from this package

packages/alchemy/src/react/hooks/useSmartAccountClient.ts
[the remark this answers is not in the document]
dphilipson: Yeah, I like this! This looks pretty similar to the hook I ended up with in Worth of Words which returns a nullable account object and it works pretty smoothly there.

dphilipson (approved):
I really like the way this ended up! That's a really nice organization to have the state management outside of React. I haven't seen that before but it feels like everyone should be doing it.

avasisht23 (approved):
some comments here, but looks good!

packages/alchemy/src/react/hooks/useAuthenticate.ts
```diff
@@ -0,0 +1,37 @@
...
+  } = useMutation(
+    {
+      mutationFn: async (authParams: AuthParams) => {
+        if (!signer) {
+          throw new ClientOnlyPropertyError("signer");
+        }
+
+        const user = await signer.authenticate(authParams);
+
+        return user;
```
avasisht23, line 26: if you want you can just return the promise right?
```suggestion
        return signer.authenticate(authParams);
```

packages/alchemy/src/config/actions/watchAccount.ts
```diff
@@ -0,0 +1,30 @@
...
+      ({ accounts }) => accounts[type],
+      onChange,
+      {
+        fireImmediately: true,
+        equalityFn(a, b) {
+          return (
+            (a?.status === "READY" &&
+              b?.status === "READY" &&
+              a.account.address === b.account.address) ||
+            a?.status === b?.status
```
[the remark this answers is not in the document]
moldy530: 🤦 yea you're right what you've got is what I want. wtf was I doing lol

packages/alchemy/package.json
```diff
@@ -56,8 +69,22 @@
...
+    "react": {
+      "optional": true
+    },
+    "react-dom": {
+      "optional": true
+    },
+    "@tanstack/react-query": {
+      "optional": true
+    }
+  },
```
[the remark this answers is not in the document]
moldy530: this is the way you're intended to set optional deps that are also peer dependencies

packages/core/src/account/smartContractAccount.ts
```diff
@@ -183,15 +183,16 @@ export async function toSmartContractAccount<
   SmartContractAccount<Name>
 > {
   const client = createBundlerClient({
-    transport,
+    // we set the retry count to 0 so that viem doesn't retry during
+    // getting the address. That call always reverts and without this
+    // viem will retry 3 times, making this call very slow
+    transport: (opts) => transport({ ...opts, chain, retryCount: 0 }),
```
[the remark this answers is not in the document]
moldy530: the only calls this client makes are for the entrypoint. all other calls are made by the smart account client usually

packages/alchemy/src/react/hooks/useSigner.ts
```diff
@@ -0,0 +1,19 @@
...
+  const { config } = useAlchemyAccountContext();
+
+  // TODO: figure out how to handle this on the server
+  // I think we need a version of the signer that can be run on the server that essentially no-ops or errors
+  // for all calls
+  return useSyncExternalStore(
+    watchSigner(config),
+    () => config.signer,
+    // We don't want to return null here, should return something of type AlchemySigner
+    () => null
```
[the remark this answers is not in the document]
moldy530: it's not just that, but server-side state becomes global to the application. that means you can run into weird issues where you accidentally log a user in server side and that has all your other requests from other users authenticated as that user

packages/alchemy/src/react/hooks/useSignerStatus.ts
```diff
@@ -0,0 +1,25 @@
...
+  isDisconnected: false,
+};
+
+export const useSignerStatus = () => {
+  const { config } = useAlchemyAccountContext();
+
+  return useSyncExternalStore(
+    watchSignerStatus(config),
+    () => getSignerStatus(config),
+    () => serverStatus
```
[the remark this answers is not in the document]
moldy530: this is requirement of the `useSyncExternalStore` hook. the value should be static

packages/alchemy/src/react/hooks/useAuthenticate.ts
moldy530: oh yea definitely... not sure why I was doing this

packages/alchemy/src/react/context.ts
moldy530: the hooks files don't actually have any `jsx` syntax in them so we don't have to do that there

packages/alchemy/src/config/store/client.ts
```diff
@@ -0,0 +1,142 @@
...
+
+// This is done this way to avoid issues with React requiring static state
+const staticState: AccountState<SupportedAccountTypes> = {
+  status: "DISCONNECTED",
+  account: undefined,
+};
+
+export const defaultAccountState = <
+  T extends SupportedAccountTypes
+>(): AccountState<T> => staticState;
```
avasisht23, line 96: does this need a generic?

avasisht23, line 96: does this need a generic?

packages/alchemy/src/react/hooks/useSignerStatus.ts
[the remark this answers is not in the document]
avasisht23: sorry, i mean instead of null - per your comment

packages/alchemy/package.json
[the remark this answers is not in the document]
avasisht23: oh interesting, should we move the optDeps into this too then at some point?

packages/alchemy/src/react/hooks/useSignerStatus.ts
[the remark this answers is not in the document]
moldy530: oh! yea I had it as `null` initially, but this was causing some flashes of content or a react error because on the hydrated state wasn't matching the server state

packages/alchemy/src/config/store/client.ts
moldy530: yea because we don't have a default account type in the code base. if we want to default to MOMA or LA then we can update this to work on that generic

packages/alchemy/package.json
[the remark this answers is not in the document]
moldy530: In this case I need to pin the `react-query` dep to a specific version. But yea I think we should do that for the `aa-signers` repo. those optional deps probably should be optional peer deps instead of optional deps

packages/alchemy/src/config/store/client.ts
```diff
@@ -0,0 +1,141 @@
...
+        iframeContainerId,
+      },
+    },
+    sessionConfig,
+  });
+
+  const search = new URLSearchParams(window.location.search);
+  if (search.has("bundle")) {
+    signer.authenticate({ type: "email", bundle: search.get("bundle")! });
+  }
```
avasisht23, line 71: ultra nit: but "bundle" is a keyword and might be worth turning into a const that we can update in case we want to modify it in the future more easily if our email redirect flow changes.

not necessary to do in this pr

avasisht23 (dismissed):

packages/alchemy/src/config/store/client.ts
moldy530: oh yea good shout

avasisht23 (approved):

PR #544, opened 2024-04-02
test: make sure generic typing on EP version works

moldy530:
# Pull Request Checklist


packages/accounts/src/light-account/account.ts
```diff
@@ -60,15 +70,16 @@ export type CreateLightAccountParams<
   factoryAddress?: Address;
   initCode?: Hex;
   version?: LightAccountVersion;
-  entryPoint?: EntryPointParameter<TEntryPointVersion>["entryPoint"];
+  entryPoint?: EntryPointDef<TEntryPointVersion, Chain>;
```
moldy530, line 73: note: this is required for TS to infer the types correctly

dphilipson:
Thanks for writing this! It might be worth turning these into actual tests since they're already written. To check that an expression typechecks without evaluating it, you can put it in the body of a function that you don't call.

moldy530:
@dphilipson ah yea, I never intended to merge this in, but we should add these tests to the PR that ends up adding Light Account V2 support.

cc @denniswon 

Will leave this PR open to be used as reference since it's a starting point for light account v2

denniswon (changes requested):

packages/accounts/src/light-account/account.ts
```diff
@@ -276,3 +286,32 @@ export async function createLightAccount({
...
+  "0.6.0"
+> = await createLightAccount({
+  transport: http(),
+  chain: sepolia,
+  signer: LocalAccountSigner.privateKeyToAccountSigner(generatePrivateKey()),
+});
+
+const lightAccount07: LightAccount<
+  LocalAccountSigner<PrivateKeyAccount>,
+  "0.7.0"
```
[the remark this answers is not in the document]
moldy530: no you don't need this, this is just here to serve as a test that the output is correctly typed. if this type didn't match what the function outputted then you'd get an error. see @dphilipson's comment above in creating tests that do this

denniswon (commented):
But get the idea. :)

packages/accounts/src/light-account/account.ts
```diff
@@ -65,13 +76,24 @@ export type CreateLightAccountParams<
...
+    ? {
+        entryPoint?: EntryPointDef<TEntryPointVersion, Chain>;
+        version?: Extract<LightAccountVersion, `v1${string}`>;
+      }
+    : TEntryPointVersion extends "0.7.0"
+    ? {
+        entryPoint: EntryPointDef<TEntryPointVersion, Chain>;
+        version: Exclude<LightAccountVersion, `v1${string}`>;
+      }
+    : { entryPoint?: never; version?: never });
```
denniswon, line 91: you can just do:

```
EQ<TEntryPointVersion, DefaultEntryPointVersion> extends true
  ? {
      entryPoint?: EntryPointDef<TEntryPointVersion, TChain>;
    }
  : {
      entryPoint: IsOneOf<TEntryPointVersion, EntryPointVersion> extends true
        ? EntryPointDef<TEntryPointVersion, TChain>
        : never;
    };
```

This is what I had commented out before.

packages/accounts/src/light-account/account.ts
```diff
@@ -65,13 +76,24 @@ export type CreateLightAccountParams<
...
 
 export async function createLightAccount<
```
denniswon, line 93: Screenshot 2024-04-03 at 11 37 50 PM

Not sure why, but if I have the same approach for simple account on aa-core, typescript errors saying the signatures are not compatible for these overloaded functions.

denniswon: huh it turns out to be zod typescript issue.

packages/accounts/src/light-account/account.ts
```diff
@@ -65,13 +76,24 @@ export type CreateLightAccountParams<
...
+    ? { entryPoint: never; version?: never }
+    : TEntryPointVersion extends "0.6.0"
+    ? {
+        entryPoint?: EntryPointDef<TEntryPointVersion, Chain>;
+        version?: Extract<LightAccountVersion, `v1${string}`>;
+      }
+    : TEntryPointVersion extends "0.7.0"
+    ? {
+        entryPoint: EntryPointDef<TEntryPointVersion, Chain>;
+        version: Exclude<LightAccountVersion, `v1${string}`>;
```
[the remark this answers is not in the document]
moldy530: I think it's safe to say that v1 versions are the only ones not supported on 0.6.0

PR #547, opened 2024-04-04
feat: entrypoint v0.7 support with strong typing of entry point version

denniswon:
```
// Example usage:

const a = await createSimpleSmartAccount({
  chain: sepolia,
  signer: LocalAccountSigner.mnemonicToAccountSigner("OWNER_MNEMONIC"),
  factoryAddress: getDefaultSimpleAccountFactoryAddress(sepolia),
  transport: http(),
});

const b = await createSimpleSmartAccount({
  chain: sepolia,
  signer: LocalAccountSigner.mnemonicToAccountSigner("OWNER_MNEMONIC"),
  factoryAddress: getDefaultSimpleAccountFactoryAddress(sepolia),
  transport: http(),
  entryPoint: getEntryPoint(sepolia),
});

const c = await createSimpleSmartAccount({
  chain: sepolia,
  signer: LocalAccountSigner.mnemonicToAccountSigner("OWNER_MNEMONIC"),
  factoryAddress: getDefaultSimpleAccountFactoryAddress(sepolia),
  transport: http(),
  entryPoint: getEntryPoint(sepolia, { version: "0.7.0" }),
});
const ep = c.getEntryPoint()
type d = typeof ep.version // "0.7.0"

```

SmartAccountClient, BundlerClient, etc. should all be automatically typed just based on the entrypoint version passed into the `createSimpleSmartAccount` method.

cc: @mokok123 @dancoombs 

# Pull Request Checklist


packages/core/src/account/smartContractAccount.ts
```diff
@@ -115,7 +116,7 @@ export type ToSmartContractAccountParams<
   Name extends string = string,
   TTransport extends Transport = Transport,
   TChain extends Chain = Chain,
-  TEntryPointVersion extends EntryPointVersion = EntryPointVersion
+  TEntryPointVersion extends EntryPointVersion = DefaultEntryPointVersion
```
moldy530, line 119: I think we need to leave the base definition as it was here

packages/core/src/account/smartContractAccount.ts
```diff
@@ -214,7 +215,7 @@ export async function toSmartContractAccount<
 export async function toSmartContractAccount({
   transport,
   chain,
-  entryPoint,
+  entryPoint = getEntryPoint(chain),
```
moldy530, line 218: we should leave this as required for this function as it was before because this way it can be easy for someone using `toSmartContractAccount` to omit it.

imagine for LA we forget to pass the EP down through to the toSCA call. user passes in EP with address override but still is valid 0.6. there will be no type errors in the LA definition or in the caller's call, but their override doesn't get used

packages/accounts/src/light-account/account.ts
moldy530, line 1: we shouldn't change this file yet if it doesn't fully support 0.7

moldy530 (changes requested):
overall LGTM, just a couple nits

packages/core/src/account/smartContractAccount.ts
denniswon: typescript fails. because function overloaded without typing defaults to default entrypoint version, and we call toSmartContractAccount in that impl. so it thinks the entrypoint is 0.6.0 (default). This is one caveat of the current approach.

denniswon: @moldy530

packages/core/src/account/smartContractAccount.ts
denniswon: make sense

packages/accounts/src/light-account/account.ts
denniswon: k. will move to light account v2 PR

packages/core/src/account/smartContractAccount.ts
moldy530: because there was a default EP being passed in yea, but if we don't pass in a default and make EP required then it's fine

packages/core/src/account/smartContractAccount.ts
```diff
@@ -115,11 +114,12 @@ export type ToSmartContractAccountParams<
   Name extends string = string,
   TTransport extends Transport = Transport,
   TChain extends Chain = Chain,
-  TEntryPointVersion extends EntryPointVersion = EntryPointVersion
+  TEntryPointVersion extends EntryPointVersion = DefaultEntryPointVersion
 > = {
   source: Name;
   transport: TTransport;
   chain: TChain;
+  entryPoint: EntryPointDef<TEntryPointVersion, TChain>;
```
moldy530, line 122: yea with this like this then setting the default for `TEntryPointVersion` `EntryPointVersion` works now

packages/core/src/account/simple.ts
```diff
@@ -186,6 +203,33 @@ export async function createSimpleSmartAccount<
...
+//
+// const c = await createSimpleSmartAccount({
+//   chain: sepolia,
+//   signer: LocalAccountSigner.mnemonicToAccountSigner("OWNER_MNEMONIC"),
+//   factoryAddress: getDefaultSimpleAccountFactoryAddress(sepolia),
+//   transport: http(),
+//   entryPoint: getEntryPoint(sepolia, { version: "0.7.0" }),
+// });
+// const ep = c.getEntryPoint()
+// type d = typeof ep.version // "0.7.0"
```
moldy530, line 235: nit: remove this or make this a `jsdoc` comment and use `@example` in the comment to include this code

moldy530 (commented):
one small nit around the comment and updating the EP version generic, but with the entryPoint parameter as required actually makes it not matter as much what the default for the generic is anymore as it will always get inferred when the caller calls it

packages/core/src/account/simple.ts
denniswon: done

moldy530 (approved):

PR #548, opened 2024-04-04
feat: light account v2 entrypoint v7 and multi owner update

denniswon:
cc: @moldy530 @dphilipson 

light account/multi owner light account unit & integration included in this PR.
Will update 1271 signing for version v2.0.0 for LA and MultiOwner LA once @dphilipson 's 1271 PR is included.

# Pull Request Checklist

---

## PR-Codex overview
The focus of this PR is to refactor file structures and update type definitions for Light Accounts and Smart Contracts.

### Detailed summary
- Added `toRecord` function in `utils/index.js`
- Updated `RecordableKeys` type definition in `types.ts`
- Renamed Light Account ABI and Factory ABI to `_v1`
- Updated imports and paths for Light Account and Smart Account client
- Modified test file paths for Light Account
- Added API key constants in e2e test constants file
- Refactored file structures for Light Account and Smart Account client
- Updated type definitions for Light Account version
- Updated method names and parameters in various files
- Added domain hashing function for ERC1271 in `hashDomain.ts`

> The following files were skipped due to too many changes: `packages/alchemy/src/config/actions/createAccount.ts`, `packages/core/src/actions/smartAccount/buildUserOperation.ts`, `packages/accounts/src/light-account/decorators/multiOwnerLightAccount.ts`, `packages/core/src/utils/index.ts`, `packages/alchemy/src/client/__snapshots__/multiOwnerLightAccountClient.test.ts.snap`, `packages/accounts/src/light-account/actions/updateOwners.ts`, `packages/alchemy/src/client/multiOwnerLightAccountClient.ts`, `packages/accounts/src/light-account/clients/multiOwnerLightAccount.ts`, `packages/accounts/src/utils/erc1271/erc1271.ts`, `packages/core/src/account/smartContractAccount.ts`, `packages/accounts/src/index.ts`, `packages/core/e2e-tests/simple-account-staging.e2e.test.ts`, `packages/accounts/src/light-account/types.ts`, `packages/accounts/src/light-account/abis/LightAccountFactoryAbi_v2.ts`, `packages/alchemy/src/client/multiOwnerLightAccountClient.test.ts`, `packages/accounts/src/light-account/accounts/account.ts`, `packages/accounts/src/light-account/abis/MultiOwnerLightAccountFactoryAbi.ts`, `packages/accounts/src/light-account/accounts/multiOwner.ts`, `packages/accounts/src/utils/erc1271/utils/hashTypedData.ts`, `packages/accounts/src/light-account/utils.ts`, `packages/accounts/src/light-account/accounts/base.ts`, `packages/alchemy/e2e-tests/multi-owner-light-account.e2e.test.ts`, `packages/accounts/src/light-account/abis/LightAccountAbi_v2.ts`, `packages/accounts/src/light-account/abis/MultiOwnerLightAccountAbi.ts`, `packages/accounts/src/light-account/e2e-tests/light-account.e2e.test.ts`, `packages/accounts/src/light-account/e2e-tests/multi-owner-light-account.e2e.test.ts`, `packages/accounts/src/light-account/e2e-tests/light-account-v2.e2e.test.ts`

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206981976511613


packages/accounts/src/light-account/utils.ts
```diff
@@ -95,21 +196,66 @@ export const LightAccountUnsupported1271Factories = new Set(
   LightAccountUnsupported1271Impls.map((x) => x.factoryAddress)
 );
 
-export const getLightAccountVersion = async <A extends SmartContractAccount>(
+export type LightAccountSource = "LightAccount" | "MultiOwnerLightAccount";
```
moldy530, line 199: I don't think we should combine the two into one factory. Ideally we have two separate definitions -- one for LA and one for MOLA. We did the same for multi owner ma and multi sig ma

packages/core/src/account/smartContractAccount.ts
```diff
@@ -213,7 +214,7 @@ export async function toSmartContractAccount<
 export async function toSmartContractAccount({
   transport,
   chain,
-  entryPoint,
+  entryPoint = getEntryPoint(chain),
```
moldy530, line 217: I thought we were leaving this as a required parameter based on convos in the simple account pr

packages/accounts/src/light-account/utils.ts
```diff
@@ -15,44 +16,100 @@ import {
...
+
+export type LightAccountVersionDef<
+  TEntryPointVersion extends EntryPointVersion
 > = {
+  factoryAddress: Address;
+  implAddress: Address;
+  entryPointVersion: OneOf<TEntryPointVersion, EntryPointVersion>;
+};
+
+export interface ILightAccountVersions {
```
moldy530, line 36: I'm not a huge fan of this setup where we use `keyof ` to represent the versions instead of the union type of strings before. We do this with `EntryPointRegistry` as well and it's doesn't seem idiomatic, as I don't see this pattern used anywhere else. @dphilipson would love your thoughts here as well

I'd rather we leverage the pattern that we had before and not try to merge MultiOwner and single owner LA into the same logic. Keeping them separate will be easier imo and we can share logic between them with helper functions. A good example of this is multisig MA vs multiowner MA.

One thing about using `keyof` is that it now masks the underlying accepted values from developers. If I hover over a type that uses this, I don't see `"v1.0.1" | ...` I just see, `keyof ...` as a developer I need to now go and find what that thing I'm taking `keyof` is if I want to understand the accepted values.

Separately having hard coded types like `LightAccountConfig<"v1.0.1", "0.6.0">` adds to the maintenance cost and is something that can easily be done with typescript over the generic anyways. You can restrict the versions where it matters (ie. LightAccount and LightAccountParams) with something like this:

```ts
LightAccount<
  TSigner extends SmartAccountSigner = SmartAccountSigner,
  TVersion extends LightAccountVersion = LightAccountVersion
  TEntryPointVersion extends EntryPointVersion = TVersion extends `v1${string}` ? "0.6.0" : "0.7.0"
>

// similar thing for LightAccountParams
```

Doing this means that supporting a new version is handled by just adding to the union string type and recording the new `impl` and `factoryAddresses`.

With the registry flow, you need to update the Registry type as well as the versions map. While that's still just 2 updates, the updates requires a bit more than in the above (this is more of a nit than anything else).

**The idiomatic typescript is more the feedback here:**
Consider `viem`'s `Client` type. It's generic on a lot of other things, but assume for a second it's only generic on `Transport` which can take multiple shapes, but let's stick to `HttpTransport | CustomTransport`. 

They just have their `Client` definition generic on `Transport` and when `HttpTransport` is used or `CustomTransport` is used the rest of the typing just follows that. We don't also maintain a `ClientRegistry` like this

```ts
type ClientRegistry = {
  HttpTransport: Client<HttpTransport>
  CustomTransport: Client<CustomTransport>
}
```

The reason is that it's an added layer of abstraction that becomes more restrictive and is already handled by typescript. Example usage:
```ts
function getClient<TTransport extends Transport>(transport: TTransport): ClientRegistery[TTransport];

// is equivalent to

function getClient<TTransport extends Transport>(transport: TTransport): Client<TTransport>;
```

packages/accounts/src/light-account/utils.ts
```diff
@@ -15,44 +16,100 @@ import {
...
+  LightAccount: {
+    "v1.0.1": LightAccountConfig<"v1.0.1", "0.6.0">;
+    "v1.0.2": LightAccountConfig<"v1.0.2", "0.6.0">;
+    "v1.1.0": LightAccountConfig<"v1.1.0", "0.6.0">;
+    "v2.0.0": LightAccountConfig<"v2.0.0", "0.7.0">;
```
moldy530, line 41: where is `LightAccountConfig` defined

packages/accounts/src/light-account/utils.ts
```diff
@@ -15,44 +16,100 @@ import {
...
 > = {
+  factoryAddress: Address;
+  implAddress: Address;
+  entryPointVersion: OneOf<TEntryPointVersion, EntryPointVersion>;
+};
+
+export interface ILightAccountVersions {
+  LightAccount: {
+    "v1.0.1": LightAccountConfig<"v1.0.1", "0.6.0">;
+    "v1.0.2": LightAccountConfig<"v1.0.2", "0.6.0">;
```
moldy530, line 39: we should still mark these as `deprecated` because they shouldn't be used going forward and we should signal to devs to migrate off of them (via upgrades)

packages/accounts/src/light-account/utils.ts
```diff
@@ -15,54 +16,201 @@ import {
...
-
-export const LightAccountVersions: Record<
-  LightAccountVersion,
-  {
-    factoryAddress: Address;
-    implAddress: Address;
-  }
+export type LightAccountType = "LightAccount" | "MultiOwnerLightAccount";
+
+export type AccountVersionDef<
```
moldy530, line 27: 👍 this is great, thanks for cleaning up the previous def which was just inlined

packages/accounts/src/light-account/utils.ts
denniswon: for multi owner and multi sig, the code and functionalities of the accounts were vastly different, but for Light vs multiownerLight, the only difference is the ownership function. Yes I could have a light account base for the common light account source, but I think using light account type param is cleaner.

packages/core/src/account/smartContractAccount.ts
denniswon: rebased to main where simple account pr is merged :)

denniswon:
@moldy530 @dphilipson this PR is ready for review. cc: @jaypaik @adam-alchemy

packages/accounts/src/light-account/utils.ts
denniswon: @dphilipson thoughts on this? @moldy530 here are the benefits of using typescript interface/type and using keyof to define versions (string). 
- we can enforce that keys do exist in the configured version definition maps and enforce fully defined versions are used by the users.
- We can distinguish between the value type of the keyed versions by doing `Interface[key]` and that will specifically give me the type of version def typed with the key.

dphilipson: For the entry point versions, it looks to me like you'd be able to do
```ts
export type EntryPointVersion = "0.6.0" | "0.7.0";
export type EntryPointRegistryBase<T> = Record<EntryPointVersion, T>;
```
Is there a reason not to do that?

The light account version case is a little trickier to write without the `keyof`. I might need to think about that one a bit. I definitely agree with Moldy that having the version type declared as unions improves error messages, but the setup with the registry with light account type keys is a pretty good one.

dphilipson: Also thinking about this a bit more, I'm not positive I like having the same function give you a LightAccount or MultiOwnerLightAccount differing just on one optional parameter. This seems like it could be confusing or error-prone for MultiOwnerLightAccount users, who can very easily forget to set that parameter, have everything typecheck, but have all their calls fail because they're using the wrong account type.

I feel the same way about defaulting to a certain version, actually- it would be really easy for a 2.0 user to forget to pass in the optional version parameter and then have everything cryptically fail. I wonder if it might make sense to deprecate the function where the light account version is optional and always require users to pass in the light account version.

moldy530: > I'm not positive I like having the same function give you a LightAccount or MultiOwnerLightAccount differing just on one optional parameter

yea this is an area where I could see devs getting confused. It also makes the docs slightly more complex because we have just one doc with a lot of branching sections. Would rather we be explicit for devs and just have `createLA` and `createMultiOwnerLA` 

> I feel the same way about defaulting to a certain version, actually- it would be really easy for a 2.0 user to forget to pass in the optional version parameter and then have everything cryptically fail

yea I agree we should deprecate the function that allows for optional version and force devs to pin their code to a specific version

packages/accounts/src/light-account/accounts/base.ts
```diff
@@ -134,22 +140,22 @@ export async function createLightAccount({
...
 
     // only upgrade undeployed accounts (storage 0) or deployed light accounts, error otherwise
     if (
       fromHex(storage, "number") !== 0 &&
       !implementationAddresses.some((x) => x === trim(storage))
     ) {
       throw new Error(
-        "could not determine if smart account implementation is light account"
+        `could not determine if smart account implementation is ${type} ${version}`
       );
```
moldy530, line 154: nit: we should convert this into a strongly typed error

packages/accounts/src/light-account/accounts/multiOwner.ts
```diff
@@ -0,0 +1,202 @@
...
+    const ownerAddress = await signer.getAddress();
+    // owners need to be dedupe + ordered in ascending order and not == to zero address
+    const owners_ = Array.from(new Set([...owners, ownerAddress]))
+      .filter((x) => hexToBigInt(x) !== 0n)
+      .sort((a, b) => {
+        const bigintA = hexToBigInt(a);
+        const bigintB = hexToBigInt(b);
+
+        return bigintA < bigintB ? -1 : bigintA > bigintB ? 1 : 0;
+      });
```
moldy530, line 141: we duplicate this logic in a lot of places (multisig, multiownerMA, and here). We should create a util for this

packages/accounts/src/light-account/accounts/multiOwner.ts
```diff
@@ -0,0 +1,202 @@
...
+    client,
+    entryPoint,
+    accountAddress,
+    getAccountInitCode,
+  });
+
+  const account = await createLightAccountBase<
+    Transport,
+    SmartAccountSigner,
+    "MultiOwnerLightAccount"
```
moldy530, line 163: it's kinda weird that this method doesn't take in the `type` parameter instead actually. If you passed in the `type` then you probs wouldn't need to explicitly specify the generics here

nit: If you passed in the `type` to this then you probs also could just pass in the version and type and let the `base` method handle doing the `AccountVersionRegistry` lookup

packages/accounts/src/light-account/clients/multiOwnerLightAccount.ts
```diff
@@ -0,0 +1,75 @@
...
+  TChain extends Chain | undefined = Chain | undefined,
+  TSigner extends SmartAccountSigner = SmartAccountSigner
+>(
+  args: CreateMultiOwnerLightAccountClientParams<Transport, TChain, TSigner>
+): Promise<
+  SmartAccountClient<
+    CustomTransport,
+    Chain,
+    MultiOwnerLightAccount<TSigner>,
+    SmartAccountClientActions<Chain, SmartContractAccount> &
```
moldy530, line 49: nit: you shouldn't need to specify `SmartAccountClientActions<Chain, SmartContractAccount>` anymore after the multisig PR. This will always be included in the actions list

packages/accounts/src/light-account/utils.ts
```diff
@@ -15,114 +14,294 @@ import {
...
+  TType extends LightAccountType,
+  TLightAccountVersion extends LightAccountVersion<TType>
+> = IAccountVersionRegistry[TType][TLightAccountVersion];
+
+/**
+ * Example Usage:
+ *
+ * type T = GetLightAccountVersionDefsForEntryPoint<"LightAccount", "0.6.0">;
+ * const t_1: T = AccountVersionRegistry.LightAccount["v1.0.2"]; // compiles
+ * const t_2: T = AccountVersionRegistry.LightAccount["v2.0.0"]; // errors
```
moldy530, line 201: jsdocs supports passing in examples so that they're well formatted

you can use the `@example` syntax

moldy530: https://jsdoc.app/tags-example

packages/accounts/src/light-account/utils.ts
moldy530, line 1: nit: this file has gotten really large. It probably makes sense to break some of this tuff out.

Some suggestions on breaking things apart:

Things like the type defintions should probably be moved either into `light-account/types.ts` or `light-account/version.ts`

the methods for getting the version can live in a `light-account/utils/version.ts`

packages/alchemy/src/client/multiOwnerLightAccountClient.ts
```diff
@@ -0,0 +1,66 @@
...
+export async function createMultiOwnerLightAccountAlchemyClient<
+  TSigner extends SmartAccountSigner = SmartAccountSigner
+>(
+  params: AlchemyMultiOwnerLightAccountClientConfig<TSigner>
+): Promise<
+  AlchemySmartAccountClient<
+    CustomTransport,
+    Chain | undefined,
+    MultiOwnerLightAccount<TSigner>
+  >
```
moldy530, line 41: you need to specify the `actions` generic on this client or else you won't get the completions

packages/alchemy/src/client/lightAccountClient.ts
```diff
@@ -25,17 +25,21 @@ export type AlchemyLightAccountClientConfig<
...
   TSigner extends SmartAccountSigner = SmartAccountSigner
 >(
   params: AlchemyLightAccountClientConfig<TSigner>
-) => Promise<
+): Promise<
   AlchemySmartAccountClient<
     CustomTransport,
     Chain | undefined,
     LightAccount<TSigner>
   >
```
moldy530, line 37: ah yea same comment here, I think this is missing the actions type. I wonder how this was working before 🤔

packages/core/src/account/smartContractAccount.ts
moldy530: can we double check this? It's still an optional param here, this should be required

packages/accounts/src/light-account/utils.ts
```diff
@@ -15,114 +14,294 @@ import {
...
+          impl: "0xae8c656ad28F2B59a196AB61815C16A0AE1c3cba".toLowerCase() as Address,
+        })
+      ),
+      entryPointVersion: "0.6.0",
+    },
+    "v2.0.0": {
+      type: "LightAccount",
+      version: "v2.0.0",
+      address: toRecord<Chain, "id", { factory: Address; impl: Address }>(
+        supportedChains,
```
moldy530, line 141: we should limit to this just the chain it's currently deployed on

packages/accounts/src/light-account/utils.ts
```diff
@@ -15,114 +14,294 @@ import {
...
+>(
+  type: TLightAccountType
+): LightAccountVersion<TLightAccountType> =>
+  (type === "LightAccount"
+    ? "v1.1.0"
+    : "v2.0.0") as LightAccountVersion<TLightAccountType>;
+
 /**
- * Utility method returning the default light account factory address given a {@link Chain} object
+ * Example Usage:
```
moldy530, line 180:
```suggestion
 * @example
```

packages/core/src/account/smartContractAccount.ts
denniswon: doubled checked. it's not optional param, and for smartcontractaccount.ts, defaultEntrypointversion is not set as the entrypoint param is required.

packages/accounts/src/light-account/utils.ts
denniswon: we have explicit createLA/createMultiOwnerLA functions. Each call createLABase respectively which calls toSmartcontractaccount. cc: @dphilipson @moldy530

packages/accounts/src/light-account/accounts/account.test.ts
```diff
@@ -86,6 +87,7 @@ const givenConnectedProvider = ({
   createLightAccountClient({
     account: {
       signer,
+      type: "LightAccount",
```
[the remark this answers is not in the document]
denniswon: nope. removed.

packages/accounts/src/light-account/utils.ts
denniswon: updated.

packages/accounts/src/light-account/utils.ts
denniswon: refactored

packages/accounts/src/light-account/utils.ts
denniswon: updated

denniswon:
@moldy530 per the comment made on the parent PR, https://github.com/alchemyplatform/aa-sdk/pull/514#issuecomment-2051167012

refactored this pr, moved all entrypoint v7 base branch related fixes/changes to the parent entrypoint v7 base PR. On this PR, light account v2 / multiowner light account v2 is supported, but per our sync today, only available on arbitrum sepolia, and other chains they fail.

Ready for final pass, and merge to main branch after the entrypoint v7 branch is merged to main.

When the contracts team update the 1271 signing with the vulnerability fix, we should do a fast follow to update the light account v2 code here with other chain supports (once light account v2 gets deployed to all chains supported)

cc: @dphilipson @dancoombs @jaypaik @adam-alchemy @rthomare

packages/accounts/src/light-account/abis/LightAccountFactoryAbi.ts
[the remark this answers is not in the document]
denniswon: there are both v1 and v2 factory and account abis.

packages/accounts/src/light-account/accounts/account.test.ts
```diff
@@ -47,9 +48,9 @@ describe("Light Account Tests", () => {
   it("should correctly encode transferOwnership data", async () => {
     const { account } = await givenConnectedProvider({ signer, chain });
     expect(
-      account.encodeTransferOwnership(
-        "0xdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"
-      )
+      (
+        account as LightAccount<SmartAccountSigner, "LightAccount">
```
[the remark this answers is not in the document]
denniswon: we don't. old changes. removed.

packages/accounts/src/light-account/accounts/account.ts
```diff
@@ -0,0 +1,174 @@
...
+  TEntryPointVersion
+> & {
+  encodeTransferOwnership: (newOwner: Address) => Hex;
+  getOwnerAddress: () => Promise<Address>;
+};
+
+export type CreateLightAccountParams<
+  TTransport extends Transport = Transport,
+  TSigner extends SmartAccountSigner = SmartAccountSigner,
+  TLightAccountVersion extends LightAccountVersion<"LightAccount"> = LightAccountVersion<"LightAccount">,
```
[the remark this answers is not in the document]
denniswon: they are decoupled already. think this comment was from the code before the updates made.

packages/alchemy/src/client/lightAccountClient.ts
denniswon: added action typing

packages/alchemy/src/client/multiOwnerLightAccountClient.ts
denniswon: done

packages/accounts/src/light-account/utils.ts
denniswon: done. only on arb sepolia

denniswon:
rebased master

dphilipson (approved):
I think this looks good. Let's merge it and cut an alpha release.

PR #553, opened 2024-04-08
feat: use erc-712 envelope in light account v2 1271 signatures

dphilipson:
Light account v2 uses a new format in its ERC-1271 signatures to prevent replay attacks where the same signed message could be approved by several accounts that shared an owner.

---

## PR-Codex overview
This PR enhances ERC1271 signing functionality in the accounts module by refactoring hashing and signing logic for typed data.

### Detailed summary
- Refactored hashing and signing logic for ERC1271 functionality.
- Added `hashTypedData` and `hashDomain` functions.
- Improved error handling and validation for typed data.
- Updated `createLightAccountBase` to support ERC1271 signing.
- Added `getErc1271SigningFunctions` to retrieve signing functions.

> The following files were skipped due to too many changes: `packages/accounts/src/utils/erc1271/utils/hashTypedData.ts`

### DIFF
--- a/packages/accounts/src/light-account/accounts/base.ts
+++ b/packages/accounts/src/light-account/accounts/base.ts
@@ -20,6 +20,7 @@ import {
   type Hex,
   type SignTypedDataParameters,
   type Transport,
+  concat,
 } from "viem";
 import {
   AccountVersionRegistry,
@@ -29,6 +30,13 @@ import {
   type LightAccountVersion,
   type LightAccountVersionDef,
 } from "../utils.js";
+import { getErc1271SigningFunctions } from "../../utils/erc1271/erc1271.js";
+
+enum SignatureType {
+  EOA = "0x00",
+  CONTRACT = "0x01",
+  CONTRACT_WITH_ADDR = "0x02",
+}
 
 export type LightAccountBase<
   TSigner extends SmartAccountSigner = SmartAccountSigner,
@@ -111,7 +119,7 @@ export async function createLightAccountBase({
   chain,
   signer,
   abi,
-  version: _version,
+  version: { version, type },
   entryPoint,
   accountAddress,
   getAccountInitCode,
@@ -120,8 +128,6 @@ export async function createLightAccountBase({
     transport,
     chain,
   });
-
-  const { version, type } = _version;
 
   const encodeUpgradeToAndCall = async ({
     upgradeToAddress,
@@ -161,7 +167,7 @@ export async function createLightAccountBase({
     });
   };
 
-  const signWith1271Wrapper = async (msg: Hex): Promise<`0x${string}`> => {
+  const signWith1271WrapperV1 = async (hashedMessage: Hex): Promise<Hex> => {
     return signer.signTypedData({
       // EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)
       // https://github.com/alchemyplatform/light-account/blob/main/src/LightAccount.sol#L236
@@ -169,17 +175,27 @@ export async function createLightAccountBase({
         chainId: Number(client.chain.id),
         name: type,
         verifyingContract: accountAddress,
-        version: version === "v2.0.0" ? "2" : "1",
+        version: "1",
       },
       types: {
         LightAccountMessage: [{ name: "message", type: "bytes" }],
       },
       message: {
-        message: msg,
+        message: hashedMessage,
       },
       primaryType: "LightAccountMessage",
     });
   };
+
+  const { signMessage: signMessageV2, signTypedData: signTypedDataV2 } =
+    getErc1271SigningFunctions({
+      accountAddress,
+      accountName: type,
+      accountVersion: "2",
+      chainId: chain.id,
+      signer,
+      wrapperTypeName: "LightAccountMessage",
+    });
 
   const account = await toSmartContractAccount({
     transport,
@@ -222,12 +238,11 @@ export async function createLightAccountBase({
         case "v1.0.2":
           throw new Error(`${type} ${version} doesn't support 1271`);
         case "v1.1.0":
-          return signWith1271Wrapper(hashMessage(message));
+          return signWith1271WrapperV1(hashMessage(message));
         case "v2.0.0":
-          // TODO: implement 1271 signing for v2.0.0
-          throw new Error(
-            `${type} ${version} 1271 signing not implemented yet`
-          );
+          const signature = await signMessageV2({ message });
+          // TODO: handle case where signer is an SCA.
+          return concat([SignatureType.EOA, signature]);
         default:
           throw new Error(`Unknown version ${type} of ${version}`);
       }
@@ -243,12 +258,11 @@ export async function createLightAccountBase({
             `Version ${version} of LightAccount doesn't support 1271`
           );
         case "v1.1.0":
-          return signWith1271Wrapper(hashTypedData(params));
+          return signWith1271WrapperV1(hashTypedData(params));
         case "v2.0.0":
-          // TODO: implement 1271 signing for v2.0.0
-          throw new Error(
-            `${type} ${version} 1271 signing not implemented yet`
-          );
+          const signature = await signTypedDataV2(params);
+          // TODO: handle case where signer is an SCA.
+          return concat([SignatureType.EOA, signature]);
         default:
           throw new Error(`Unknown version ${version} of LightAccount`);
       }
--- a/packages/core/src/utils/index.ts
+++ b/packages/core/src/utils/index.ts
@@ -148,10 +148,10 @@ export const toRecord = <
   selector: K,
   fn: (item: T) => V
 ): Record<T[K], V> =>
-  array.reduce(
-    (acc, item) => ((acc[item[selector]] = fn(item)), acc),
-    {} as Record<T[K], V>
-  );
+  array.reduce((acc, item) => {
+    acc[item[selector]] = fn(item);
+    return acc;
+  }, {} as Record<T[K], V>);
 
 export * from "./bigint.js";
 export * from "./bytes.js";
--- a/packages/accounts/src/utils/erc1271/erc1271.ts
+++ b/packages/accounts/src/utils/erc1271/erc1271.ts
@@ -1,0 +1,97 @@
+import type {
+  Address,
+  SmartAccountSigner,
+  ToSmartContractAccountParams,
+} from "@alchemy/aa-core";
+import { hashDomain } from "./utils/hashDomain.js";
+import { concat, hashMessage, hashTypedData } from "viem";
+import { hashType, hashStruct } from "./utils/hashTypedData.js";
+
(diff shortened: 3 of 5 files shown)


packages/core/src/utils/index.ts
```diff
@@ -148,10 +148,10 @@ export const toRecord = <
...
   fn: (item: T) => V
 ): Record<T[K], V> =>
-  array.reduce(
-    (acc, item) => ((acc[item[selector]] = fn(item)), acc),
-    {} as Record<T[K], V>
-  );
+  array.reduce((acc, item) => {
+    acc[item[selector]] = fn(item);
+    return acc;
+  }, {} as Record<T[K], V>);
```
dphilipson, line 154: The linter was complaining about the use of the comma operator.

packages/accounts/src/light-account/accounts/base.ts
```diff
@@ -73,6 +81,7 @@ export type CreateLightAccountBaseParams<
   abi: Abi;
   signer: TSigner;
   accountAddress: Address;
+  accountType: TLightAccountType;
```
denniswon, line 84: No need. You can just do the type field of version (which is version def). I do that already so you can use the same type var.

packages/accounts/src/light-account/accounts/base.ts
```diff
@@ -181,6 +191,16 @@ export async function createLightAccountBase({
     });
   };
 
+  const { signMessage: signMessageV2, signTypedData: signTypedDataV2 } =
```
denniswon, line 194: Can we put this in sign with 1271 wrapper function and check the light account version and handle there?

packages/accounts/src/light-account/accounts/multiOwner.ts
```diff
@@ -70,7 +70,12 @@ export type CreateMultiOwnerLightAccountParams<
     TLightAccountVersion,
     TEntryPointVersion
   >,
-  "getAccountInitCode" | "entryPoint" | "version" | "abi" | "accountAddress"
+  | "getAccountInitCode"
+  | "entryPoint"
+  | "accountType"
```
denniswon, line 75: Ditto: no need for account type parameter

packages/accounts/src/light-account/erc1271/erc1271.ts
```diff
@@ -0,0 +1,97 @@
+import type {
+  Address,
+  SmartAccountSigner,
+  ToSmartContractAccountParams,
+} from "@alchemy/aa-core";
+import { hashDomain } from "./utils/hashDomain.js";
+import { concat, hashMessage, hashTypedData } from "viem";
+import { hashType, hashStruct } from "./utils/hashTypedData.js";
+
+export interface GetErc1271SigningFunctionsParams {
```
denniswon, line 10: Why interface not type?

packages/accounts/src/light-account/erc1271/erc1271.ts
```diff
@@ -0,0 +1,97 @@
...
+  accountVersion: string;
+  chainId: number;
+  signer: SmartAccountSigner;
+  wrapperTypeName: string;
+}
+
+export const getErc1271SigningFunctions = ({
+  accountAddress,
+  accountName,
+  accountVersion,
```
denniswon, line 22: Version is always 2 here so no need for the version input. Might want to guard only for version 2 here

denniswon (commented):
Overall LGTM! Thanks !

packages/accounts/src/light-account/erc1271/erc1271.ts
dphilipson: I'm trying to decouple these functions from light account, since this signing mechanism is a general proposal that other accounts can use. That's also the reason why I don't prepend the `SignatureType` enum in these functions but instead have the light account impl call these and prepend afterwards.

packages/accounts/src/light-account/erc1271/erc1271.ts
dphilipson: I thought using `interface` when possible was best practice. [Here's an explanation](https://www.typescriptlang.org/play#example/types-vs-interfaces) from the official site that recommends interfaces and gives some reasoning why.

packages/accounts/src/light-account/accounts/base.ts
dphilipson: Ah thanks, that's good to know. Will change.

packages/accounts/src/light-account/accounts/base.ts
dphilipson: I just tried it out and I'm not sure I like it that way. The issue is that then there's branching in two different places: inside `signWith1271Wrapper` and also in the switch statement on the version below. I think I like it more to have distinct functions for each version of the account.

To clarify this a bit, I tried renaming to `signWith1271WrapperV1` to make clear that it should only be used with V1 of accounts.

denniswon:
@jaypaik @howydev @adam-alchemy please take a look at this PR and would appreciate if you could review on the 1271 signature requirements. Thanks!

denniswon:
@dphilipson seems like unit tests did not pass

packages/accounts/src/light-account/erc1271/erc1271.ts
jaypaik: Wondering if it's worth fetching this and other domain separator values from `account.eip712Domain()` (if that context is readily available from somewhere). From 2.0.0, we support https://eips.ethereum.org/EIPS/eip-5267.

packages/accounts/src/light-account/accounts/base.ts
```diff
@@ -161,26 +167,36 @@ export async function createLightAccountBase({
...
+    getErc1271SigningFunctions({
+      accountAddress,
+      accountName: type,
+      accountVersion: "2",
+      chainId: chain.id,
+      signer,
+      wrapperTypeName: "LightAccountMessage",
```
jaypaik, line 197: If it simplifies things, this can also be fixed to something like "ERC1271Wrapper" across all accounts that the SDK supports.

jaypaik (commented):
lgtm! a couple comments.

packages/accounts/src/light-account/accounts/base.ts
dphilipson: Yea, that's the string that frangio uses in his example code.

I thought I'd go with `LightAccountMessage` here for two reasons:

1. The encoding fails if the wrapper type shares a name with a type in the payload. If all accounts using this encoding use the same wrapper type name, it makes it more likely that there will be a type name conflict at some point if someone needs to send a "meta" message containing a similarly encoded message.
2. This name will be displayed to end users in their wallet software. I thought "LightAccountMessage" is a little more friendly and also clearer about where this encoding is coming from.

But I don't feel too strongly about this. It doesn't really simplify things either way, we can make this string whatever we want.

packages/accounts/src/light-account/erc1271/erc1271.ts
dphilipson: I could be wrong, but my feeling is doing that would be less secure. The case I'm worried about is where you call the SDK's `createLightAccount` function but you pass it an account address which is not a `LightAccount`, either by mistake or because you were tricked by a bad actor. Then you would end up signing messages with the domain separator reported by that account, which could run into the replay attack we were trying to avoid in the first place if that domain separator is for a different SCA that shares an owner with the light account.

packages/accounts/src/light-account/erc1271/erc1271.ts
moldy530: oh nice, I've been wondering why I've seen uses of `interface` in a lot of alchemy's codebase and could never find a clear answer for why!

packages/accounts/src/light-account/erc1271/erc1271.ts
moldy530: > The case I'm worried about is where you call the SDK's createLightAccount function but you pass it an account address which is not a LightAccount

I'm still getting caught up on the changes, but in `main` the way LightAccount handles signing is by fetching the LightAccount version using the on-chain implAddress. This way, if you pass in an address that isn't LA, you would end up getting an error

packages/accounts/src/light-account/accounts/base.ts
```diff
@@ -222,12 +238,11 @@ export async function createLightAccountBase({
...
         case "v1.1.0":
-          return signWith1271Wrapper(hashMessage(message));
+          return signWith1271WrapperV1(hashMessage(message));
         case "v2.0.0":
-          // TODO: implement 1271 signing for v2.0.0
-          throw new Error(
-            `${type} ${version} 1271 signing not implemented yet`
-          );
+          const signature = await signMessageV2({ message });
+          // TODO: handle case where signer is an SCA.
```
moldy530, line 244: oof yea, I'm not sure how we can specify this... we can update the `SmartAccountWithSigner` type to include an enum of the signer type to be `EOA` or `SCA` and default to `EOA` in all cases for now which would be backwards compatible

packages/accounts/src/light-account/erc1271/erc1271.ts
moldy530: > I'm trying to decouple these functions from light account

nit: should we then move this into a `utils/signing` folder either in `core` or here in `accounts`?

packages/accounts/src/light-account/erc1271/utils/hashTypedData.ts
```diff
@@ -0,0 +1,278 @@
+// Copied from https://github.com/wevm/viem/blob/viem%402.9.12/src/utils/signature/hashTypedData.ts.
+// Unfortunately, viem does not export any of these functions
```
moldy530, line 2: wait yea they do, we use `hashTypedData` from `viem` in a few places. [eg `multi-owner/signer.ts`](https://github.com/alchemyplatform/aa-sdk/blob/main/packages/accounts/src/msca/plugins/multi-owner/signer.ts#L77)

dphilipson: Oh you're right, I should update the comment. Viem exports `hashTypedData`, but it doesn't export `hashStruct` or `hashDomain` which are the ones we need.

moldy530: ahhhh gotcha

packages/accounts/src/light-account/erc1271/erc1271.ts
dphilipson: > fetching the LightAccount version using the on-chain implAddress
That's a good idea, but I just looked up how it's implemented and the way it's written now it's good as a check against mistakes but doesn't provide security against malicious contracts, since it's looking up the impl storage slot and a malicious contract can put whatever it wants there.

If you wanted security you'd need to check the proxy codehash too, which I maybe wouldn't recommend since it could reject other proxy impls that have upgraded to light account.

PR #563, opened 2024-04-10
fix: add the signer header when using aa-alchemy

moldy530:
# Pull Request Checklist

---

## PR-Codex overview
This PR enhances Alchemy client functionality by adding headers update capability and improving smart account client handling.

### Detailed summary
- Added `updateHeaders` method to modify headers in `fetchOptions`
- Updated headers in `lightAccountClient.test.ts`
- Improved handling of headers in `bundlerClient.ts`
- Enhanced `rpcClient.ts` to update headers with `fetchOptions`
- Added `isSmartAccountWithSigner` and `getSignerTypeHeader` in `smartAccountClientFromRpc.ts`
- Updated tests in `smartAccountClient.test.ts` for header setting

### DIFF
--- a/packages/alchemy/src/client/rpcClient.ts
+++ b/packages/alchemy/src/client/rpcClient.ts
@@ -1,14 +1,21 @@
-import { createBundlerClient, type ConnectionConfig } from "@alchemy/aa-core";
-import { http, type Chain } from "viem";
+import {
+  createBundlerClient,
+  type ConnectionConfig,
+  type NoUndefined,
+} from "@alchemy/aa-core";
+import { http, type Chain, type HttpTransportConfig } from "viem";
 import { AlchemyChainSchema } from "../schema.js";
+import { VERSION } from "../version.js";
 import type { ClientWithAlchemyMethods } from "./types.js";
 
 export const createAlchemyPublicRpcClient = ({
   chain: chain_,
   connectionConfig,
+  fetchOptions = {},
 }: {
   connectionConfig: ConnectionConfig;
   chain: Chain;
+  fetchOptions?: NoUndefined<HttpTransportConfig["fetchOptions"]>;
 }): ClientWithAlchemyMethods => {
   const chain = AlchemyChainSchema.parse(chain_);
 
@@ -17,16 +24,27 @@ export const createAlchemyPublicRpcClient = ({
       ? `${chain.rpcUrls.alchemy.http[0]}/${connectionConfig.apiKey ?? ""}`
       : connectionConfig.rpcUrl;
 
+  fetchOptions.headers = {
+    ...fetchOptions.headers,
+    "Alchemy-AA-Sdk-Version": VERSION,
+  };
+
+  if (connectionConfig.jwt != null) {
+    fetchOptions.headers = {
+      ...fetchOptions.headers,
+      Authorization: `Bearer ${connectionConfig.jwt}`,
+    };
+  }
+
   return createBundlerClient({
     chain: chain,
-    transport: http(rpcUrl, {
-      ...(connectionConfig.jwt != null && {
-        fetchOptions: {
-          headers: {
-            Authorization: `Bearer ${connectionConfig.jwt}`,
-          },
-        },
-      }),
-    }),
-  });
+    transport: http(rpcUrl, { fetchOptions }),
+  }).extend(() => ({
+    updateHeaders(newHeaders: HeadersInit) {
+      fetchOptions.headers = {
+        ...fetchOptions.headers,
+        ...newHeaders,
+      };
+    },
+  }));
 };
--- a/packages/core/src/client/bundlerClient.ts
+++ b/packages/core/src/client/bundlerClient.ts
@@ -75,24 +75,25 @@ export function createBundlerClient(
 
   const client = (() => {
     if (resolvedTransport.config.type === "http") {
-      const { url, fetchOptions } = resolvedTransport.value as {
+      const { url, fetchOptions: fetchOptions_ } = resolvedTransport.value as {
         fetchOptions: HttpTransportConfig["fetchOptions"];
         url: string;
       };
+
+      const fetchOptions = fetchOptions_ ?? {};
+
+      if (url.toLowerCase().indexOf("alchemy") > -1) {
+        fetchOptions.headers = {
+          ...fetchOptions.headers,
+          "Alchemy-AA-Sdk-Version": VERSION,
+        };
+      }
 
       return createClient<Transport, Chain>({
         ...baseParameters,
         transport: http(url, {
           ...resolvedTransport.config,
-          fetchOptions: {
-            ...fetchOptions,
-            headers: {
-              ...fetchOptions?.headers,
-              ...(url.toLowerCase().indexOf("alchemy") > -1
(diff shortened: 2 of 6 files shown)


packages/alchemy/src/client/rpcClient.ts
```diff
@@ -17,16 +43,25 @@ export const createAlchemyPublicRpcClient = ({
...
-      ...(connectionConfig.jwt != null && {
-        fetchOptions: {
-          headers: {
-            Authorization: `Bearer ${connectionConfig.jwt}`,
-          },
-        },
-      }),
-    }),
-  });
+    transport: http(rpcUrl, { fetchOptions: dynamicFetchOptions }),
```
moldy530, line 58: viem's transport takes the fetch options and passes them down to the various functions that consume them. but it does so by doing things like 
```
const httpTransport = args => {
  const { fetchOptions } = args; 
  return () => fetchOptions.foo;
}
```

packages/core/src/client/bundlerClient.ts
```diff
@@ -79,20 +79,20 @@ export function createBundlerClient(
         fetchOptions: HttpTransportConfig["fetchOptions"];
         url: string;
       };
+      if (fetchOptions) {
+        fetchOptions.headers = {
+          ...fetchOptions.headers,
+          ...(url.toLowerCase().indexOf("alchemy") > -1
+            ? { "Alchemy-AA-Sdk-Version": VERSION }
+            : undefined),
+        };
```
dphilipson, line 88: (Nit) This looks a little odd in that copying the object would normally be done to avoid mutating the input, but here we're mutating the input anyways by setting `fetchOptions.headers` to a different value.

If you don't mind mutating the input, this might be a little easier to read as
```ts
if (fetchOptions && url.toLowerCase().indexOf("alchemy") > -1) {
  fetchOptions.headers["Alchemy-AA-Sdk-Version"] = VERSION;
}
```

packages/alchemy/src/client/rpcClient.ts
```diff
@@ -1,14 +1,40 @@
...
-import { http, type Chain } from "viem";
+import { http, type Chain, type HttpTransportConfig } from "viem";
 import { AlchemyChainSchema } from "../schema.js";
+import { VERSION } from "../version.js";
 import type { ClientWithAlchemyMethods } from "./types.js";
 
+export type DynamicFetchOptions = HttpTransportConfig["fetchOptions"] & {
+  get headers(): HeadersInit | undefined;
+  set headers(newFetchOptions: HeadersInit | undefined);
+};
```
dphilipson, line 10: What makes this type different from
```ts
HttpTransportConfig["fetchOptions"] & { headers: HeadersInit | undefined }
```
?

packages/alchemy/src/client/rpcClient.ts
```diff
@@ -1,14 +1,40 @@
...
+  return {
+    ...initialFetchOptions,
+    get headers() {
+      return headers;
+    },
+    set headers(newHeaders) {
+      headers = newHeaders;
+    },
+  };
+};
```
dphilipson, line 28: Similarly, why is this getter and setter different than the default behavior of a field? This function looks like it might be equivalent to
```ts
const createDynamicFetchOptions = (
  initialFetchOptions: HttpTransportConfig["fetchOptions"] = {}
): DynamicFetchOptions => {
  return {
    ...initialFetchOptions,
    headers: { ...initialFetchOptions.headers },
  };
};
```

dphilipson (commented):
Thanks for looking at this! I might need a bit of explanation for what's going on with the access functions.

packages/core/src/client/bundlerClient.ts
```diff
@@ -79,20 +79,20 @@ export function createBundlerClient(
         fetchOptions: HttpTransportConfig["fetchOptions"];
         url: string;
       };
+      if (fetchOptions) {
+        fetchOptions.headers = {
+          ...fetchOptions.headers,
+          ...(url.toLowerCase().indexOf("alchemy") > -1
+            ? { "Alchemy-AA-Sdk-Version": VERSION }
```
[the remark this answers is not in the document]
moldy530: this is what our logging is looking for so we can't really change it

packages/core/src/client/bundlerClient.ts
```diff
@@ -79,20 +79,20 @@ export function createBundlerClient(
...
+            : undefined),
```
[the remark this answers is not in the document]
moldy530: yea same here, this is what our logging is looking for

packages/alchemy/src/client/smartAccountClient.ts
```diff
@@ -143,8 +143,8 @@ export function createAlchemySmartAccountClient({
     gasManagerConfig,
     useSimulation,
     feeEstimator,
-    customMiddleware,
     gasEstimator,
     signUserOperation,
+    customMiddleware,
```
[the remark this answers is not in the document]
moldy530: ah it was left over. I initially made the changes here before realizing I should move it into the internal one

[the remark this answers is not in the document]
moldy530: ah it was left over. I initially made the changes here before realizing I should move it into the internal one

packages/alchemy/src/client/rpcClient.ts
moldy530: yea I think you might be right. will try it without the getter setters. initially I thought I might need a mutable variable that the getter and setter accessed instead of the default behaviour. I'll give it a shot now that I have tests and see what happens

moldy530: ok tested it out and you're right this works just fine by mutating the object directly. updated everything to make it cleaner

packages/core/src/client/bundlerClient.ts
moldy530: ah ok so I tried this out. if I do it the way you've got it here, I get a TS error that `HeadersInit` doesn't have a property `"Alchemy-AA-Sdk-Version"` so I can't index directly into it this way. So I still need to use the spread in the reassignment. I've cleaned it up though so it's much easier to read

denniswon:
Overall, LGTM. !

denniswon (approved):

packages/core/src/client/bundlerClient.ts
dphilipson: I think that type error might be highlighting a real bug, actually. `fetchOptions.headers` is type `HeadersInit`, which means it could also be a list of key-value pairs or an instance of the `Headers` class, which would make this code not work.

packages/alchemy/src/client/rpcClient.ts
```diff
@@ -17,16 +24,27 @@ export const createAlchemyPublicRpcClient = ({
...
-    }),
-  });
+    transport: http(rpcUrl, { fetchOptions }),
+  }).extend(() => ({
+    updateHeaders(newHeaders: HeadersInit) {
+      fetchOptions.headers = {
+        ...fetchOptions.headers,
+        ...newHeaders,
+      };
+    },
```
dphilipson, line 48: Like other comment, `HeadersInit` could be types that won't spread correctly here.

PR #565, opened 2024-04-11
feat: add useAddPasskey hook

avasisht23:
# Pull Request Checklist

### DIFF
--- a/packages/alchemy/src/react/hooks/useAddPasskey.ts
+++ b/packages/alchemy/src/react/hooks/useAddPasskey.ts
@@ -1,0 +1,45 @@
+"use client";
+
+import { useMutation, type UseMutateFunction } from "@tanstack/react-query";
+import { ClientOnlyPropertyError } from "../../config/errors.js";
+import { useAlchemyAccountContext } from "../context.js";
+import { useSigner } from "./useSigner.js";
+
+export type UseAddPasskeyResult = {
+  addPasskey: UseMutateFunction<
+    string[],
+    Error,
+    CredentialCreationOptions | undefined | void,
+    unknown
+  >;
+  isAddingPasskey: boolean;
+  error: Error | null;
+};
+
+export function useAddPasskey(): UseAddPasskeyResult {
+  const { queryClient } = useAlchemyAccountContext();
+  const signer = useSigner();
+
+  const {
+    mutate: addPasskey,
+    isPending: isAddingPasskey,
+    error,
+  } = useMutation(
+    {
+      mutationFn: async (params?: CredentialCreationOptions | void) => {
+        if (!signer) {
+          throw new ClientOnlyPropertyError("signer");
+        }
+
+        return signer.addPasskey(params ?? undefined);
+      },
+    },
+    queryClient
+  );
+
+  return {
+    addPasskey,
+    isAddingPasskey,
+    error,
+  };
+}
--- a/packages/alchemy/src/react/index.ts
+++ b/packages/alchemy/src/react/index.ts
@@ -7,6 +7,7 @@ export {
 export { NoAlchemyAccountContextError } from "./errors.js";
 export type * from "./hooks/useAccount.js";
 export { useAccount } from "./hooks/useAccount.js";
+export { useAddPasskey } from "./hooks/useAddPasskey.js";
 export type * from "./hooks/useAuthenticate.js";
 export { useAuthenticate } from "./hooks/useAuthenticate.js";
 export type * from "./hooks/useBundlerClient.js";


packages/alchemy/src/react/hooks/useAddPasskey.ts
```diff
@@ -0,0 +1,45 @@
...
+  const { queryClient } = useAlchemyAccountContext();
+  const signer = useSigner();
+
+  const {
+    mutate: addPasskey,
+    isPending: isAddingPasskey,
+    error,
+  } = useMutation(
+    {
+      mutationFn: async (params?: CredentialCreationOptions | void) => {
```
avasisht23, line 29: reason I did `... | void`: without it, react query makes you explicit do `undefined` for optional inputs to a mutation.

ref: https://github.com/TanStack/query/issues/1077#issuecomment-1698583282.

packages/alchemy/src/react/hooks/useAddPasskey.ts
```diff
@@ -0,0 +1,45 @@
...
+        if (!signer) {
+          throw new ClientOnlyPropertyError("signer");
+        }
```
moldy530, line 32: nit: you shouldn't need this because `useSigner` already handles that for you

moldy530: as a result you should be able to update this to just be:

```
useMutation({
  mutationFn: signer.addPasskey
})
```

moldy530 (commented):
Can we write a doc for this as well as part of this PR?

moldy530:
oh meta note on all of these (and I need to do this in my hooks as well). We need to support devs to pass `onSuccess`, `onError` and `onSettled` to the props.

Maybe we should have our hooks follow a format like:
```
const useSomeHook = (props: UseSomeHookProps, mutationOpts?: Omit<UseMutationOpts, "mutationFn" | "mutationKey">
```

packages/alchemy/src/react/index.ts
```diff
@@ -7,6 +7,7 @@ export {
 export { NoAlchemyAccountContextError } from "./errors.js";
 export type * from "./hooks/useAccount.js";
 export { useAccount } from "./hooks/useAccount.js";
+export { useAddPasskey } from "./hooks/useAddPasskey.js";
```
moldy530, line 10: also add the `export type * from "./hooks/myNewHook`

packages/alchemy/src/react/hooks/useAddPasskey.ts
moldy530: it's safe to assume that `useMutation` will run on the client only. so you can probs get away with doing `signer!.`

packages/alchemy/src/react/hooks/useAddPasskey.ts
```diff
@@ -0,0 +1,53 @@
...
+  const signer = useSigner();
+
+  const {
+    mutate: addPasskey,
+    isPending: isAddingPasskey,
+    error,
+  } = useMutation(
+    {
+      mutationFn: async (params: UseAddPasskeyParams) => {
+        return signer!.addPasskey(params ?? undefined);
```
avasisht23, line 41: see my comment in conversation on why i did this vs. your suggestion

packages/alchemy/src/react/hooks/useAddPasskey.ts
moldy530: oh i see, ok yea that makes sense

packages/alchemy/src/react/types.ts
```diff
@@ -0,0 +1,26 @@
...
+    | ((
+        data: TData | undefined,
+        error: Error | null,
+        variables: TVariable,
+        context: unknown
+      ) => unknown)
+    | undefined;
+} & Partial<
+  Omit<
+    UseMutationOptions<TData, Error, TVariable, unknown>,
```
moldy530, line 23: does this not already include the `on*` methods? can this type be simplified to just the `Omit<...>`?

avasisht23: Ah good catch didn't update this, but yup will change

site/react/useAddPasskey.md
```diff
@@ -0,0 +1,41 @@
...
+    - name: twitter:title
+      content: useAddPasskey
+  - - meta
+    - name: twitter:description
+      content: An overview of the useAddPasskey hook
+---
+
+# useAddPasskey
+
+The `useAddPasskey` hook adds a new passkey to a user's Embedded Account.
```
moldy530, line 23:
```suggestion
The `useAddPasskey` hook adds a new passkey to a user's Embedded Account.

::: warning
This requires your user to be logged in.
:::
```

moldy530 (approved):

site/react/useAddPasskey.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Import
```
moldy530, line 25: we need a `## Params` section as well to highlight the optional MutationFn params. We can just link out to React Query for that. 

To avoid duplication of code in all the hooks, you can use a `.md` snippet as per https://vitepress.dev/guide/markdown#markdown-file-inclusion

site/react/useAddPasskey.md
```diff
@@ -0,0 +1,41 @@
...
+
+<<< @/snippets/react/useAddPasskey.tsx
+
+## Return Type
+
+```ts
+import { type useAddPasskeyResult } from "@alchemy/aa-alchemy/react";
+```
+
+Returns a `string[]` list of all passkey authenticatorIds, including the newly added one.
```
moldy530, line 41: wait this isn't right, it should be the different properties returned by `useAddPasskey`

site/react/useAddPasskey.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+## Usage
+
+<<< @/snippets/react/useAddPasskey.tsx
+
+## Return Type
+
+```ts
+import { type useAddPasskeyResult } from "@alchemy/aa-alchemy/react";
```
moldy530, line 38:
```suggestion
import { type UseAddPasskeyResult } from "@alchemy/aa-alchemy/react";
```

site/react/useAddPasskey.md
avasisht23: doing this below

site/snippets/react/useAddPasskey.tsx
```diff
@@ -0,0 +1,12 @@
+import { useAddPasskey } from "@alchemy/aa-alchemy/react";
+
+export function ComponentWithAddPasskey() {
+  // Assumes the app has context of a signer with an authenticated user
```
rthomare, line 4: Can we add which context provider we explicitly? i.e. `CreateConfig`

packages/alchemy/src/react/types.ts
```diff
@@ -0,0 +1,11 @@
+import type { UseMutationOptions } from "@tanstack/react-query";
+
+export type BaseHookMutationArgs<
```
rthomare, line 3: nit would be good to add jsdocs here.

packages/alchemy/src/react/hooks/useAddPasskey.ts
```diff
@@ -0,0 +1,53 @@
+"use client";
+
+import { useMutation, type UseMutateFunction } from "@tanstack/react-query";
+import { useAlchemyAccountContext } from "../context.js";
+import type { BaseHookMutationArgs } from "../types.js";
+import { useSigner } from "./useSigner.js";
+
+export type UseAddPasskeyData = string[];
```
rthomare, line 8: nit jsdocs

packages/alchemy/src/react/hooks/useAddPasskey.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type UseAddPasskeyParams = CredentialCreationOptions | undefined | void;
```
rthomare, line 10: nit jsdocs

packages/alchemy/src/react/hooks/useAddPasskey.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type UseAddPasskeyMutationArgs = BaseHookMutationArgs<
```
rthomare, line 12: nit jsdocs

packages/alchemy/src/react/hooks/useAddPasskey.ts
```diff
@@ -0,0 +1,53 @@
...
+  UseAddPasskeyData,
+  UseAddPasskeyParams
+>;
+
+export type UseAddPasskeyResult = {
```
rthomare, line 17: nit jsdocs

packages/alchemy/src/react/hooks/useAddPasskey.ts
```diff
@@ -0,0 +1,53 @@
...
+    UseAddPasskeyData,
+    Error,
+    UseAddPasskeyParams,
+    unknown
+  >;
+  isAddingPasskey: boolean;
+  error: Error | null;
+};
+
+export function useAddPasskey(
```
rthomare, line 28: nit jsdocs

rthomare (approved):

rthomare:
I think this looks good. I'm not going to block on this because we haven't explicitly aligned as a team but I'd really like us to get in the habit of jsdocs. Viem does a good job and there is probably little work given AI can generate most of the text for us. 

Example https://github.com/wevm/viem/blob/9cae9b77b31c0a1388795de8056704cfde004260/src/actions/wallet/sendTransaction.ts#L86

packages/alchemy/src/react/hooks/useAddPasskey.ts
```diff
@@ -0,0 +1,53 @@
+"use client";
```
rthomare, line 1: why do we need useClient? is this for future proofing for a nextjs app?

avasisht23:
chatted [here](https://alchemyinsights.slack.com/archives/C0669LA6JQH/p1713306485257169?thread_ts=1713196802.873379&cid=C0669LA6JQH), will follow up after the stack merges with jsdocs everywhere

PR #568, opened 2024-04-11
feat: add useSignMessage and useSignTypedData hooks

avasisht23:
# Pull Request Checklist

---

## PR-Codex overview
The focus of this PR is to add new hooks `useSignMessage` and `useSignTypedData` for signing messages and typed data in Alchemy. 

### Detailed summary
- Added `useSignMessage` and `useSignTypedData` hooks for signing messages and typed data
- Updated error handling in hooks
- Created components for using the new hooks in React applications

> The following files were skipped due to too many changes: `site/react/useSignTypedData.md`


packages/alchemy/src/react/hooks/useSignMessage.ts
```diff
@@ -0,0 +1,72 @@
+"use client";
+
+import { useMutation, type UseMutateFunction } from "@tanstack/react-query";
+import type { Hex, SignableMessage } from "viem";
+import { useAlchemyAccountContext } from "../context.js";
+import { useBundlerClient } from "./useBundlerClient.js";
+import { type UseSmartAccountClientResult } from "./useSmartAccountClient.js";
+
+export type SignedMessageInfo = {
```
moldy530, line 9: nit: maybe we just call this `SignedMessage`

packages/alchemy/src/react/hooks/useSignMessage.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (!client) {
+          throw new Error("client must be defined");
+        }
+
+        const signature = await client.signMessageWith6492(params);
+        const isValid = await bundlerClient.verifyMessage({
+          address: client.getAddress(),
+          message: params.message,
+          signature,
+        });
```
moldy530, line 55: let's avoid doing this actually. this was only done in the demo app because I needed to be able to test. most flows by users won't be doing this

packages/alchemy/src/react/hooks/useSignTypedData.ts
```diff
@@ -0,0 +1,69 @@
...
+      mutationFn: async (params: SignTypedDataArgs) => {
+        if (!client) {
+          throw new Error("client must be defined");
+        }
+        const signature = await client.signTypedDataWith6492({ ...params });
+        const isValid = await bundlerClient.verifyTypedData({
+          address: client.getAddress(),
+          signature,
+          ...params.typedData,
+        });
```
moldy530, line 52: same here, let's not do this

packages/alchemy/src/react/hooks/useSignMessage.ts
```diff
@@ -0,0 +1,72 @@
...
+
+        return {
+          signature,
+          isValid,
+        };
```
moldy530, line 60: let's just return the signature here

packages/alchemy/src/react/hooks/useSignMessage.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export function useSignMessage({
+  client,
+}: UseSignMessageArgs): UseSignMessageResult {
+  const { queryClient } = useAlchemyAccountContext();
+  const bundlerClient = useBundlerClient();
+
+  const {
+    mutate: signMessage,
+    data: signedMessageInfo,
```
moldy530, line 40: I wonder if it makes sense to have this hook also return `signMessageAsync` so that a user can do `const signature = await signMessageAsync`?

packages/alchemy/src/react/hooks/useSignMessage.ts
```diff
@@ -0,0 +1,72 @@
...
+    isPending: isSigningMessage,
+    error,
+  } = useMutation(
+    {
+      mutationFn: async (params: SignMessageArgs) => {
+        if (!client) {
+          throw new Error("client must be defined");
```
moldy530, line 47: can we turn this into a strongly typed error?

packages/alchemy/src/react/hooks/useSignTypedData.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export function useSignTypedData({
+  client,
+}: UseSignTypedDataArgs): UseSignTypedDataResult {
+  const { queryClient } = useAlchemyAccountContext();
+  const bundlerClient = useBundlerClient();
+
+  const {
+    mutate: signTypedData,
+    data: signedTypedDataInfo,
```
moldy530, line 38: same comment as above to include `signTypedDataAsync` as an export as well

moldy530 (commented):
annnnnd docs

packages/alchemy/src/react/index.ts
```diff
@@ -14,6 +14,8 @@ export type * from "./hooks/useBundlerClient.js";
 export { useBundlerClient } from "./hooks/useBundlerClient.js";
 export { useDisconnectSigner } from "./hooks/useDisconnectSigner.js";
 export { useExportWallet } from "./hooks/useExportWallet.js";
+export { useSignMessage } from "./hooks/useSignMessage.js";
+export { useSignTypedData } from "./hooks/useSignTypedData.js";
```
moldy530, line 18: also add the `export type * from "./hooks/myNewHook`

packages/alchemy/src/react/hooks/useSignMessage.ts
avasisht23: agree with this

packages/alchemy/src/react/hooks/useSignTypedData.ts
```diff
@@ -0,0 +1,76 @@
...
+    SignTypedDataArgs,
+    unknown
+  >;
+  signedTypedData: UseSignTypedDataData | undefined;
+  isSigningTypedData: boolean;
+  error: Error | null;
+};
+
+export function useSignTypedData({
+  client,
```
moldy530, line 47: this one is missing the mutation params

site/react/useSignMessage.md
```diff
@@ -0,0 +1,41 @@
...
+    - name: twitter:title
+      content: useSignMessage
+  - - meta
+    - name: twitter:description
+      content: An overview of the useSignMessage hook
+---
+
+# useSignMessage
+
+The `useSignMessage` hook enables signing a message on behalf of the user's Embedded Account.
```
moldy530, line 23: we should call out here that this will use 6492 if the account is not deployed yet

site/react/useSignTypedData.md
```diff
@@ -0,0 +1,41 @@
...
+    - name: twitter:title
+      content: useSignTypedData
+  - - meta
+    - name: twitter:description
+      content: An overview of the useSignTypedData hook
+---
+
+# useSignTypedData
+
+The `useSignTypedData` hook enables signing typed data on behalf of the user's Embedded Account.
```
moldy530, line 23: we should call out here that this uses 6492 if the account is not deployed yet

site/snippets/react/useSignMessage.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { useState } from "react";
+
+export function ComponentWithAddPasskey() {
+  // Assumes the app has context of a signer with an authenticated user
+  const [message, setMessage] = useState("");
+  const client = useSmartAccountClient({
+    type: "MultiOwnerModularAccount", // alternatively pass in "LightAccount",
+    accountParams: {}, // optional param for overriding any account specific properties
+  });
+  const { signMessage, isSigningMessage } = useSignMessage(client);
```
moldy530, line 14:
```suggestion
  const { signMessage, isSigningMessage } = useSignMessage({ client });
```

site/snippets/react/useSignTypedData.tsx
```diff
@@ -0,0 +1,49 @@
...
+  useSmartAccountClient,
+} from "@alchemy/aa-alchemy/react";
+import { useState } from "react";
+
+export function ComponentWithAddPasskey() {
+  // Assumes the app has context of a signer with an authenticated user
+  const [typedDataMessage, setTypedDataMessage] = useState("");
+  const client = useSmartAccountClient({
+    type: "MultiOwnerModularAccount", // alternatively pass in "LightAccount",
+    accountParams: {}, // optional param for overriding any account specific properties
```
moldy530, line 12:
```suggestion

```

site/snippets/react/useSignTypedData.tsx
```diff
@@ -0,0 +1,49 @@
...
+  });
+  const { signTypedData, isSigningTypedData } = useSignTypedData(client);
```
moldy530, line 14:
```suggestion
  const { signTypedData, isSigningTypedData } = useSignTypedData({ client });
```

site/react/useSignTypedData.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Import
+
+```ts
+import { useSignTypedData } from "@alchemy/aa-alchemy/react";
+```
+
+## Usage
```
moldy530, line 31: we need a `## Params` section as well

site/react/useSignMessage.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Import
+
+```ts
+import { useSignMessage } from "@alchemy/aa-alchemy/react";
+```
+
+## Usage
```
moldy530, line 31: we need a `## Params` section as well

site/react/useSignMessage.md
```diff
@@ -0,0 +1,41 @@
...
+
+<<< @/snippets/react/useSignMessage.tsx
+
+## Return Type
+
+```ts
+import { type useSignMessageResult } from "@alchemy/aa-alchemy/react";
```
moldy530, line 38:
```suggestion
import { type UseSignMessageResult } from "@alchemy/aa-alchemy/react";
```

site/react/useSignMessage.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+Returns a `Hex` representation of the signed message.
```
moldy530, line 41: Same comment as the other prs here, we need to enumerate what makes up `UseSignMessageResult`

site/react/useSignTypedData.md
```diff
@@ -0,0 +1,41 @@
...
+
+<<< @/snippets/react/useSignTypedData.tsx
+
+## Return Type
+
+```ts
+import { type useSignTypedDataResult } from "@alchemy/aa-alchemy/react";
```
moldy530, line 38:
```suggestion
import { type UseSignTypedDataResult } from "@alchemy/aa-alchemy/react";
```

site/react/useSignTypedData.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+Returns a `Hex` representation of the signed typed data.
```
moldy530, line 41: see above

site/snippets/react/useSignTypedData.tsx
```diff
@@ -0,0 +1,48 @@
+import {
+  useSignTypedData,
+  useSmartAccountClient,
+} from "@alchemy/aa-alchemy/react";
+import { useState } from "react";
+
+export function ComponentWithSignTypedData() {
+  // Assumes the app has context of a signer with an authenticated user
+  const [typedDataMessage, setTypedDataMessage] = useState("");
+  const client = useSmartAccountClient({
```
rthomare, line 10: this is an example where it would be nice to show the onsuccess or onerror. People would need to use that hook pretty frequently here.

packages/alchemy/src/react/errors.ts
```diff
@@ -5,3 +5,9 @@ export class NoAlchemyAccountContextError extends BaseError {
     super(`${hookName} must be used within a AlchemyAccountProvider`);
   }
 }
+
+export class ClientUndefinedError extends BaseError {
```
rthomare, line 9: nit would name this `ClientUndefinedHookError`

packages/alchemy/src/react/hooks/useSignTypedData.ts
```diff
@@ -0,0 +1,70 @@
...
+    queryClient
+  );
+
+  return {
+    signTypedData,
+    signTypedDataAsync,
+    signedTypedData,
+    isSigningTypedData,
+    error,
+  };
```
rthomare, line 69: Quick thoughts: 
1. The `signedTypeData` and `error` imply that we hold some type of state. But we also provide the async version of the function. 
   1. This would be confusing for a downstream consumer of this hook. 
2. I know it seems inconvenient, but we should make it so that people explicitly use the onError or onSuccess callback. So it is more obvious how that would work then, with the signedTypeData and error results.
   1. We expose a separate hook called `useAsyncSignTypedData` that expects the user to use the async callback and handle the success and failures themselves.

What are your thoughts? That might also simplify the docs a bit. But realize this would be more code. Side note this has been a problem I've had with react-query for a long time 😬

packages/alchemy/src/react/hooks/useSignMessage.ts
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        return client.signMessageWith6492(params);
+      },
+      ...mutationArgs,
+    },
+    queryClient
+  );
+
+  return {
```
rthomare, line 64: same comment as line 69 in `packages/alchemy/src/react/hooks/useSignTypedData.ts`

site/snippets/react/useSignMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+} from "@alchemy/aa-alchemy/react";
+import { useState } from "react";
+
+export function ComponentWithSignMessage() {
+  // Assumes the app has context of a signer with an authenticated user
+  const [message, setMessage] = useState("");
+  const client = useSmartAccountClient({
+    type: "MultiOwnerModularAccount",
+  });
+  const { signMessage, isSigningMessage } = useSignMessage({ client });
```
rthomare, line 13: same comment as line 10 in site/snippets/react/useSignTypedData.tsx

site/snippets/react/useSignMessage.tsx
```diff
@@ -0,0 +1,30 @@
+import {
+  useSignMessage,
+  useSmartAccountClient,
+} from "@alchemy/aa-alchemy/react";
+import { useState } from "react";
+
+export function ComponentWithSignMessage() {
+  // Assumes the app has context of a signer with an authenticated user
```
rthomare, line 8: would be good to call out you need ConfigContext

site/snippets/react/useSignTypedData.tsx
```diff
@@ -0,0 +1,48 @@
+import {
+  useSignTypedData,
+  useSmartAccountClient,
+} from "@alchemy/aa-alchemy/react";
+import { useState } from "react";
+
+export function ComponentWithSignTypedData() {
+  // Assumes the app has context of a signer with an authenticated user
```
rthomare, line 8: would be good to call out you need ConfigContext

packages/alchemy/src/react/hooks/useSignMessage.ts
```diff
@@ -0,0 +1,71 @@
+"use client";
```
rthomare, line 1: nit jsdocs

packages/alchemy/src/react/hooks/useSignTypedData.ts
```diff
@@ -0,0 +1,70 @@
+"use client";
```
rthomare, line 1: nit jsdocs

site/snippets/react/useSignTypedData.tsx
avasisht23: could you clarify? you mean the method on this hook to add an onSuccess / onError method?

packages/alchemy/src/react/hooks/useSignMessage.ts
denniswon: why do we need "use client" here? it's the responsibility of app developers to set these if they are using nextjs. this line is irrelevant for apps that do not use nextjs?

packages/alchemy/src/react/hooks/useSignMessage.ts
```diff
@@ -0,0 +1,71 @@
...
+    isPending: isSigningMessage,
+    error,
+  } = useMutation(
+    {
+      mutationFn: async (params: SignMessageArgs) => {
+        if (!client) {
+          throw new ClientUndefinedError("useSignMessage");
+        }
+
+        return client.signMessageWith6492(params);
```
denniswon, line 57: what about just signMessage? are we intentionally not surfacing non 6492 message signing method via our hooks?

moldy530: This is following what other libs do as well, like wagmi

packages/alchemy/src/react/hooks/useSignMessage.ts
avasisht23: Yeah it's irrelevant out of nextjs, but future proofs for nextjs client side rendering

See moldys pr here https://github.com/alchemyplatform/aa-sdk/pull/539#discussion_r1545463146

packages/alchemy/src/react/hooks/useSignTypedData.ts
moldy530: Doesn't the async method in react-query still plug in into the mutation lifecycle? So on success and on error still fire, as well as the loading states and error states? I kinda like the convenience of having the async method exported with the hooks, we should probs add it to some of the other hooks too actually because people might want to chain certain calls in their business logic, which is only doable with the async method. At which point having async versions of each hook becomes kind of cumbersome to maintain. 

Fwiw this is how wagmi handles it as well

avasisht23: based on the useMutation docs, async does plug into the lifecycle: https://tanstack.com/query/latest/docs/framework/react/reference/useMutation#usemutation

I think it's a 2 way door to introduce a separate async hook later. agree it's cumbersome. esp. if wagmi handles things this way too, let's keep async mutate on the same hook as mutate.

rthomare: > Doesn't the async method in react-query still plug in into the mutation lifecycle? So on success and on error still fire, as well as the loading states and error states? I kinda like the convenience of having the async method exported with the hooks, we should probs add it to some of the other hooks too actually because people might want to chain certain calls in their business logic, which is only doable with the async method. At which point having async versions of each hook becomes kind of cumbersome to maintain.

It does. I'm not super opinionated about this, but this is a personal preference ergonomic issue that I think we'll only find out more about down the road. I agree on the less code we maintain the better. So let's punt on my idea, and keep it to one hook for now and then we can break it out later. Just might be a point of confusion later down the road.

packages/alchemy/src/react/hooks/useSignMessage.ts
avasisht23: resolved to punt for now, per discussion on that other comment

rthomare (dismissed):

rthomare (approved):

PR #580, opened 2024-04-16
feat: paymaster middleware update per entrypoint v7 user operation

denniswon:
# Pull Request Checklist

---

## PR-Codex overview
This PR updates package versions and adds support for `EntryPointVersion` in various files.

### Detailed summary
- Updated package versions to `3.8.2-alpha.0`
- Added support for `EntryPointVersion` in multiple files
- Defined new schemas for `UserOperationFeeOptions`

> The following files were skipped due to too many changes: `packages/core/src/middleware/defaults/paymasterAndData.ts`, `packages/core/e2e-tests/simple-account.e2e.test.ts`, `site/CHANGELOG.md`, `packages/alchemy/CHANGELOG.md`, `packages/core/src/middleware/defaults/gasEstimator.ts`, `packages/accounts/CHANGELOG.md`, `CHANGELOG.md`, `packages/core/CHANGELOG.md`, `packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts`, `packages/core/src/types.ts`, `site/packages/aa-core/smart-account-client/types/userOperationOverrides.md`, `packages/alchemy/src/middleware/gasManager.ts`


dphilipson (dismissed):

packages/alchemy/src/middleware/gasManager.ts
```diff
@@ -190,67 +185,146 @@ export function alchemyGasManagerMiddleware<C extends ClientWithAlchemyMethods>(
...
-    paymasterAndData: async (struct, { overrides, feeOptions, account }) => {
+    paymasterAndData: async (
+      struct,
+      { overrides: overrides_, feeOptions, account }
+    ) => {
+      // handle one-off overrides for bypassing paymaster middleware
+      if (
+        account.getEntryPoint().version === "0.6.0" &&
+        (overrides_ as UserOperationOverrides<"0.6.0">)?.paymasterAndData ===
+          "0x"
```
avasisht23, line 238: should be use `EmptyHex` here?

packages/alchemy/src/middleware/gasManager.ts
```diff
@@ -190,67 +185,146 @@ export function alchemyGasManagerMiddleware<C extends ClientWithAlchemyMethods>(
...
+      ) {
+        return {
+          ...struct,
+          paymasterAndData: "0x",
```
avasisht23, line 242: same here, `EmptyHex`?

packages/alchemy/src/middleware/gasManager.ts
```diff
@@ -190,67 +185,146 @@ export function alchemyGasManagerMiddleware<C extends ClientWithAlchemyMethods>(
...
+        };
+      } else if (
+        account.getEntryPoint().version === "0.7.0" &&
+        (overrides_ as UserOperationOverrides<"0.7.0">)?.paymasterData === "0x"
```
avasisht23, line 246: `EmptyHex`?

avasisht23 (commented):
high level comments and this overall looks good, but would love comments from @dphilipson too!

packages/alchemy/src/middleware/gasManager.ts
dphilipson: That's not a thing at the moment. `EmptyHex` is a type, not a value.

If you're suggesting creating a constant for it, e.g.
```ts
export const EMPTY_HEX: EmptyHex = "0x";
```
I dunno if I think it improves readability really, but I wouldn't be opposed if other people like it.

packages/alchemy/src/middleware/gasManager.ts
dphilipson, line 246: Technically, "0x" can be valid data for a paymaster. There could conceivably exist a paymaster that doesn't use this field- maybe it makes its decision of whether to valid purely based on the `msg.sender` alone. If so, we wouldn't want to treat this case as there being no paymaster set.

It might make more sense to look at the `paymaster` field, which would contain the paymaster address, to make this determination.

packages/core/src/client/schema.ts
```diff
@@ -57,6 +57,16 @@ export const UserOperationFeeOptionsSchema = z
...
 
+export const UserOperationFeeOptionsSchema_v7 =
+  UserOperationFeeOptionsSchema_v6.extend({
+    paymasterVerificationGasLimit: UserOperationFeeOptionsFieldSchema,
+    paymasterPostOpGasLimit: UserOperationFeeOptionsFieldSchema,
+  })
+    .partial()
+    .strict();
+
+export const UserOperationFeeOptionsSchema = UserOperationFeeOptionsSchema_v7;
```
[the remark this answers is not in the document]
denniswon: I did not set to union because it's unncessary has UserOperationFeeOptionsSchema_v7 extends UserOperationFeeOptionsSchema_v6

packages/core/src/middleware/defaults/paymasterAndData.ts
```diff
@@ -1,13 +1,19 @@
...
   const entryPoint = account.getEntryPoint();
   if (entryPoint.version === "0.6.0") {
-    (struct as UserOperationStruct<"0.6.0">).paymasterAndData = "0x";
+    (struct as UserOperationStruct<"0.6.0">).paymasterAndData =
+      (overrides as UserOperationOverrides<"0.6.0">)?.paymasterAndData ?? "0x";
+  } else {
+    // Remove paymasterVerificationGasLimit field from uo struct
+    // filled during the `eth_estimateUserOperationGas` call from the default gasEstimator
+    delete (struct as UserOperationStruct<"0.7.0">)
+      .paymasterVerificationGasLimit;
```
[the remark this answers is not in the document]
denniswon: this field is computed from gasEstimator.

packages/alchemy/src/middleware/gasManager.ts
denniswon: EmptyHex is type not a constant.

packages/alchemy/src/middleware/gasManager.ts
denniswon: ditto

packages/alchemy/src/middleware/gasManager.ts
denniswon: paymaster middleware setting 0x as the valid paymaster data is fine. but setting overrides' paymasterAndData or paymasterData field to '0x' is an explicit indication of bypassing the paymaster data api call to the paymaster/gas manager.

dphilipson: I actually don't quite understand what this function is doing, I need to read more, but it feels wrong that one particular valid value of `paymasterData` would cause this middleware to have different behavior than any other valid value of `paymasterData`. That seems like it could cause this to have unexpectedly different behavior for certain paymasters.

denniswon: Also, this PR changes the previous behavior where 

1) override value paymasterData / paymasterAndData only allows '0x' specifically, per discussion with @moldy530 from previous PRs. In fact, we already check against '0x' for overrides.paymasterAndData in `gasManager`.
2) bypassing paymaster is handled by the paymaster middleware used not by the smart account client. We used to not even run paymaster middleware when '0x' is supplied to the overrides, but instead, I think it should be left to the paymaster middleware how to handle overrides value:

The default paymater/alchemy gas manager will bypass paymaster if '0x' paymasterData field is used as an override, but if the dev uses their own paymaster middle by overriding with their own, it is up to them on how to handle the paymaster override values.

cc: @moldy530 @avasisht23 @dphilipson I think this approach is more robust and easier to understand for devs in terms of devEx. For consistency of middleware and overrideMiddleware logic for all middleswares, don't think have a special logic for a certain middleware is desired. 

This change in the PR is open for discussion. If you guys believe this requires more review, happy to separate out into a separate PR.

packages/alchemy/src/middleware/gasManager.ts
avasisht23: ah gotcha, misread the file. thanks!

packages/alchemy/src/middleware/gasManager.ts
dphilipson: It's different for `paymasterAndData` because "0x" there means you're not using a paymaster; otherwise, `paymasterAndData` starts with the paymaster address. Thus, "0x" is not a valid valid of `paymasterAndData` if you're not using a paymaster, so it makes sense to use as a marker for the middleware.

By contrast, "0x" is a valid value for `paymasterData` even if you are using a paymaster. That's why I suggested having the trigger for this behavior be that the `paymaster` field is absent, which more directly mirrors the 0.6 behavior.

dphilipson: Oh wait, maybe I see. This is for the gas manager so it wouldn't make sense to be overriding `paymasterData` anyways. So it doesn't make sense to have any value as the override, so we've chosen one particular value as the marker to change the behavior. Is that right?

dphilipson: Let me see if I understand this correctly:

If someone applies the gas manager middleware, then by default for any user operation sent it will request paymaster gas and data from the gas manager and fill it in to the operation.

We want an escape hatch so someone who has this middleware applied can still choose to send user operations that do not use the gas manager. To do so, in 0.6 we allow them to pass an override of `paymasterAndData: "0x"`. If they do, the operation will use no paymaster.

What does a 0.6 user do if they have the gas manage middleware and they want to use a different paymaster?

denniswon: yes, @dphilipson this is specifically for alchemy gas manager. The default paymaster is no op, and if the devs use their own paymaster, they take care of how they would handle the override value.

same behavior for all other middleware.

For alchemy gas manager, they specify gas policy id, and paymaster, paymasterdata will be set by the gas manager. If gas manager does for some reason return 0x for paymaster data, then yes we might need to update the logic here where we don't regard 0x paymaster bypassing, but when that becomes the case in the future, we can change our own gas manager paymaster.

cc: @dphilipson @avasisht23 @moldy530

dphilipson: Are you saying that if the gas manager middleware is applied, then a 0.6 user can opt to have **no** paymaster if they pass "0x" as `paymasterAndData`, but there's no way for them to opt to have a **different** paymaster?

denniswon: nope.

denniswon: using 2 paymasters at the same time isn't allowed this way. we discussed this (cc: with @moldy530 ) on this, and decided that it is an edge case that we or customers are not sure if that's needed. and if they really want 2 paymasters at the same time, the correct/suggested way is two use two separate clients.

dphilipson: Okay, I guess this code is fine then in that it's similar to the 0.6 behavior, but we should revisit the 0.6 behavior later. It's unintuitive and undiscoverable that you can override to no paymaster but not to a different paymaster.

denniswon: if you want to override paymaster, need to override at the client level, using paymaster middleware override. One off override is only for bypassing paymaster for client configured with a paymaster.

avasisht23: we should prob make "0x" a constant like `NO_ALCHEMY_GASMASTER_OVERRIDE` or something, as it's a magic number in this situation.

yeah, this middleware is only for using the alchemy gas manager. And agree, 2 gas managers at the same time on the same alchemy smart account client isn't possible.

could you clarify one more time if there's any way for this middleware method to be used with a non-Alchemy paymaster? otherwise, this feels fine to me. 

lmk what you think @dphilipson

dphilipson: I don't understand why I wouldn't be able to pass an override of `paymasterAndData` of some other paymaster's data and have it use that paymaster. That's the behavior I would expect.

Then "0x" wouldn't be a special case (in 0.6). It would follow the same logic of any other override for `paymasterAndData` in that it would replace the `paymasterAndData` field in the op and then skip the gas manager middleware.

dphilipson (dismissed):

PR #581, opened 2024-04-16
feat: add useSendUserOperation and useDropAndReplaceUserOperation hooks

avasisht23:
# Pull Request Checklist

---

## PR-Codex overview
This PR adds new hooks `useSendUserOperation` and `useDropAndReplaceUserOperation` for interacting with Alchemy smart accounts in React components.

### Detailed summary
- Updated error message in `NoAlchemyAccountContextError` class
- Added `useSendUserOperation` and `useDropAndReplaceUserOperation` hooks
- Created corresponding components for each hook
- Updated sidebar links
- Updated exports in `alchemy/src/react/index.ts`
- Added meta information for new hooks in documentation pages

> The following files were skipped due to too many changes: `packages/alchemy/src/react/hooks/useSendUserOperation.ts`


packages/alchemy/src/react/hooks/useSendUserOperation.ts
```diff
@@ -0,0 +1,97 @@
...
+  type UseMutateAsyncFunction,
+  type UseMutateFunction,
+} from "@tanstack/react-query";
+import type { SupportedAccounts } from "../../config/types.js";
+import { useAlchemyAccountContext } from "../context.js";
+import { ClientUndefinedError } from "../errors.js";
+import type { BaseHookMutationArgs } from "../types.js";
+import { type UseSmartAccountClientResult } from "./useSmartAccountClient.js";
+
+export type UseSendUserOperationMutationArgs<
```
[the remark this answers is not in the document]
avasisht23: oh yeah i had that at first, but took it out in favor of separating concerns. but down to bring it back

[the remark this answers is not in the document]
avasisht23: oh yeah the if i do this then this hooks result becomes a union type of `Hash | SendUOResult`, where `SendUOResult` is an object with the uohash and the uo request. little more special to deal with devex wise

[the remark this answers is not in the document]
avasisht23: ill just still return the uo request in the object even if the hash is the txn hash, problem solved

packages/alchemy/src/react/hooks/useDropAndReplaceUserOperation.ts
```diff
@@ -0,0 +1,86 @@
...
+    isPending: isDroppingAndReplacingUserOperation,
+    error,
+  } = useMutation(
+    {
+      mutationFn: async (
+        params: DropAndReplaceUserOperationParameters<TAccount>
+      ) => {
+        if (!client) {
+          throw new ClientUndefinedError("useDropAndReplaceUserOperation");
+        }
```
denniswon, line 71: ditto comment made in another hook pr:

or all out hooks, we should check from AlchemyAccountContext which state the user/account is in, and throw informative error messages instead of just throwing null exceptions or other generic error messages

packages/alchemy/src/react/hooks/useSendUserOperation.ts
```diff
@@ -0,0 +1,108 @@
...
+    unknown
+  >;
+  sendUserOperationResult:
+    | SendUserOperationResult<TEntryPointVersion>
+    | undefined;
+  isSendingUserOperation: boolean;
+  error: Error | null;
+};
+
+export function useSendUserOperation<
```
denniswon, line 58: this is only meant to be used for smart accounts? we should check if account is connected to the client, not just that client exists. in general, throw more informative state error back to the caller.

Makes sense to group signer/smart account/modular account hooks separately in the subdirs of hooks for different accounts, and also have more common state checking/error messages/code shared.

packages/alchemy/src/react/hooks/useDropAndReplaceUserOperation.ts
moldy530: See my comment in the use client actions hook about this. I don't think this is a generic error. There's nothing to check in the account context that will make it more informative. The client here is a parameter to the hook and can be null. If it's null we error. Beyond that what else can we pull from context? What if the user is passing in a client they didn't create via context and it's still null?

packages/alchemy/src/react/hooks/useSendUserOperation.ts
moldy530: The client will throw if it is not connected to an account already and the typing on this specific client parameter forbids it from not not being connected anyways since it has to be a LA client or MA client

packages/alchemy/src/react/hooks/useDropAndReplaceUserOperation.ts
denniswon: you are right. I left a similar comment on signer hook prs, and left similar comments here, but I realized that the core SDK takes care of these already.

rthomare (dismissed):

rthomare (approved):

PR #603, opened 2024-04-21
feat: estimate user operation gas smart account client action and how to guide

denniswon:
# Pull Request Checklist

---

## PR-Codex overview
This PR adds functionality for estimating user operation gas in the Smart Account Client. 

### Detailed summary
- Added `estimateUserOperationGas` action
- Updated `buildUserOperation` function
- Added metadata to `estimateUserOperationGas.md`
- Added `_initUserOperation` internal function

> The following files were skipped due to too many changes: `packages/core/src/actions/smartAccount/internal/initUserOperation.ts`, `site/using-smart-accounts/estimate-gas/estimate-user-op-gas.md`, `site/using-smart-accounts/estimate-gas/estimate-user-op-gas-state-overrides.md`, `site/using-smart-accounts/entrypoint-v7.md`


packages/core/src/actions/smartAccount/estimateUserOperationGas.ts
```diff
@@ -0,0 +1,94 @@
...
+          initCode: account.getInitCode(),
+          sender: account.address,
+          nonce,
+          callData,
+          signature,
+        } as Deferrable<UserOperationStruct<TEntryPointVersion>>)
+      : ({
+          factory: conditionalReturn(
+            account.isAccountDeployed().then((deployed) => !deployed),
+            account.getFactoryAddress()
```
dphilipson, line 75: This looks slightly inefficient in that it will make the call to `getFactoryAddress` even in the case where it's already deployed, likewise with `getFactoryData` just below. It may be worth changing the signature here so the second argument to `conditionalReturn` is `() => account.getFactoryAddress()`.

site/using-smart-accounts/estimate-gas/estimate-user-op-gas-state-overrides.md
```diff
@@ -0,0 +1,121 @@
...
+
+`SmartAccountClient` of `aa-sdk` provides default [`gasEstimator`](/packages/aa-core/smart-account-client/middleware/index#gasEstimator) that internally includes calling `eth_estimateUserOperationGas` to the bundler in addition to applying user operation overrides or fee options for populating the user operation gas fields in a most desired mannder. If you are looking to estimate gas of a user operation without building the entire user operation through the middleware pipeline, you can call the [`estimateUserOperationGas`](/packages/aa-core/smart-account-client/actions/estimateUserOperationGas) action on the `SmartAccountClient` to directly fetch network user operation gas estimates from the bundler. The action returns `UserOperationEstimateGasResponse` containing the estimated gas values.
```
dphilipson, line 63: (Nit, typos)

* "provides default `gasEstimator`" -> "provides a default `gasEstimator`"
* "mannder" -> "manner"

likewise where this paragraph is repeated below.

site/using-smart-accounts/estimate-gas/estimate-user-op-gas-state-overrides.md
```diff
@@ -0,0 +1,121 @@
...
+
+::: tip Note 1
+The actual gas estimation or fee estimation is performed by the bundler connected to the `SmartAccountClient`, so depending on the bundler you are using, the gas estimate value might be different.
```
dphilipson, line 66: (Minor) I find the references to fee estimation in this paragraph and the one above a little bit confusing, since `estimateUserOperationGas` does not provide fee estimation. I might take them out, or if you want to keep them around then clarify somewhere that `estimateUserOperationGas` estimates gas quantities but not fees.

site/using-smart-accounts/estimate-gas/estimate-user-op-gas-state-overrides.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+::: tip Storage Slot Calculation
+Calculating the storage location depends on the structure of the contract’s storage and the specific variable you want to override. For the DAI contract example, the storage location for the wards mapping for your address is calculated based on the [Solidity storage layout rules](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays):
+
+```ts
+const yourAddress = "0x...";
+const overrideStorageLocation = keccak256(
+  concatenate(hexZeroPad(yourAddress), hexZeroPad(0x0))
+); // 0xSTORAGE_SLOT_OF_YOUR_ADDRESS
```
dphilipson, line 54: After computing `overrideStorageLocation`, we might want to provide further examples of how to use that to pass state overrides to `estimateUserOperationGas`. This page never actually explains what parameter to pass to use the state override.

It's probably enough to have an example, and link to Geth's docs if people want more information about the format of the state overrides.

site/using-smart-accounts/estimate-gas/estimate-user-op-gas.md
```diff
@@ -0,0 +1,87 @@
...
+
+::: tip Note 1
+The actual gas estimation or fee estimation is performed by the bundler connected to the `SmartAccountClient`, so depending on the bundler you are using, the gas estimate value might be different.
+:::
+
+::: tip Note 2
+Note that the `estimateUserOperationGas` action returns the bare result of gas estimates returned directly from the connected bundler without applying user operation gas overrides or client fee options that are applied from the default `gasEstimator` used when constructing the actual user operation request to send to the network.
+:::
```
dphilipson, line 37: I think we might want to give this page a reference to the one about state overrides. As written, a reader could likely end up reading this first and have no idea that state overrides are possible.

Another possible organization, since as written there's a lot of duplicated text and examples between this and the state override page, would be to have this page be the one that describes this endpoint and options mentioning that state overrides are possible, and then link to the other page which can focus entirely on the format of state overrides without having to reexplain the main `estimateUserOperationGas` api.

dphilipson (commented):
Looks great! A few suggestions inline.

packages/core/src/actions/smartAccount/estimateUserOperationGas.ts
denniswon: this is deferrable though, a promise right? so until at the end, these promises will not be actually executed/awaited?

site/using-smart-accounts/estimate-gas/estimate-user-op-gas.md
denniswon: this is a how to guide. and for estimate gas section, there are two docs: one with and one without state overrides.
The packages doc gives param/return types/reference to state override from geth, etc. And this doc has embedded link to that package doc page for `estimateUserOperationGas`.

packages/core/src/actions/smartAccount/estimateUserOperationGas.ts
moldy530: +1 to dphils comment the inputs to conditional return should be lambdas or at least the second parameter should be. They're promises, but they're instantiated immediately and will start running. calling `await` at somepoint just polls the completed state of the promise before unwrapping it

denniswon: makes sense. will udpate.

site/migration-guides/entrypoint-v7.md
```diff
@@ -0,0 +1,231 @@
...
+
+Although some bundlers, such as Alchemy [`rundler`](https://github.com/alchemyplatform/rundler), does support the `stateOverride` feature for user operation gas estimation for both EntryPoint v0.6 and v0.7 user operations, until Entrypoint v0.6, the [`eth_estimateUserOperationGas`](https://eips.ethereum.org/EIPS/eip-4337#-eth_estimateuseroperationgas) RPC method did not include the optional `stateOverride` parameter per specification. From EntryPoint v0.7, however, `stateOverride` has been added to officially added as an optional parameter to the gas estimation method, thanks to the removal of network requirement to support state override feature.
```
dphilipson, line 192: Typo: "does support the `stateOverride`" -> "do support the `stateOverride`"

Also,
> thanks to the removal of network requirement to support state override feature
is incorrect. `delegateAndRevert` is not useful for implementing the `stateOverride` field of bundler methods. It's intended use is internal to bundlers, and it does not allow general state overrides that would be necessary to support the `stateOverride` parameter.

dphilipson (commented):
Left a note about `delegateAndRevert`.

site/using-smart-accounts/estimate-gas/estimate-user-op-gas-state-overrides.md
denniswon: good catch. I forgot to add the example. Added!

site/migration-guides/entrypoint-v7.md
denniswon: @dphilipson this note is directly from official 0.7 release note:
Screenshot 2024-04-23 at 3 01 06 PM

denniswon: Updated as "The state overrides during transaction calls (e.g., [`eth_call`](https://docs.alchemy.com/reference/eth-call) and [`eth_estimateGas`](https://docs.alchemy.com/reference/eth-estimategas)) allow for more flexible testing and gas estimation. However, not all networks support this feature. For networks lacking state overrides, `EntryPoint` v0.7 introduced the `delegateAndRevert()` helper in the `EntryPoint` contract to accommodate nodes without state overrides in `eth_call` and `eth_estimateGas`. It aids in validating and estimating gas for transactions by simulating conditions without permanently changing the on-chain state."

packages/core/src/actions/smartAccount/estimateUserOperationGas.ts
```diff
@@ -0,0 +1,94 @@
...
+    ? account.encodeBatchExecute(uo)
+    : typeof uo === "string"
+    ? uo
+    : account.encodeExecute(uo);
+
+  const signature = account.getDummySignature();
+
+  const nonce = account.getNonce(overrides?.nonceKey);
+
+  const struct =
```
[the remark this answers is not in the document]
denniswon: added `initUserOperation` internal function to be shared.

denniswon:
@moldy530 @dphilipson final look/stamp please. thanks!

packages/core/src/actions/smartAccount/estimateUserOperationGas.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { UserOperationEstimateGasResponse } from "../../types.js";
+import { deepHexlify, resolveProperties } from "../../utils/index.js";
+import { _initUserOperation } from "./internal/initUserOperation.js";
+import type {
+  SendUserOperationParameters,
+  UserOperationContext,
+} from "./types.js";
+
+export async function estimateUserOperationGas<
+  TTransport extends Transport = Transport,
```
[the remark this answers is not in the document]
moldy530: can you address these?

packages/core/src/actions/smartAccount/internal/initUserOperation.ts
```diff
@@ -0,0 +1,76 @@
...
+import { ChainNotFoundError } from "../../../errors/client.js";
+import type { UserOperationStruct } from "../../../types.js";
+import { conditionalReturn, type Deferrable } from "../../../utils/index.js";
+import type {
+  SendUserOperationParameters,
+  UserOperationContext,
+} from "../types.js";
+
+export async function _initUserOperation<
+  TTransport extends Transport = Transport,
```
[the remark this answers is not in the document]
moldy530: can you address these?

packages/core/src/actions/smartAccount/estimateUserOperationGas.ts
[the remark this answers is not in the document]
denniswon: done

site/migration-guides/entrypoint-v7.md
dphilipson: Yeah, the point that was incorrect in the original wording was that it implied that `delegateAndRevert` supported the **state override parameter** of `eth_estimateUserOperationGas`. The wording in 4337's note doesn't say this: it says that `delegateAndRevert` is used to support the implementation of validation and gas estimation, which is true.

dphilipson: Actually, I would say the current wording still has the incorrect implication that `delegateAndRevert` allows the state overrides parameter on networks that don't support it (really: it's less network-support and more node client-support). It's in the section titled "`stateOverride` field added to `UserOperationOverrides` for validating and estimating gas for transactions".

I might just take out mention of `delegateAndRevert` in this section. It's not related to the `stateOverride` parameter.

dphilipson:
I added a few more comments about `delegateAndRevert` in the thread above. Other than that, everything here looks good to me.

site/migration-guides/entrypoint-v7.md
denniswon: ok

denniswon:
```
// https://etherscan.io/address/0x0000000071727de22e5e9d8baf0edac6f37da032
export const EntryPointAbi_v7 = [
  {
    inputs: [
      { internalType: "bool", name: "success", type: "bool" },
      { internalType: "bytes", name: "ret", type: "bytes" },
    ],
    name: "DelegateAndRevert",
    type: "error",
  },
  {
    inputs: [
      { internalType: "uint256", name: "opIndex", type: "uint256" },
      { internalType: "string", name: "reason", type: "string" },
    ],
    name: "FailedOp",
    type: "error",
  },
  ```
  
  Entrypoint v7 added this to entrypoint contract, not in entrypoint v6. then, what is this update referring to?
  
  @dphilipson

File: site/migration-guides/entrypoint-v7.md (hunk 6 above)
Comment (dphilipson) [reply site/migration-guides/entrypoint-v7.md]: Good question. Here's the full explaination as I'm thinking of it.

One other change in the entry point from 0.6 to 0.7 is that the 0.6 entry point used to contain some functions which were intended to be called by bundlers purely for simulation's sake: `simulateValidation` and `simulateHandleOp`.

In 0.7, they decided that there was no need to deploy functions that would only ever be used off-chain, so instead they defined a new contract, `EntryPointSimulations`, which extends `EntryPoint` with the simulation functions. `EntryPointSimulations` is not deployed on-chain; instead, bundlers are now expected to use state overrides in their internal calls to `eth_call`/`eth_estimateGas`. The bundler will still call `simulateValidation` or `simulateHandleOp` on the entry point address as before, but since the real entry point doesn't have these methods anymore, the bundler needs to spoof the state of the blockchain to replace the entry point code with the version that has these methods.

Bundlers can do this using state overrides, but the 4337 authors didn't want to _assume_ that state overrides are available. This may be because state overrides are not part of the standard eth api, but rather an extension added by Geth (and copied by other node implementations that want to be compatible). Notice how state overrides don't appear in the [official eth jsonrpc spec for `eth_call`](https://ethereum.github.io/execution-apis/api-documentation/).

So the entry point provides an escape hatch for node clients that don't support state overrides, `delegateAndRevert`, which effectively allows the caller to do the same thing by "overriding" the entry point's code with the entry point simulation code. `delegateAndRevert` uses delegatecall internally, and in general a delegatecall can be thought of as replacing a contract's code with a different contract for the duration of a single call.

`delegateAndRevert` is a lot less powerful than full state overrides. `delegateAndRevert` only allows the caller to replace the code of the entry point contract, while state overrides allow the caller to replace the code, storage, balance, or nonce of any address. But for the purpose of calling the `EntryPointSimulation` functions, just replacing the entry point code is enough.

Note that this is all totally separate from the optional state overrides parameter to `eth_estimateUserOperationGas`. That parameter is giving the caller the full set of override options as you would get when using overrides in an `eth_call`, which as stated above is much more powerful than what `delegateAndRevert` allows.

**Comment by denniswon** [comment]
@dphilipson updated per your suggestion.

```
#### `stateOverride` field added to `UserOperationOverrides` for validating and estimating gas for transactions

The state overrides during transaction calls (e.g., [`eth_call`](https://docs.alchemy.com/reference/eth-call) and [`eth_estimateGas`](https://docs.alchemy.com/reference/eth-estimategas)) allow for more flexible testing and gas estimation. Specifically, `EntryPoint` v0.7 introduced the `delegateAndRevert()` helper in the `EntryPoint` contract to accommodate nodes without state overrides in `eth_call` and `eth_estimateGas` for validating and estimating gas for transactions in simulated conditions without permanently changing the on-chain state.

Although some bundlers, such as Alchemy [`rundler`](https://github.com/alchemyplatform/rundler), do support the `stateOverride` feature for user operation gas estimation for both EntryPoint v0.6 and v0.7 user operations, until Entrypoint v0.6, the [`eth_estimateUserOperationGas`](https://eips.ethereum.org/EIPS/eip-4337#-eth_estimateuseroperationgas) RPC method did not include the optional `stateOverride` parameter per specification. From EntryPoint v0.7, `stateOverride` has been officially added as an optional parameter to the gas estimation method.

To take advantage of such update, `aa-sdk` supports `stateOverride` value as part of the [`UserOperationOverrides`](/resources/types#useroperationoverrides) to be used during the gas estimation of simualted user operation transactions.
```

File: site/using-smart-accounts/entrypoint-v7.md [hunk 10]
```diff
@@ -181,11 +181,11 @@ Note that per [`ERC-4337`](https://eips.ethereum.org/EIPS/eip-4337#parameters),
...
+Although some bundlers, such as Alchemy [`rundler`](https://github.com/alchemyplatform/rundler), do support the `stateOverride` feature for user operation gas estimation for both EntryPoint v0.6 and v0.7 user operations, until Entrypoint v0.6, the [`eth_estimateUserOperationGas`](https://eips.ethereum.org/EIPS/eip-4337#-eth_estimateuseroperationgas) RPC method did not include the optional `stateOverride` parameter per specification. From EntryPoint v0.7, `stateOverride` has been officially added as an optional parameter to the gas estimation method.
 
-With such update, `aa-sdk` accepts `stateOverride` value as part of the [`UserOperationOverrides`](/resources/types#useroperationoverrides) to be used during the gas estimation.
+To take advantage of such update, `aa-sdk` supports `stateOverride` value as part of the [`UserOperationOverrides`](/resources/types#useroperationoverrides) to be used during the gas estimation of simualted user operation transactions.
```
dphilipson, line 188: Typo: "simualted" -> "simulated"

dphilipson (dismissed):
Alright, looks good, thank you for all the revisions! One last typo and I think it's good to go.

PR #606, opened 2024-04-21
feat: handle one off bypassing paymaster and data middleware case

denniswon:
breaking change: ***- dummyPaymasterAndData middleware override used to be () => Hex, but in this PR, changed to ClientMiddlewareFn just like all other middlewares.***

Update: breaking change avoided with dummyPaymasterAndData middleware override being changed to:

```
() =>
      | UserOperationRequest<"0.6.0">["paymasterAndData"]
      | Pick<UserOperationRequest<"0.7.0">, "paymaster" | "paymasterData">;
```

To make it not breaking change, and also make it not as confusing for entrypoint v0.7 operations.

on the next major version update, we (should) would change the override.dummyPaymasterAndData to be `ClientMiddlewareFn` as well like other middleware.

```
yarn test:run-e2e -- -t "simple account v7 should successfully execute with alchemy paymaster info"
yarn run v1.22.22
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
$ vitest run --config vitest.config.e2e.ts -t 'simple account v7 should successfully execute with alchemy paymaster info'

 RUN  v0.31.4 /Users/denniswon/alchemy/aa-sdk/packages/alchemy

 ↓ |aa-alchemy| e2e-tests/light-account.e2e.test.ts (13) [skipped]
 ↓ |aa-alchemy| e2e-tests/modular-account-multi-owner.e2e.test.ts (12) [skipped]
 ↓ |aa-alchemy| e2e-tests/multi-owner-light-account.e2e.test.ts (12) [skipped]
 ↓ |aa-alchemy| e2e-tests/multisig-account.e2e.test.ts (5) [skipped]
 ✓ |aa-alchemy| e2e-tests/simple-account-v7.e2e.test.ts (13) 19749ms

 Test Files  1 passed | 4 skipped (5)
      Tests  1 passed | 54 skipped (55)
   Start at  21:59:30
   Duration  21.82s (transform 56ms, setup 28ms, collect 1.39s, tests 19.75s, environment 0ms, prepare 53ms)

✨  Done in 22.41s.
```

e2e test for simple account with entrypoint v0.7 and alchemy smart account client + gas manager passes. cc: @dphilipson @moldy530 

# Pull Request Checklist

---

## PR-Codex overview
This PR introduces enhancements to paymaster and data handling in various middleware functions and user operation structures.

### Detailed summary
- Added functions `bypassPaymasterAndData` and `concatPaymasterAndData`
- Updated `dummyPaymasterAndData` type in multiple files
- Added `parsePaymasterAndData` function
- Updated `UserOperationRequest` types in multiple files

> The following files were skipped due to too many changes: `site/packages/aa-core/smart-account-client/index.md`, `packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts`, `packages/core/src/middleware/actions.ts`, `packages/core/src/utils/userop.ts`, `packages/alchemy/e2e-tests/simple-account-v7.e2e.test.ts`, `packages/alchemy/src/middleware/gasManager.ts`


packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts
```diff
@@ -51,7 +56,9 @@ export async function _runMiddlewareStack<
     client.middleware.feeEstimator,
     client.middleware.gasEstimator,
     client.middleware.customMiddleware,
-    client.middleware.paymasterAndData,
+    overrides && bypassPaymasterAndData(overrides)
+      ? bypassPaymasterMiddleware
+      : client.middleware.paymasterAndData,
```
moldy530, line 61: we shouldn't do this here. if we do, then none of the code you have written in the alchemy paymaster will even run. we should instead let paymaster middlewares handle paymaster overrides as they want to. some paymaster middlewares might be ok with overrides that are non `0x`

denniswon: that was the logic/reasoning behind making that change in the pr https://github.com/alchemyplatform/aa-sdk/pull/580 that you @moldy530 are suggesting here.

However, during testing, @dphilipson and @alex-miao gave a devEx feedback that for passing in '0x', the expectation would be to bypass the configured paymaster. That was why this was added.

cc: @avasisht23 @rthomare

moldy530: what I'm saying is that if you add this here then https://github.com/alchemyplatform/aa-sdk/pull/606/files#diff-cf46ef1442acb1802f8f34e60aea43e0ef05ce00d0078f2018ba2b1bf3dca895R151 will never be hit. If the idea is that passing in `0x` should skip all middleware then there's no need to add the changes to the `gasManager` middleware

packages/core/src/middleware/defaults/bypassPaymasterMiddleware.ts
```diff
@@ -0,0 +1,29 @@
+import type { UserOperationOverrides, UserOperationStruct } from "../../types";
+import type { ClientMiddlewareFn } from "../types";
+
+export const bypassPaymasterMiddleware: ClientMiddlewareFn = async (
```
moldy530, line 4: why not just use the `noopMiddleware`? you're already checking if `paymaster` === `0x` before this ever gets called so this is not doing anything extra

packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts
alex-miao: As a user I wouldn't expect that it would skip all middleware but rather it run all the middleware and just overriding the single overridden value at the end.

dphilipson: Hm, I might actually expect it would skip the paymaster middleware. If I'm going to be overriding the paymaster information, I wouldn't expect anything of value to come out of running the paymaster middleware so doing so adds extra calls and waiting that I don't need.

On top of that, it's likely that the reason I'm overriding the paymaster is because the operation I'm sending is incompatible with the paymaster in the middleware, so if I do try to run the paymaster middleware on my op then the middleware would throw.

dphilipson: (I agree with moldy's point about how adding the check here makes it unnecessary to add the check in the gas manager middleware)

moldy530: Oh I misread and thought that check was happening in the paymasterAndData middleware, not the overridden fee and gas estimators 👍 Yea the fee and gas estimator will run

packages/core/src/middleware/defaults/bypassPaymasterMiddleware.ts
moldy530: > specifically, the paymaster gas limits that were supplied from the gas estimator.

if the override paymaster field is `0x` then the gas estimator shouldn't be returning any paymaster fields. The middleware should result in no paymaster code running anywhere and no paymaster fields being set earlier in the pipeline. Noop should work here based on that

packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts
denniswon: yes

packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts
```diff
@@ -51,7 +56,9 @@ export async function _runMiddlewareStack<
     client.middleware.feeEstimator,
     client.middleware.gasEstimator,
     client.middleware.customMiddleware,
-    client.middleware.paymasterAndData,
+    overrides && bypassPaymasterAndData(overrides)
+      ? bypassPaymasterMiddleware
```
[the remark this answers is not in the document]
denniswon: yes it could, now that gas Estimator response includes `paymasterVerificationGas`, and this should be omitted for bypassing paymaster.

packages/core/src/middleware/defaults/bypassPaymasterMiddleware.ts
denniswon: yes it could, now that gas Estimator response includes paymasterVerificationGas, and this should be omitted for bypassing paymaster.

Thus, bypassmiddleware is not the same as noopmiddleware.

If paymaster is configured, gas Estimator response WILL include the paymasterVerificationGas field.

packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts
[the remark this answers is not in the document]
moldy530: > now that gas Estimator response includes paymasterVerificationGas, and this should be omitted for bypassing paymaster

Why is that? it should respect the `overrides.paymasterAndData === "0x"`?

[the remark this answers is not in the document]
moldy530: like here for example: https://github.com/alchemyplatform/aa-sdk/pull/606/files#diff-cf46ef1442acb1802f8f34e60aea43e0ef05ce00d0078f2018ba2b1bf3dca895R128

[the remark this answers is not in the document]
moldy530: I would expect that all middlewares that can return paymaster fields respect the `paymasterAndData` override if it is passed

packages/core/src/middleware/defaults/bypassPaymasterMiddleware.ts
denniswon: ditto: my question to @dphilipson 

"okay, @dphilipson just to confirm, if paymaster address is not set on the user op struct sent for estimate user operation gas RPC, the bundler will never send the paymaster gas limits? Not just rundler, but all other bundlers. This is not clear on the 4337 standards."

packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts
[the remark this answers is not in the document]
dphilipson: Hm, the standard says, in the estimation response:

* **paymasterVerificationGasLimit** value used for paymaster verification (if paymaster exists in the UserOperation)
* **paymasterPostOpGasLimit** value used for paymaster post op execution (if paymaster exists in the UserOperation)

I would interpret that as saying that it doesn't return values if there's no paymaster, either omitting the fields entirely or returning `null`. Our bundler does one of these, don't remember which (I would guess returning null).

I'm inclined to say it's okay to not go out of the way to delete the fields here and just trust whatever the bundler is returning. For one thing, I would expect any bundler which is returning zeroes for the paymaster fields when there's no paymaster would also **accept** zeroes for the paymaster fields when there's no paymaster, so it should be fine.

denniswon:
updated the PR with:
additional omit paymaster field from gas estimator for bypass case
Added a number of missing jsDocs

packages/alchemy/src/middleware/gasManager.ts
```diff
@@ -123,15 +123,9 @@ export function alchemyGasManagerMiddleware<C extends ClientWithAlchemyMethods>(
...
-          const entryPoint = account.getEntryPoint();
           // if user is bypassing paymaster to fallback to having the account to pay the gas (one-off override),
           // we cannot delegate gas estimation to the bundler because paymaster middleware will not be called
-          if (
-            (entryPoint.version === "0.6.0"
-              ? (overrides as UserOperationOverrides<"0.6.0">)?.paymasterAndData
-              : (overrides as UserOperationOverrides<"0.7.0">)
-                  ?.paymasterData) === "0x"
-          ) {
+          if (overrides && bypassPaymasterAndData(overrides)) {
```
dphilipson, line 128: Correct me if I'm wrong, but since `bypassPaymasterAndData` accepts `undefined`, you don't need the `overrides &&` here, likewise below.

dphilipson (approved):
This looks good to me. Really appreciate all the doc comments. I'm not as familiar with the middleware stack as @moldy530 so I'd definitely appreciate a second look if he has a chance.

Question since I didn't notice it mentioned anywhere: do we also need to disable the `dummyPaymasterAndData` middleware when there's a "0x" paymaster override?

denniswon:
@dphilipson we don't have control over the overridden dummy paymaster middleware. When the user sets the paymaster (for example gas manager), both dummy paymaster middleware and paymaster middleware is overridden. 

The dummy paymaster middleware is supposed to set the paymaster address + dummy data for the dummy paymaster middleware. (we should add this to our doc of using custom paymaster)

denniswon:
Asana task https://app.asana.com/0/1205598840815267/1207162808443182/f for "The dummy paymaster middleware is supposed to set the paymaster address + dummy data for the dummy paymaster middleware. (we should add this to our doc of using custom paymaster)" @dphilipson

moldy530:
> we don't have control over the overridden dummy paymaster middleware. When the user sets the paymaster (for example gas manager), both dummy paymaster middleware and paymaster middleware is overridden

wait why don't we just do what we do with dummyPaymaster? we can skip running dummy paymaster if the override is `0x` and just noop there. that means nothing downstream will get a paymaster field so doesn't matter if people override the gas estimator

packages/core/src/client/types.ts
```diff
@@ -42,7 +42,7 @@ export type ClientMiddlewareConfig<
   "dummyPaymasterAndData" | "paymasterAndData"
 > & {
   paymasterAndData?: {
-    dummyPaymasterAndData: () => Hex;
+    dummyPaymasterAndData: ClientMiddlewareFn<TContext>;
```
denniswon, line 45: @dphilipson @moldy530 this would be a breaking change for devs who are overriding this dummy paymaster and data middleware.

But this is necessary because from dummy paymaster and data, we need to check the entrypoint version from the account (either hoisted or passed in).

@moldy530 let me know if you have an idea for preventing breaking change for this.

packages/alchemy/src/middleware/gasManager.ts
```diff
@@ -75,34 +96,106 @@ export type RequestGasAndPaymasterAndDataResponse<
...
+    config.dummyData ??
+    "fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0000000000000000000000000000000007aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa1c";
 
-    return `${paymasterAddress}${dummyData}` as Address;
-  };
+  return {
+    ...struct,
+    ...(account.getEntryPoint().version === "0.7.0"
+      ? {
+          paymaster: `${paymasterAddress}`,
```
dphilipson, line 183: Do you not need to set `paymasterData: dummyData` in this case?

dphilipson (approved):
Had a question. If the answer is "no", then lgtm.

denniswon:
breaking change: ***- dummyPaymasterAndData middleware override used to be () => Hex, but in this PR, changed to ClientMiddlewareFn just like all other middlewares.***

packages/core/src/middleware/defaults/bypassPaymasterMiddleware.ts
moldy530, line 1: I thought we landed on the fact that you don't need this?

packages/core/src/middleware/actions.ts
```diff
@@ -45,16 +61,9 @@ export const middlewareActions =
...
-      dummyPaymasterAndData: overrides.paymasterAndData?.dummyPaymasterAndData
-        ? async (struct, { account }) =>
-            account.getEntryPoint().version === "0.6.0"
-              ? {
-                  ...struct,
-                  paymasterAndData:
-                    overrides.paymasterAndData!.dummyPaymasterAndData(),
-                }
-              : struct
-        : defaultPaymasterAndData,
```
moldy530, line 57: This was fine as it was and just needed support for `0.7.0`

```ts
async (struct, { account }) =>
            account.getEntryPoint().version === "0.6.0"
              ? {
                  ...struct,
                  paymasterAndData:
                    overrides.paymasterAndData!.dummyPaymasterAndData(),
                }
              : {
                 paymaster: // take the first 20 bytes of overrides.paymasterAndData!.dummyPaymasterAndData(),
                 paymasterData: // take the remaining bytes
               }
```

packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts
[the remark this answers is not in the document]
moldy530: if above where we have `client.middleware.dummyPaymaster` run we add the same check for the override and just noop then we don't need this bypass function at all

packages/core/src/middleware/actions.ts
denniswon: paymasterData? I thought we only set 'paymaster' field. @dphilipson do we need to set dummy paymasterData as well for gas estimation and paymaster query?

denniswon: @moldy530 okay so just parse from dummyPaymasterAndData returned Hex as Address?

dphilipson (commented):
Actually want to resolve the discussion in slack, taking back my approval for now. Sorry!

packages/core/src/utils/userop.ts
```diff
@@ -120,3 +148,50 @@ export function applyUserOpOverrideOrFeeOption(
...
+ *                         entrypoint v0.7 user operation request
+ * @param paymasterAndData.paymaster the paymaster address
+ * @param paymasterAndData.paymasterData the paymaster data
+ * @returns the paymasterAndData hex value of entrypoint v0.6 user operation request paymasterAndData field
+ */
+export const concatPaymasterAndData = ({
+  paymaster,
+  paymasterData,
+}: Pick<UserOperationRequest<"0.7.0">, "paymaster" | "paymasterData">): Hex =>
+  `${paymaster}${paymasterData}` as Hex;
```
dphilipson, line 197: This is going to end up with an "0x" in the middle, right? I'd suggest using  [`concat` from viem](https://viem.sh/docs/utilities/concat) and writing
```ts
concat(paymaster, paymasterData)
```

packages/core/src/middleware/actions.ts
denniswon: discussed here https://alchemyinsights.slack.com/archives/C0669LA6JQH/p1714084846367119

and updated with the aligned approach

packages/core/src/middleware/actions.ts
```diff
@@ -47,13 +68,33 @@ export const middlewareActions =
...
+            account.getEntryPoint().version === "0.7.0"
               ? {
                   ...struct,
-                  paymasterAndData:
-                    overrides.paymasterAndData!.dummyPaymasterAndData(),
+                  ...((isHex(
+                    overrides.paymasterAndData!.dummyPaymasterAndData()
+                  )
+                    ? parsePaymasterAndData(
+                        overrides.paymasterAndData!.dummyPaymasterAndData() as Hex
```
dphilipson, line 78: If you put `overrides.paymasterAndData!.dummyPaymasterAndData()` into a temp variable, you wouldn't need these casts because `isHex` would work as a type guard (plus you wouldn't need to keep writing this large expression over and over). I'd suggest:
```ts
overrides.paymasterAndData?.dummyPaymasterAndData
        ? async (struct, { account }) => {
            const data = overrides.paymasterAndData!.dummyPaymasterAndData();
            return account.getEntryPoint().version === "0.7.0"
              ? {
                  ...struct,
                  ...(isHex(data)
                    ? parsePaymasterAndData(data)
                    : concatPayamsterAndData(data)
                  ),
                }
                // ...
```
and maybe consider turning some of ternaries into if-else blocks for readability.

dphilipson (commented):
Think it needs a hex concat fix, but otherwise I like it!

packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts
```diff
@@ -46,12 +73,21 @@ export async function _runMiddlewareStack<
     throw new AccountNotFoundError();
   }
 
+  const { dummyPaymasterAndData, paymasterAndData } = {
+    dummyPaymasterAndData: bypassPaymasterAndData(overrides)
+      ? bypassPaymasterMiddleware
```
moldy530, line 78:
```suggestion
      ? noopMiddleware
```

packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts
```diff
@@ -46,12 +73,21 @@ export async function _runMiddlewareStack<
...
+      : client.middleware.dummyPaymasterAndData,
+    paymasterAndData: bypassPaymasterAndData(overrides)
+      ? bypassPaymasterMiddleware
```
moldy530, line 81:
```suggestion
      ? noopMiddleware
```

packages/core/src/middleware/defaults/bypassPaymasterMiddleware.ts
moldy530, line 1: kill this

packages/core/src/middleware/actions.ts
denniswon: > `isHex would work as a type guard `

Actually it doesn't which is why had to cast...

packages/core/src/actions/smartAccount/internal/runMiddlewareStack.ts
denniswon: we need to set `paymasterAndData` field v0.6 to `0x`. Noop just passthrough the struct.
I guess during resolveProperties, undefined field is set to '0x'

moldy530: yea setting to 0x should be handled already. otherwise this wouldn't have worked without this function before

packages/core/src/middleware/actions.ts
dphilipson: It didn't work before because you're evaluating `dummyPaymasterAndData()` multiple times, so the type system can't tell that it's the same value. `isHex` should work as a type guard if you extract the value, its signature is
```ts
function isHex(
  value: unknown,
  { strict = true }: { strict?: boolean } = {},
): value is Hex;
```

dphilipson: Here's what it looks like without the casts using `isHex` as a type guard, if you want it. Verified locally.
```ts
      dummyPaymasterAndData: overrides.paymasterAndData?.dummyPaymasterAndData
        ? async (struct, { account }) => {
            const data = overrides.paymasterAndData!.dummyPaymasterAndData();
            const paymasterOverrides =
              account.getEntryPoint().version === "0.7.0"
                ? isHex(data)
                  ? parsePaymasterAndData(data)
                  : data
                : {
                    paymasterAndData: isHex(data)
                      ? data
                      : concatPaymasterAndData(data),
                  };
            return { ...struct, ...paymasterOverrides };
          }
        : defaultPaymasterAndData,
```

packages/core/src/utils/userop.ts
```diff
@@ -120,3 +148,50 @@ export function applyUserOpOverrideOrFeeOption(
...
+ * Utility method for parsing the paymaster address and paymasterData from the paymasterAndData hex string
+ *
+ * @param paymasterAndData the paymaster and data hex string to parse.
+ *                         The hex string refers to the paymasterAndData field of entrypoint v0.6 user operation request
+ * @returns the parsed paymaster and paymasterData fields of entrypoint v0.7 user operation request paymaster and paymasterData field
+ */
+export const parsePaymasterAndData = (
+  paymasterAndData: Hex
+): Pick<UserOperationRequest<"0.7.0">, "paymaster" | "paymasterData"> => ({
+  paymaster: `0x${paymasterAndData.substring(2, 42)}` as Address,
```
dphilipson, line 179: Nit: you can have
```ts
paymaster: paymasterAndData.substring(0, 42) as Address
```
and you don't need the template stirng.

packages/core/src/utils/userop.ts
```diff
@@ -120,3 +148,50 @@ export function applyUserOpOverrideOrFeeOption(
...
+ *                         entrypoint v0.7 user operation request
+ * @param paymasterAndData.paymaster the paymaster address
+ * @param paymasterAndData.paymasterData the paymaster data
+ * @returns the paymasterAndData hex value of entrypoint v0.6 user operation request paymasterAndData field
+ */
+export const concatPaymasterAndData = ({
+  paymaster = "0x",
+  paymasterData = "0x",
+}: Pick<UserOperationRequest<"0.7.0">, "paymaster" | "paymasterData">): Hex =>
+  toHex(concat([paymaster, paymasterData]));
```
dphilipson, line 197: I don't think you want `toHex` here. The output of `concat` is already hex; calling `toHex` converts a string into its ASCII representation, e.g. `toHex("0x10") === "0x30783130").

dphilipson (changes requested):
Think you have a hex conversion bug. If we get that fixed I think we're there.

site/packages/aa-core/smart-account-client/index.md
```diff
@@ -115,7 +115,7 @@ A new instance of a `SmartAccountClient`.
 - `paymasterAndData?: ClientMiddlewareConfig["paymasterAndData"]` -- [optional] if you would like to use a paymaster, then this config must be supplied.
 
   - `paymasterAndData: ClientMiddlewareFn` -- this middleware function calculates the paymaster and data fields of your User Operation after estimating gas and fees
-  - `dummyPaymasterAndData: () => Hex` -- this just returns a Hex string to be used during gas and fee estimation. This will depend on your paymaster provider and must be a value that accurately resembles the gas cost of using your paymaster and does not revert during validation.
```
moldy530, line 118: Should revert this change

site/packages/aa-core/smart-account-client/middleware/index.md
```diff
@@ -38,7 +38,7 @@ Each middleware is a function that takes in a user operation object, performs it
 
 ### paymasterAndData
 
-`paymasterAndData` middleware is responsible for computing `paymasterAndData` fields of user operation after estimating gas and fees. `dummyPaymasterAndData` is a dummy middleware that returns a Hex string for gas and fee estimation. This will depend on your paymaster provider and must be a value that accurately resembles the gas cost of using your paymaster and does not revert during validation.
```
moldy530, line 41: Same here

site/third-party/paymasters.md
```diff
@@ -33,16 +33,22 @@ const client = createSmartAccountClient({
   chain,
   transport: http("RPC_URL"),
   paymasterAndData: {
+    dummyPaymasterAndData: async (userop, opts) => {
```
moldy530, line 36: And this one

moldy530 (changes requested):
Gotta change the docs for he dummy paymaster and data to what it was before + the new union type

packages/core/src/utils/userop.ts
denniswon: thanks. good catch. updtaed.

denniswon:
updated thanks for the thorough review! @dphilipson @moldy530

dphilipson (approved):
Looks good! I would still consider my suggested change with calling `dummyPaymasterAndData` once, not just because it doesn't need casts but also because calling a user-provided function once instead of multiple times can be a bit better e.g. if they add log statements to it for debugging, but I won't block because of that.

denniswon:
@dphilipson udpated per your suggestion. Thanks!

PR #613, opened 2024-04-22
feat: add frax and zora defaults

alex-miao:
# Pull Request Checklist

---

## PR-Codex overview
**Focus:** Introducing support for Fraxtal, Fraxtal Testnet, Zora, and Zora Sepolia networks in the Alchemy platform.

### Detailed summary
- Added support for `fraxtal`, `fraxtalTestnet`, `zora`, and `zoraSepolia` chains
- Updated various configurations and addresses for the newly supported networks

### DIFF
--- a/packages/accounts/plugindefs/multisig/config.ts
+++ b/packages/accounts/plugindefs/multisig/config.ts
@@ -9,8 +9,11 @@ import {
   arbitrum,
   arbitrumSepolia,
   base,
+  fraxtal,
+  fraxtalTestnet,
+  zora,
+  zoraSepolia,
 } from "@alchemy/aa-core";
-import { fraxtal, fraxtalTestnet, zora, zoraSepolia } from "viem/chains";
 import { parseAbiParameters } from "viem";
 import type { PluginGenConfig } from "../types";
 import { MultisigPluginAbi } from "./abi.js";
--- a/packages/accounts/plugindefs/multisig/config.ts
+++ b/packages/accounts/plugindefs/multisig/config.ts
@@ -10,6 +10,7 @@ import {
   arbitrumSepolia,
   base,
 } from "@alchemy/aa-core";
+import { fraxtal, fraxtalTestnet, zora, zoraSepolia } from "viem/chains";
 import { parseAbiParameters } from "viem";
 import type { PluginGenConfig } from "../types";
 import { MultisigPluginAbi } from "./abi.js";
@@ -28,6 +29,10 @@ export const MultisigPluginGenConfig: PluginGenConfig = {
     [arbitrum.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
     [arbitrumSepolia.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
     [base.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
+    [fraxtal.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
+    [fraxtalTestnet.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
+    [zora.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
+    [zoraSepolia.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
   },
   chain: sepolia,
   installConfig: {
--- a/packages/accounts/src/light-account/utils.ts
+++ b/packages/accounts/src/light-account/utils.ts
@@ -14,10 +14,13 @@ import {
   polygonAmoy,
   polygonMumbai,
   sepolia,
+  fraxtal,
+  fraxtalTestnet,
+  zora,
+  zoraSepolia,
   toRecord,
   type SmartContractAccount,
 } from "@alchemy/aa-core";
-import { fraxtal, fraxtalTestnet, zora, zoraSepolia } from "viem/chains";
 import { fromHex, type Address, type Chain } from "viem";
 import type { LightAccountBase } from "./accounts/base";
 import type {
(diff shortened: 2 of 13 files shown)


packages/accounts/plugindefs/multisig/config.ts
```diff
@@ -28,6 +29,10 @@ export const MultisigPluginGenConfig: PluginGenConfig = {
     [arbitrum.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
     [arbitrumSepolia.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
     [base.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
+    [fraxtal.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
+    [fraxtalTestnet.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
+    [zora.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
+    [zoraSepolia.id]: "0x000000000000A53f64b7bcf4Cd59624943C43Fc7",
```
denniswon, line 35: are our 4 plugins deployed and available in all these chains? cc: @adam-alchemy

packages/alchemy/src/gas-manager.ts
```diff
@@ -4,18 +4,15 @@ import {
...
   optimism,
   optimismSepolia,
   polygon,
   polygonAmoy,
   polygonMumbai,
-  sepolia,
-  zora,
-  zoraSepolia,
-} from "viem/chains";
+} from "@alchemy/aa-core";
```
alex-miao, line 14: Was there a reason we were importing from `viem` instead of `aa-core` before?

packages/accounts/src/light-account/utils.ts
```diff
@@ -43,6 +44,10 @@ export const supportedChains: Chain[] = [
   base,
   baseGoerli,
   baseSepolia,
+  fraxtal,
+  fraxtalTestnet,
+  zora,
+  zoraSepolia,
```
denniswon, line 50: we should add these to aa-core as alchemy chains, and use those. even if they are not alchemy rpc url, can we configure these in aa-core with Blast url? if not, let's at least add comment on why these are from viem and others are not.

packages/accounts/src/msca/utils.ts
```diff
@@ -97,6 +98,10 @@ export const getDefaultMultiOwnerModularAccountFactoryAddress = (
     case arbitrumSepolia.id:
     case base.id:
     case baseGoerli.id:
+    case fraxtal.id:
+    case fraxtalTestnet.id:
+    case zora.id:
```
denniswon, line 103: ditto as above

packages/core/src/entrypoint/0.6.ts
```diff
@@ -80,6 +81,10 @@ export default {
     [base.id]: "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789",
     [baseGoerli.id]: "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789",
     [baseSepolia.id]: "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789",
+    [fraxtal.id]: "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789",
+    [fraxtalTestnet.id]: "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789",
+    [zora.id]: "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789",
+    [zoraSepolia.id]: "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789",
```
denniswon, line 87: ditto as above

packages/accounts/src/light-account/utils.ts
alex-miao: Would adding a non-Alchemy URL would break things because we assume we can append an Alchemy API key to the URLs

packages/alchemy/src/gas-manager.ts
moldy530: nope I think that was a mistake, good catch

packages/alchemy/src/gas-manager.ts
```diff
@@ -4,18 +4,15 @@ import {
...
+import { fraxtal, fraxtalTestnet, zora, zoraSepolia } from "viem/chains";
```
moldy530, line 15: +1 to dennis's comment about consolidating these into aa-core

alex-miao: Added to aa-core without alchemy url

packages/accounts/src/light-account/utils.ts
alex-miao: Added to aa-core without an alchemy base url

moldy530 (commented):
LGTM, but we should add a doc that explains how someone might use Zora or Frax. Namely, they will need to set the RPC url in their transport (or for AlchemyClient the `rpcUrl` property) to be their backend which should handle splitting traffic to our bundler and their node url of choice.

We'll follow up and add better split traffic support to the alchemy smart account client

site/third-party/bundlers.md
```diff
@@ -41,3 +41,7 @@ with your provider on what the correct logic is.
 It might be the case that you want to use a different RPC provider for your bundler traffic and your node traffic. This is a common use case, and you can do this by leveraging the [`split`](/packages/aa-core/split-transport) transport and passing it to your `createSmartAccountClient` call. For example:
 
 <<< @/snippets/aa-core/splitTransport.ts
+
+## Zora and Fraxtal
+
+Using a split Bundler and Node RPC setup is required for Fraxtal, Fraxtal Testnet, Zora, and Zora Sepolia networks since Alchemy currently only supports Account Abstraction endpoints for those networks.
```
moldy530, line 47: do we have any suggested 3rd party providers for RPC for theses chains? might be helpful to link out to them here

moldy530 (approved):
LGTM, maybe add a few RPC providers for each frax and zora?

site/third-party/bundlers.md
alex-miao: Good idea, added links

moldy530 (approved):

PR #614, opened 2024-04-22
feat: revised LightAccount v2 1271 sigs

dphilipson:
Updates the `signMessage` and `signTypedData` implementations for `LightAccount` v2. The updated version no longer attempts to use nested typed data to display the original message to the user, as doing so creates a security risk in which valid signing requests can be disguished in a way that concels their original contents.

Instead, the new `LightAccount` v2 uses the same signing mechanism as v1.1.0, except that it also prepends an enum indicating if the owner is an EOA or smart contract account, as a way to both save gas and to provide more accurate gas estimates.

---

## PR-Codex overview
This PR refactors the `createLightAccountBase` function in `base.ts` to remove usage of `getErc1271SigningFunctions` and directly call `signWith1271WrapperV1`.

### Detailed summary
- Removed `getErc1271SigningFunctions` import and direct calls to `signWith1271WrapperV1`.
- Updated signature handling for different account versions.
- Removed unused functions `hashDomain.ts` and `hashTypedData.ts`.


denniswon (commented):
LGTM, but do we need to remove the files for 1271? no big opinion. wdyt? @moldy530

packages/accounts/src/light-account/accounts/base.ts
```diff
@@ -242,7 +231,7 @@ export async function createLightAccountBase({
         case "v1.1.0":
           return signWith1271WrapperV1(hashMessage(message));
```
jaypaik, line 232: Should this be `await`ed too?

jaypaik (approved):
🐐

packages/accounts/src/light-account/accounts/base.ts
```diff
@@ -242,7 +231,7 @@ export async function createLightAccountBase({
...
         case "v2.0.0":
-          const signature = await signMessageV2({ message });
+          const signature = await signWith1271WrapperV1(hashMessage(message));
```
moldy530, line 234: ah ok so we decided to keep the v1 approach for v2?

moldy530 (approved):

moldy530:
> but do we need to remove the files for 1271
yea we should delete them and ideally open a pr with viem to have them export them

packages/accounts/src/light-account/accounts/base.ts
dphilipson: `return x` and `return await x` do the same thing (unless in a try-block), so the linter would probably complain at me.

dphilipson:
Yeah, no need to include functions we're not using, we can always get them back from version control if we need them again.

PR #607, opened 2024-04-21
feat: remove generic entry point version specification requirement for better devex

denniswon:
# Pull Request Checklist

- [ ] Did you add new tests and confirm existing tests pass? (`yarn test`)
- [ ] Did you update relevant docs? (docs are found in the `site` folder, and guidelines for updating/adding docs can be found in the [contribution guide](https://github.com/alchemyplatform/aa-sdk/blob/main/CONTRIBUTING.md))
- [ ] Do your commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [ ] Does your PR title also follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [ ] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples)? (e.g. `feat!: breaking change`)
- [ ] Did you run lint (`yarn lint:check`) and fix any issues? (`yarn lint:write`)
- [ ] Did you follow the [contribution guidelines](https://github.com/alchemyplatform/aa-sdk/blob/main/CONTRIBUTING.md)?

---

## PR-Codex overview
This PR updates various TypeScript files across packages to adjust type constraints and improve code readability.

### Detailed summary
- Updated type constraints in multiple TypeScript files for better readability and maintainability
- Added utility method `getDefaultLightAccountFactoryAddress` to retrieve default light account factory address
- Refactored error handling in `InvalidUserOperationError` constructor
- Adjusted type guards and type definitions for better type safety and clarity

> The following files were skipped due to too many changes: `packages/alchemy/src/client/types.ts`, `packages/core/src/types.ts`, `packages/alchemy/src/middleware/gasManager.ts`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #615, opened 2024-04-23
fix: make multisig e2e tests pass

moldy530:
# Pull Request Checklist

---

## PR-Codex overview
The focus of this PR is to introduce the `import` plugin, update data types, and enhance multisig functionality.

### Detailed summary
- Added `import` plugin to `.eslintrc`
- Updated data types in several files
- Enhanced multisig functionality in various plugins and actions

> The following files were skipped due to too many changes: `packages/accounts/src/msca/plugins/multisig/utils/formatSignatures.ts`, `packages/core/src/actions/smartAccount/types.ts`, `packages/core/src/client/decorators/smartAccountClient.ts`, `packages/core/src/actions/smartAccount/internal/sendUserOperation.ts`, `packages/core/src/actions/smartAccount/buildUserOperationFromTx.ts`, `packages/accounts/src/msca/plugins/multisig/actions/signMultisigUserOperation.ts`, `packages/core/src/actions/smartAccount/checkGasSponsorshipEligibility.ts`, `packages/alchemy/e2e-tests/multisig-account.e2e.test.ts`, `packages/core/src/actions/smartAccount/buildUserOperationFromTxs.ts`, `packages/accounts/src/msca/plugins/multisig/middleware.ts`, `packages/accounts/src/msca/e2e-tests/multisig-modular-account.e2e.test.ts`

### DIFF
--- a/packages/accounts/src/msca/plugins/multisig/actions/proposeUserOperation.ts
+++ b/packages/accounts/src/msca/plugins/multisig/actions/proposeUserOperation.ts
@@ -10,8 +10,8 @@ import {
   type UserOperationOverrides,
 } from "@alchemy/aa-core";
 import { type Chain, type Client, type Transport } from "viem";
-import { combineSignatures, getSignerType } from "../index.js";
-import { type ProposeUserOperationResult, type Signature } from "../types.js";
+import { splitAggregatedSignature } from "../index.js";
+import { type ProposeUserOperationResult } from "../types.js";
 
 export async function proposeUserOperation<
   TTransport extends Transport = Transport,
@@ -61,29 +61,22 @@ export async function proposeUserOperation<
   const request = await client.signUserOperation({
     uoStruct: builtUo,
     account,
+    context: {
+      userOpSignatureType: "UPPERLIMIT",
+    },
   });
 
-  const signerType = await getSignerType({
-    client,
-    signature: request.signature,
-    signer: account.getSigner(),
+  const splitSignatures = await splitAggregatedSignature({
+    request,
+    aggregatedSignature: request.signature,
+    account,
+    // split works on the assumption that we have t - 1 signatures
+    threshold: 2,
   });
-
-  const signatureObj: Signature = {
-    signature: request.signature,
-    signer: await account.getSigner().getAddress(),
-    signerType,
-    userOpSigType: "UPPERLIMIT",
-  };
 
   return {
     request,
-    signatureObj,
-    aggregatedSignature: combineSignatures({
-      signatures: [signatureObj],
-      upperLimitMaxFeePerGas: request.maxFeePerGas,
-      upperLimitMaxPriorityFeePerGas: request.maxPriorityFeePerGas,
-      upperLimitPvg: request.preVerificationGas,
-    }),
+    signatureObj: splitSignatures.signatures[0],
+    aggregatedSignature: request.signature,
   };
 }
--- a/packages/accounts/src/msca/plugins/multisig/middleware.ts
+++ b/packages/accounts/src/msca/plugins/multisig/middleware.ts
@@ -6,23 +6,47 @@ import {
   isValidRequest,
   resolveProperties,
   type ClientMiddlewareFn,
+  type UserOperationRequest_v6,
+  type UserOperationRequest_v7,
 } from "@alchemy/aa-core";
-import { isHex, type Hex } from "viem";
+import { type Hex } from "viem";
 import { isMultisigModularAccount } from "../../account/multisigAccount.js";
 import {
   InvalidContextSignatureError,
   MultisigAccountExpectedError,
 } from "../../errors.js";
-import { getThreshold } from "./actions/getThreshold.js";
 import {
   combineSignatures,
   getSignerType,
   splitAggregatedSignature,
+  type MultisigUserOperationContext,
 } from "./index.js";
 
-export const multisigSignatureMiddleware: ClientMiddlewareFn<{
-  signature: Hex;
-}> = async (struct, { account, client, context }) => {
+/**
+ * A signer middleware to be used with Multisig Account Clients.
+ * This middleware handles correctly aggregating signatures passed through
+ * as context when sending UserOperations, proposing UserOperations, or adding signatures to a UserOperation.
+ *
+ * @param struct the user operation struct to be signed
+ * @param param the parameters to be passed to the middleware
+ * @param param.account the account to be used for signing
+ * @param param.client the smart account client that will be used for RPC requests
+ * @param param.context the context object containing the signatures to be aggregated {@link MultisigUserOperationContext}
+ * @returns a Promise containing a UserOperation with an aggregated signature in the `signature` field
+ */
+export const multisigSignatureMiddleware: ClientMiddlewareFn<
+  MultisigUserOperationContext
+> = async (struct, { account, client, context }) => {
+  // if the signature is not present, this has to be UPPERLIMIT because it's likely a propose operation
+  if (
+    !context ||
+    (context.userOpSignatureType === "ACTUAL" &&
+      !context.signatures &&
+      !context.aggregatedSignature)
+  ) {
+    throw new InvalidContextSignatureError();
+  }
+
   if (!isSmartAccountWithSigner(account)) {
     throw new SmartAccountWithSignerRequiredError();
   }
@@ -32,6 +56,7 @@ export const multisigSignatureMiddleware: ClientMiddlewareFn<{
   }
 
   const resolvedStruct = await resolveProperties(struct);
+
   const request = deepHexlify(resolvedStruct);
   if (!isValidRequest(request)) {
     throw new InvalidUserOperationError(resolvedStruct);
@@ -47,36 +72,50 @@ export const multisigSignatureMiddleware: ClientMiddlewareFn<{
     signer: account.getSigner(),
   });
 
-  // TODO: this needs to actually check the account's installed plugins and fetch the multisig plugin address
-  const threshold = await getThreshold(client, { account });
-
-  // if there is no override, then return the dummy signature
-  if (context?.signature == null) {
+  // then this is a propose operation
+  if (
+    context.userOpSignatureType === "UPPERLIMIT" &&
+    context?.signatures?.length == null &&
+    context?.aggregatedSignature == null
+  ) {
     return {
       ...resolvedStruct,
-      signature: await account.getDummySignature(),
+      signature: combineSignatures({
+        signatures: [
+          {
+            signature,
+            signer: await account.getSigner().getAddress(),
+            signerType,
+            userOpSigType: context.userOpSignatureType,
+          },
+        ],
+        upperLimitMaxFeePerGas: request.maxFeePerGas,
+        upperLimitMaxPriorityFeePerGas: request.maxPriorityFeePerGas,
+        upperLimitPvg: request.preVerificationGas,
+        usingMaxValues: false,
+      }),
     };
   }
 
-  if (!isHex(context.signature)) {
+  if (context.aggregatedSignature == null || context.signatures == null) {
     throw new InvalidContextSignatureError();
   }
 
+  // otherwise this is a sign operation
   const {
     upperLimitPvg,
     upperLimitMaxFeePerGas,
     upperLimitMaxPriorityFeePerGas,
-    signatures,
   } = await splitAggregatedSignature({
-    aggregatedSignature: context.signature,
-    threshold: Number(threshold),
+    aggregatedSignature: context.aggregatedSignature,
+    threshold: context.signatures.length + 1,
     account,
     request,
   });
 
   const finalSignature = combineSignatures({
-    signatures: signatures.concat({
-      userOpSigType: "ACTUAL",
+    signatures: context.signatures.concat({
+      userOpSigType: context.userOpSignatureType,
       signerType,
       signature,
       signer: await account.getSigner().getAddress(),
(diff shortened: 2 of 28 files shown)


packages/core/src/actions/smartAccount/signUserOperation.ts
```diff
@@ -42,12 +43,14 @@ export async function signUserOperation<
...
+  const request = deepHexlify(args.uoStruct);
+  if (!isValidRequest(request)) {
+    throw new InvalidUserOperationError(args.uoStruct);
+  }
+
+  const entryPoint = account.getEntryPoint();
+  request.signature = await account.signUserOperationHash(
+    entryPoint.getUserOperationHash(request)
+  );
+  return request as UserOperationRequest<TEntryPointVersion>;
```
moldy530, line 55: this method shouldn't be using middleware, but instead be calling the account's signUOHash method

packages/core/src/actions/smartAccount/internal/sendUserOperation.ts
```diff
@@ -40,7 +44,7 @@ export async function _sendUserOperation<
 
   const request = await client.middleware
     .signUserOperation(args.uoStruct, {
-      ...args,
+      overrides,
```
denniswon, line 47: did we need overrides here? thought the multisig signatures are in context and that's why we introduced context in addition to overrides?

packages/core/src/actions/smartAccount/sendTransaction.ts
```diff
@@ -62,6 +62,7 @@ export async function sendTransaction<
     account: account as SmartContractAccount,
     uoStruct,
     context,
+    overrides,
```
denniswon, line 65: why do we need overrides passed into the internal _senduserop? Do we use it there? or is this just passed in because overrides are available and there is no reason not to?

packages/core/src/actions/smartAccount/sendUserOperation.ts
```diff
@@ -50,5 +50,6 @@ export const sendUserOperation: <
     account,
     uoStruct,
     context,
+    overrides,
```
denniswon, line 53: ditto: why do we need overrides passed into the internal _senduserop? Do we use it there? or is this just passed in because overrides are available and there is no reason not to?

packages/core/src/actions/smartAccount/internal/sendUserOperation.ts
moldy530: but what if someone else adds custom middleware and needs the overrides?

moldy530: also we do use the overrides to figure out how to correctly craft the aggregated signatures in the middleware

packages/core/src/actions/smartAccount/sendUserOperation.ts
moldy530: see above

packages/core/src/actions/smartAccount/signUserOperation.ts
moldy530: `signUserOperationHash` is meant to be a pass through to the account's method. the default signing middleware just calls `signUoHash` , but custom signer middleware is used to change that logic or leverage context for changing the contents to be signed by the account. Think of it this way, if `client.signUserOperationHash` called the middleware and the middleware is meant to call `signHash`, then you have a circular dependency. The account passthrough actions are just a convenience so instead of doing `client.account.foo` you can just do `client.foo` (eg. `client.account.address` vs `client.getAddress`)

The original intent of the Signer Middleware is that it was used during the `runMiddlewareStack`. multisig expected the signing middleware to run during all calls that call the middleware stack (including `buildUserOperation`). It was meant for custom signing logic not on the account (`account.signUserOperationHash` is for custom signing logic) but for custom signing logic during sending a UO. For multisig, multiple accounts instance might need to sign a UO hash which they do without the client, but when sending a UO (or building a UO) the sending client is responsible for generating a valid UO. This is where the middleware is used

packages/accounts/src/msca/plugins/multisig/actions/proposeUserOperation.ts
```diff
@@ -52,11 +58,15 @@ export async function proposeUserOperation<
...
-    uo,
-    overrides,
-  });
+  const builtUo = await client
+    .buildUserOperation({
+      account,
+      uo,
+      overrides,
+    })
+    .then((x) => multisigSignatureMiddleware(x, { account, client, overrides }))
```
moldy530, line 67: @dphilipson @denniswon want to revisit the argument against having the signer middleware run with all functions that call the middleware? It seems weird to have signing middleware not run as part of the stack and leads to some gnarly maintenance costs. It's too easy to forget to call the `client.middleware.signer` where it's needed.

I'd like to add it back. If the signer allows signing without prompts, then not having it part of the build UO flow doesn't add much safety. So there's not a compelling security argument against it.

packages/core/src/actions/smartAccount/signUserOperation.ts
denniswon: Signer Middleware is NOT used during the `runMiddlewareStack`. middleware pipeline is not supposed to sign the user operation struct.

this signUserOperation action is a SmartAccountClientAction, not SmartAccountAction. runMiddlewareStack does NOT call the signer middleware, but instead _sendUseroperation call as a one-off signer middleware run that defaults to account signUserOperationHash (userOpSigner), but can be customized if the signer middleware is overridden.

denniswon: TLDR: 

> "client.signUserOperationHash called the middleware and the middleware is meant to call signHash, then you have a circular dependency."

--> `signUserOperationHash` is an account method, and client does not have `signUserOperationHash`, but `signUserOperation` action. The signer middleware only default have the connected account call `signUserOperationHash`.

packages/core/src/actions/smartAccount/internal/sendUserOperation.ts
denniswon: oh, because we are not guaranteed anymore that all the middlewares are run before the internal senduseroperation. For example. `signUserOperation` action calls the signer middleware (although you removed on this PR, which I commented about below), and for that middleware, _sendUserOperation should get the overrides.

packages/accounts/src/msca/plugins/multisig/actions/proposeUserOperation.ts
denniswon: no, that is incorrect. It should not run per middleware stack.

checkSponsorshipEligiblity, DropAndReplace, simulateUserOperation should not necessarily have the user operation request signed. 

cc: @dphilipson

denniswon (changes requested):

packages/accounts/src/msca/plugins/multisig/actions/proposeUserOperation.ts
moldy530: see the thread for context as to why it should be in the middleware stack as an opt-in

denniswon: Caught up with that thread https://alchemyinsights.slack.com/archives/C0669LA6JQH/p1713888009511439


However, I still don't understand why signing needs to be in the middleware stack. sign user operation was added as a middleware not because we want all user op building/estimating/simulation etc to sign, but as a mechanism to have custom signing at the control of the user.

It seems like just for propose user op for multi sig, where we want it to be part of the middleware stack? so this is the only "opt-in" case right? Why can't we utilize custom Middleware ? isn't that custom middleware's purpose?

moldy530: > sign user operation was added as a middleware not because we want all user op building/estimating/simulation etc to sign

when I added it that actually was the intent and the original assumption that `proposeUserOperation` made. I've update the approach in this PR for how propose etc works to leverage `signUserOperation`. But if we want to use `signUserOperation` as a method then maybe middleware isn't the best place for that custom logic to live. we can probs actually move it out of middleware entirely and allow it to be an overridable action... lemme take a look at doing that actually cuz I hate it in the middleware object if it's not run as part of the middleware stack

moldy530: > lemme take a look at doing that actually cuz I hate it in the middleware object if it's not run as part of the middleware stack

never mind, this isn't super easy to do actually because of the order in which actions get added it's not possible for the base actions to use overrides. gonna leave it as middleware even though it has no business being middleware anymore

packages/accounts/src/msca/plugins/multisig/middleware.ts
```diff
@@ -47,47 +72,96 @@ export const multisigSignatureMiddleware: ClientMiddlewareFn<{
...
   });
 
   return {
     ...resolvedStruct,
     signature: finalSignature,
   };
 };
+
+const isUsingMaxValues = (
+  request: UserOperationRequest_v6 | UserOperationRequest_v7,
```
denniswon, line 140: you can just do `UserOperationRequest`: this is same thing as union type

with this pr https://github.com/alchemyplatform/aa-sdk/pull/607

packages/core/src/actions/smartAccount/buildUserOperationFromTx.ts
```diff
@@ -19,6 +19,81 @@ import type {
...
+  entryPoint: entryPointAddress,
+});
+```
+ * 
+ * @param client the smart account client to use for RPC requests
+ * @param args {@link SendTransactionParameters}
+ * @param overrides optional {@link UserOperationOverrides} to use for any of the fields
+ * @param context if the smart account client requires additinoal context for building UOs
+ * @returns a Promise containing the built user operation
+ */
```
denniswon, line 96: let's put this on account kit doc (buildUserOperationFromTx or s.t.) and just put a link to that here.

packages/core/src/actions/smartAccount/checkGasSponsorshipEligibility.ts
```diff
@@ -10,7 +10,35 @@ import type {
...
 
-export const checkGasSponsorshipEligibility: <
+/**
+ * This function verifies the eligibility of the connected account for gas sponsorship concerning the upcoming `UserOperation` (UO) that is intended to be sent.
+ * Internally, this method invokes [`buildUserOperation`](./buildUserOperation.md), which navigates through the middleware pipeline, including the `PaymasterMiddleware`. Its purpose is to construct the UO struct meant for transmission to the bundler. Following the construction of the UO struct, this function verifies if the resulting structure contains a non-empty `paymasterAndData` field.
+ * You can utilize this method before sending the user operation to confirm its eligibility for gas sponsorship. Depending on the outcome, it allows you to tailor the user experience accordingly, based on eligibility.
+ *
+ * @example
+ * ```ts
+ * import { smartAccountClient } from "./smartAccountClient";
```
denniswon, line 20: ditto: why put the doc here?

packages/core/src/actions/smartAccount/types.ts
```diff
@@ -44,15 +44,29 @@ export type SendUserOperationParameters<
...
 
+//#region BuildUserOperationParameters
+export type BuildUserOperationParameters<
+  TAccount extends SmartContractAccount | undefined,
+  TContext extends UserOperationContext | undefined =
+    | UserOperationContext
+    | undefined,
+  TEntryPointVersion extends GetEntryPointFromAccount<TAccount> = GetEntryPointFromAccount<TAccount>
+> = SendUserOperationParameters<TAccount, TContext, TEntryPointVersion>;
+//#endregion BuildUserOperationParameters
```
denniswon, line 55: this is just for readability right?

denniswon (commented):
Overall LGTM. But for multi sig code, would really like a reference doc for the steps/functionalities involved for multisig signing, and even better, more code comments that explains the flow.

packages/core/src/actions/smartAccount/buildUserOperationFromTx.ts
```diff
@@ -19,6 +19,81 @@ import type {
...
+ * @param skipSigning
+ * @param skipSigning
+ * @param skipSigning
+ * @param skipSigning
+ * @param skipSigning
+ * @param skipSigning
+ * @param skipSigning
+ * @param skipSigning
+ * @param skipSigning
+ * @param skipSigning
```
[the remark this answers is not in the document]
moldy530: oh weird... eslint went crazy on this? need to check

packages/core/src/actions/smartAccount/buildUserOperationFromTx.ts
moldy530: I copied from there. I'd rather not link out to to account kit. the goal of the JSDocs is so devs don't have to leave their IDEs for docs

packages/core/src/actions/smartAccount/buildUserOperationFromTxs.ts
```diff
@@ -11,20 +11,94 @@ import type { UserOperationOverrides } from "../../types";
 import { bigIntMax } from "../../utils/index.js";
 import { buildUserOperation } from "./buildUserOperation.js";
 import type {
+  BuildTransactionParameters,
   BuildUserOperationFromTransactionsResult,
-  SendTransactionsParameters,
+  UserOperationContext,
 } from "./types";
 
+/**
```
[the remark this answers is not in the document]
moldy530: see above.

packages/core/src/actions/smartAccount/buildUserOperationFromTx.ts
```diff
@@ -19,6 +19,81 @@ import type {
...
+    from, // ignored
+    to,
+    data: encodeFunctionData({
+      abi: ContractABI.abi,
+      functionName: "func",
+      args: [arg1, arg2, ...],
+    }),
+  },
+];
+const uoStruct = await smartAccountClient.buildUserOperationFromTxs({
```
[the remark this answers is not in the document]
moldy530: oh weird. maybe the docs are broken too cuz I copied from there. lemme double check

packages/core/src/actions/smartAccount/checkGasSponsorshipEligibility.ts
moldy530: see above about JSDocs keeping users in their IDE for ergo

packages/core/src/actions/smartAccount/types.ts
moldy530: yea

dphilipson (approved):

dphilipson (approved):
Looks good to me, thanks for doing this!

PR #632, opened 2024-04-30
fix: determine entry point version from light account version

dphilipson:
## PR-Codex overview
This PR updates the `entryPoint` initialization in the `account.ts` file to include the version from `AccountVersionRegistry`.

### Detailed summary
- Updated `entryPoint` initialization to include version from `AccountVersionRegistry`
- Utilized `AccountVersionRegistry` to fetch version and address information for `LightAccount`

### DIFF
--- a/packages/accounts/src/light-account/accounts/account.ts
+++ b/packages/accounts/src/light-account/accounts/account.ts
@@ -108,7 +108,10 @@ export async function createLightAccount({
   signer,
   initCode,
   version = defaultLightAccountVersion("LightAccount"),
-  entryPoint = getEntryPoint(chain),
+  entryPoint = getEntryPoint(chain, {
+    version: AccountVersionRegistry["LightAccount"][version]
+      .entryPointVersion as any,
+  }),
   accountAddress,
   factoryAddress = AccountVersionRegistry["LightAccount"][version].address[
     chain.id


packages/accounts/src/light-account/accounts/account.ts
```diff
@@ -108,7 +108,10 @@ export async function createLightAccount({
   signer,
   initCode,
   version = defaultLightAccountVersion("LightAccount"),
-  entryPoint = getEntryPoint(chain),
+  entryPoint = getEntryPoint(chain, {
+    version: AccountVersionRegistry["LightAccount"][version]
+      .entryPointVersion as any,
```
dphilipson, line 113: I couldn't think of a way to do this without the `any` cast. TypeScript can't infer that this type matches the type for the light account version in `version`.

moldy530: is `GetLightAccountVersion` generic on entrypoint version? it could take in the `TEntryPointVersion` which might strongly type them

moldy530: oh wait:
```
TEntryPointVersion extends GetEntryPointForLightAccountVersion<
    "LightAccount",
    TLightAccountVersion
  > = GetEntryPointForLightAccountVersion<"LightAccount", TLightAccountVersion>
```

weird. I'd expect this to allow the inference

moldy530 (approved):
approving to unblock, might need to revisit this and see why the version can't be tied to the ep version.

I ran into some other typing weirdness with the `getEntryPoint` method so there's def some edge cases to resolve

packages/accounts/src/light-account/accounts/account.ts
dphilipson: Yeah, I'm not quite sure why either, but the error message is a bit too complicated to easily decipher 😞.

PR #649, opened 2024-05-06
feat(ui): start adding use authenticate calls to the modal

moldy530:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on adding new dependencies, updating UI components, and refactoring file structure in the Alchemy project.

### Detailed summary
- Added new dependencies like `@tanstack/react-form` and `@tanstack/zod-form-adapter`
- Updated UI components like `AlchemyLogo`, `ChevronRight`, `MailIcon`
- Refactored file structure for better organization

> The following files were skipped due to too many changes: `packages/alchemy/src/react/icons/hourglass.tsx`, `packages/alchemy/src/tailwind/components/buttons.ts`, `examples/ui-demo/env.mjs`, `packages/alchemy/src/react/components/auth/hooks/usePasskeyVerify.ts`, `packages/alchemy/src/react/components/AuthModal/AuthSection.tsx`, `packages/alchemy/src/react/hooks/useAuthenticate.ts`, `packages/alchemy/src/react/components/auth/card/main.tsx`, `examples/ui-demo/src/app/providers.tsx`, `packages/alchemy/src/react/components/button.tsx`, `examples/ui-demo/src/app/page.tsx`, `packages/alchemy/src/react/components/auth/card/add-passkey.tsx`, `packages/alchemy/src/react/components/auth/card/loading/passkey.tsx`, `packages/alchemy/src/react/components/auth/card/index.tsx`, `packages/alchemy/src/react/icons/passkey.tsx`, `packages/alchemy/src/react/components/auth/sections/EmailAuth.tsx`, `packages/alchemy/src/react/components/auth/card/loading/email.tsx`, `yarn.lock`, `packages/alchemy/src/react/context.tsx`


packages/alchemy/src/react/components/auth/modal/useAuthModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+          <AuthCard header={header} sections={sections} />
+          <div
+            className="modal-backdrop"
+            onClick={() => closeAuthModal()}
+          ></div>
+        </dialog>
+      );
+    },
+    [header, sections]
+  );
```
dphilipson, line 37: We _may_ want to write this so that `AuthModal` is the same instance even if `header` or `sections` change. As written, if either of those change, then `AuthModal` will be a different function instance which will trigger a remount. I can especially see this happening with `sections`, where someone might end up constructing new array instances to pass in on every render.

One way we could write this:
```tsx
const headerRef = useLatestRef(header);
const sectionsRef = useLatestRef(sections);

// Ensure same instance on every render.
const AuthModal = useRef(
  ({ className }: ModalProps) => {
    // ...
    <AuthCard header={headerRef.current} sections={sectionsRef.current} />
    // ...
  }
).current;
```
where `useLatestRef` does what it sounds like:
```ts
function useLatestRef<T>(value: T): MutableRefObject<T> {
  const ref = useRef(value);
  ref.current = value;
  return ref;
}
```

dphilipson (commented):
Looks great! One comment for discussion.

packages/alchemy/src/react/components/auth/modal/useAuthModal.tsx
dphilipson: There may be some edge cases where that proposed solution won't work, since changing `header` or `sections` wouldn't trigger a re-render in `AuthModal`. Normally this would be fine since if you're calling `useAuthModal` chances are you're about re-render some JSX that has the `AuthModal` in it, but I could imagine some kinds of application structure where you don't use `AuthModal` immediately and it not re-rendering could be a problem. If you think that's worth worrying about, we probably need some mechanism to send the new values to the `AuthModal` to trigger a re-render.

moldy530: Oh hmmm I actually ran into this not with section and headers since I was expecting these to be mostly static. Even in the tech spec one proposal is to bake this `uiConfig` into the `createConfig` function which is supposed to be a static reference. 

I ran into this with the `input` component. If you have the input focused and the input validator generates an error then the whole input component is re-rendered because the error prop is different. This causes the input to lose focus. For Input I need to come back to handle that better, probably with the way you've proposed.

For AuthModal or similar components in other libs, my expectation would be that changes to those props would result in a re-render. 

One thing I was thinking about is actually NOT returning the `<AuthModal />` component to the dev at all. Instead, I think the hook should inject the modal into the dom (likely using a portal). Then all of these edge cases with re-renders are confined within the hooks implementation and we don't have to make assumptions about external consumers. I think this is how most UI libraries do it normally. What are your thoughts on taking this approach instead?

moldy530: as I think about this more... `useAuthModal` should be a hook that works much like the rest of the `use*` react hooks we have. The config for `useAuthModal` should actually be part of the `createConfig` that means that the `AlchemyAccountContext` will always have the `<AuthModal />` rendered after any children. 

the return of `useAuthModal` then will just return the actions for `show` and `close` modal

moldy530: @dphilipson what do you think about: https://github.com/alchemyplatform/aa-sdk/pull/655? I've not addressed the re-renders that can happen when ui-config changes, but now at least the rendering of the dialog is done in the same component that receives the props which should match expectations better?

dphilipson (dismissed):
With the change in https://github.com/alchemyplatform/aa-sdk/pull/655 looks great!

PR #770, opened 2024-06-20
feat: border radius customization and auth card background stuff

tinaszheng:
This PR adds support for customizing a `rounded-modal` tailwind class and also moves around the background/border radius definitions into the inner components

image
image

# Pull Request Checklist

### DIFF
--- a/account-kit/react/src/tailwind/types.ts
+++ b/account-kit/react/src/tailwind/types.ts
@@ -32,6 +32,9 @@ export interface AccountKitTheme {
     "bg-surface-warning": ColorVariantRecord;
     "bg-surface-error": ColorVariantRecord;
   };
+  borderRadius: {
+    modal: string;
+  };
 }
 
 export type AccountKitThemeColor = keyof AccountKitTheme["colors"];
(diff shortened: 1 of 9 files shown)


account-kit/react/src/tailwind/types.ts
```diff
@@ -32,6 +32,9 @@ export interface AccountKitTheme {
     "bg-surface-warning": ColorVariantRecord;
     "bg-surface-error": ColorVariantRecord;
   };
+  borderRadius: {
+    modal: string;
+  };
```
moldy530, line 37: I wonder if this should just be a union type of the `RADIUS_OPTIONS["id"]` options below. Then in places where we have `rounded-*` within our components we just do `rounded-${akTheme.borderRadius}`. To support the styling via demo app it gets tricky I guess since it needs to be a var, but we can do what we do with the colors:

1. hoist the `akTheme.borderRadius` config into a `var` like `--ak-ui-border-radius`. We could have the translation to pixels based on the enum value
2. change the `rounded-*` definitions in our classes to `rounded-[var(--ak-ui-border-radius)]`
3. in the demo app override the variable using the same translation to pixels?

Open question: are there cases where the border radius is different for the button than for the modal for example? or is the idea that they all have `rounded-*`? If some are different, then what we really need is someway to override the `xs`, `s` etc definitions in the theme.

tinaszheng: ooh good q, i didn't think they could change all the border radiuses but let me ask julia to clarify!

PR #789, opened 2024-06-26
feat: read auth card props from context, simplify UI config

bswags:
A few changes that I think will make the integration *much simpler* for developers, very open to feedback:
- The UI config is now passed as an optional property to a re-exported and wrapped `createConfig`
- The UI config is then made available via a `useUiConfig` hook that reads from the context
- The `AuthCard` now reads most of its props (besides `className`) from the ui config
- There's an internal auth card content that the `AuthCard` and `AuthModal` render directly, to eliminate external props
- Some nesting within the UI config props was removed

In addition to making life easier for integrators, this prevents the auth card and auth modal from getting out of sync

# Pull Request Checklist

---

## PR-Codex overview
This PR updates the UI configuration and refactors components in the account-kit/react project.

### Detailed summary
- Moved `AuthIllustrationStyle` to a separate file
- Updated UI configuration in various components
- Refactored components to use `useUiConfig` hook
- Updated `createConfig` to include UI configuration
- Updated `LoadingAuth`, `PasskeyAdded`, and `LoadingPasskeyAuth` components

> The following files were skipped due to too many changes: `account-kit/react/src/components/auth/modal.tsx`, `account-kit/react/src/components/auth/card/loading/email.tsx`, `examples/ui-demo/src/components/preview/AuthCardWrapper.tsx`, `account-kit/react/src/components/auth/card/add-passkey.tsx`, `examples/ui-demo/src/app/providers.tsx`, `account-kit/react/src/context.tsx`, `account-kit/react/src/components/auth/card/index.tsx`


account-kit/react/src/components/auth/context.ts
```diff
@@ -14,6 +15,7 @@ export type AuthStep =
 type AuthContextType = {
   authStep: AuthStep;
   setAuthStep: (step: AuthStep) => void;
+  uiConfig?: AlchemyAccountsUIConfig;
```
bswags, line 18: @moldy530 I wasn't entirely sure if this was the right place for this, but it seems to make the most sense because it's internal state related to the auth flow/UI, and introducing a new context just for the UI config felt way too overkill

account-kit/react/src/context.tsx
```diff
@@ -38,7 +42,20 @@ export const AlchemyAccountContext = createContext<
 >(undefined);
 
 export type AlchemyAccountsUIConfig = {
-  auth?: AuthCardProps & { addPasskeyOnSignup?: boolean };
+  header?: ReactNode;
```
bswags, line 45: imo it doesn't make sense to nest this within `auth` anymore, since there is no real distinction between the UI and "auth" - it's all auth UI config... if there's a forward-looking reason I don't know about, happy to revert, but it felt weird to be spread across top-level and a nested object (and introduced type weirdness where the top level could be present but not the inner auth object)

account-kit/react/src/components/auth/card/add-passkey.tsx
```diff
@@ -24,7 +24,7 @@ const BENEFITS = [
 ];
 
 // eslint-disable-next-line jsdoc/require-jsdoc
-export const AddPasskey = ({ config }: { config: AuthCardProps }) => {
+export const AddPasskey = ({ config }: { config: AlchemyAccountsUIConfig }) => {
```
moldy530, line 27: nit: any reason not to use this as it comes from `useAuthContext` below?

account-kit/react/src/components/auth/card/index.tsx
```diff
@@ -1,54 +1,55 @@
...
 /**
  * React component containing an Auth view with configured auth methods
+ * and options based on the config passed to the AlchemyAccountProvider
  *
  * @param props Card Props
- * @param props.header optional header for the card (good place to put your app name or logo)
- * @param props.showSignInText optional boolean to show the sign in text (defaults to true)
- * @param props.sections array of sections, each containing an array of auth types
+ * @param props.className Optional class name to apply to the card
  * @returns a react component containing the AuthCard
```
moldy530, line 23: quick nit, can you format the `@param` tags to look like `@param {Type} {name} {description}`. This will make it line up with the doc gen cli expectations

account-kit/react/src/components/auth/card/loading/email.tsx
```diff
@@ -7,10 +7,10 @@ import { useAuthContext, type AuthStep } from "../../context.js";
 import { Spinner } from "../../../../icons/spinner.js";
 import { ls } from "../../../../strings.js";
 import { EmailIllustration } from "../../../../icons/illustrations/email.js";
-import type { AuthCardProps } from "../index.js";
+import type { AlchemyAccountsUIConfig } from "../../../../context.js";
 
 interface LoadingEmailProps {
-  config: AuthCardProps;
+  config: AlchemyAccountsUIConfig;
```
moldy530, line 13: here and below, similar comment as the `AddPasskey` component. We should be able to just pull this form the `useAuthContext()` hook now right? or is there still a step that relies on this?

account-kit/react/src/components/auth/context.ts
moldy530: yea I wonder if instead we should add an export for `createConfig` in `@account-kit/react` which wraps the `createConfig` call from `core` but adds a param for `uiConfig` now. That way it's part of the return type there and you shouldn't need `useAuthContext` to get the `uiConfig` and instead you can just use the `useAccountContext` or w/e it was called (or maybe add a hook called `useUiConfig`)

moldy530: that way you can just use that hook everywhere you need the ui config and you don't have to prop drill it anywhere

account-kit/react/src/context.tsx
moldy530: yea the initial reason for this was to support the case where we might have like sign message pop-ups or something. If you're using `useSendUserOperation` or `useSignMessage` you might get a sign message or tx pop-up if you opt in to that

moldy530 (commented):
refactor generally makes sense to me. my suggestion is to move the `uiConfig` into a `createConfig` param. That means we need to have a separate definition for `createConfig` exported from `@account-kit/react` but it takes `uiConfig` as input and returns it as part of the ClientConfig. Now a developer just needs to pass in that config and it correctly gets added to the `AlchemyAccountContext` this then allows us to have a `useUiConfig()` hook that tries to read that from the context

account-kit/react/src/components/auth/card/index.tsx
```diff
@@ -1,54 +1,55 @@
...
  */
-export const AuthCard = (
-  props: AuthCardProps & { showNavigation?: boolean; showClose?: boolean }
-) => {
-  const { showClose = false, onAuthSuccess, hideError } = props;
+export const AuthCard = (props: AuthCardProps) => {
```
moldy530, line 25: oh nice I had a question about this in https://github.com/alchemyplatform/aa-sdk/pull/792/files#r1655692359, but looks like we're merging these props

account-kit/react/src/components/auth/card/index.tsx
moldy530: bump

account-kit/react/src/types.ts
```diff
@@ -1,4 +1,32 @@
 import type { UseMutationOptions } from "@tanstack/react-query";
+import type { ReactNode } from "react";
+import type { AuthType } from "./components/auth/types.js";
+
+export type AlchemyAccountsUIConfig = {
+  header?: ReactNode;
+  showSignInText?: boolean;
```
moldy530, line 7: these and sections still feel like they should probably be scoped further to just some `authModal` or similarly named prop. when we add the connect wallet button or connect wallet modal, these won't make as much sense as root props

account-kit/react/src/components/auth/card/main.tsx
```diff
@@ -2,15 +2,17 @@ import { Fragment } from "react";
...
-  header = null,
-  showSignInText = true,
-  sections,
-}: AuthCardProps) => {
+export const MainAuthContent = () => {
+  const { auth } = useUiConfig();
+
+  const header = auth?.header ?? null;
+  const sections = auth?.sections;
+  const showSignInText = auth?.showSignInText ?? false;
```
bswags, line 14: @moldy530 I was thinking that as a follow-up, it might be nice if the `useUiConfig()` hook was where we assigned some defaults, so that some of these properties weren't optional and thus up to the caller to assign a default (e.g. for the `illustrationStyle`)

account-kit/react/src/components/auth/card/add-passkey.tsx
```diff
@@ -48,7 +49,7 @@ export const AddPasskey = ({ config }: { config: AuthCardProps }) => {
         {BENEFITS.map(({ title, icon: Icon, description }) => (
           <div key={title} className="flex gap-2">
             <div className="h-5 w-5 flex items-center justify-center">
-              <Icon illustrationStyle={config.illustrationStyle ?? "flat"} />
+              <Icon illustrationStyle={illustrationStyle ?? "flat"} />
```
moldy530, line 52: With the hook, we could move this prop directly into the Icon components which would be kinda nice

account-kit/react/src/components/auth/card/main.tsx
moldy530: Yeah that's a good idea, like you could pass `ui.auth: true` to create config and get a default setup or something and then the UI config hook just handles the defaults

account-kit/react/src/createConfig.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export type AlchemyAccountsConfigWithUI = AlchemyAccountsConfig & {
+  ui?: AlchemyAccountsUIConfig;
+};
+
+/**
+ * Wraps the `createConfig` that is exported from `@aa-sdk/core` to allow passing
+ * an additional argument, the configuration object for the Auth Components UI
+ * (the modal and AuthCard).
+ *
```
moldy530, line 16: nit: if you add an `@example` tag here the docs will also have a usage section added to it

moldy530 (approved):

PR #841, opened 2024-07-18
feat: add injected wallets to the UX

moldy530:
# Pull Request Checklist

This adds a button group class and fixes up the injected wallets config so that we can render them in the UX

---

## PR-Codex overview
The focus of this PR is to enhance authentication configuration in the UI by adding support for passkey authentication and external wallets display options.

### Detailed summary
- Added support for passkey authentication in UI configuration
- Enhanced UI to display external wallets based on configuration
- Updated button styling in UI components
- Improved authentication methods in UI components

### DIFF
--- a/account-kit/react/src/components/auth/sections/EmailAuth.tsx
+++ b/account-kit/react/src/components/auth/sections/EmailAuth.tsx
@@ -5,12 +5,12 @@ import { useAuthenticate } from "../../../hooks/useAuthenticate.js";
 import { useSigner } from "../../../hooks/useSigner.js";
 import { ChevronRight } from "../../../icons/chevron.js";
 import { MailIcon } from "../../../icons/mail.js";
+import { ls } from "../../../strings.js";
 import { Button } from "../../button.js";
 import { IS_SIGNUP_QP } from "../../constants.js";
 import { Input } from "../../input.js";
 import { useAuthContext } from "../context.js";
 import type { AuthType } from "../types.js";
-import { ls } from "../../../strings.js";
 
 type EmailAuthProps = Extract<AuthType, { type: "email" }>;
 
@@ -58,6 +58,7 @@ export const EmailAuth = ({
 
   return (
     <form
+      className="w-full"
       onSubmit={(e) => {
         e.stopPropagation();
         e.preventDefault();
--- a/account-kit/react/src/components/auth/sections/InjectedProvidersAuth.tsx
+++ b/account-kit/react/src/components/auth/sections/InjectedProvidersAuth.tsx
@@ -4,7 +4,6 @@ import { useConnect } from "../../../hooks/useConnect.js";
 import { Button } from "../../button.js";
 import { useAuthContext } from "../context.js";
 
-// eslint-disable-next-line jsdoc/require-jsdoc
 export const InjectedProvidersAuth = () => {
   const { chain } = useChain();
   const { connectors, connect } = useConnect({
@@ -33,7 +32,7 @@ export const InjectedProvidersAuth = () => {
   }
 
   return (
-    <div className="flex flex-row gap-2 flex-wrap">
+    <>
       {connectors.map((connector) => {
         return (
           <Button
@@ -58,6 +57,6 @@ export const InjectedProvidersAuth = () => {
           </Button>
         );
       })}
-    </div>
+    </>
   );
 };
--- a/account-kit/react/src/components/button.tsx
+++ b/account-kit/react/src/components/button.tsx
@@ -1,10 +1,11 @@
 import {
   forwardRef,
+  useLayoutEffect,
+  useState,
   type ButtonHTMLAttributes,
   type DetailedHTMLProps,
   type ReactNode,
 } from "react";
-import { GoogleIcon } from "../icons/google.js";
 
 type ButtonProps = (
   | { variant?: "primary" | "secondary" | "link"; icon?: never }
@@ -20,6 +21,8 @@ type ButtonProps = (
 
 export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
   ({ variant, children, icon, className, ...props }, ref) => {
+    const [localRef, setLocalRef] = useState<HTMLButtonElement | null>(null);
+
     const btnClass = (() => {
       switch (variant) {
         case "secondary":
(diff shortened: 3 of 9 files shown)


moldy530:
image
image

account-kit/react/src/components/auth/card/main.tsx
```diff
@@ -14,7 +14,9 @@ export const MainAuthContent = () => {
   return (
     <>
       {header}
-      {showSignInText && <h3 className="font-semibold text-lg">Sign in</h3>}
+      {(showSignInText == null || showSignInText) && (
```
rthomare, line 17: lol this is slightly conufsing. So if sign in text is true or null we'd want to show signin text? Is there a better default we could use here?

account-kit/react/src/components/auth/sections/EmailAuth.tsx
```diff
@@ -5,12 +5,12 @@ import { useAuthenticate } from "../../../hooks/useAuthenticate.js";
 import { useSigner } from "../../../hooks/useSigner.js";
 import { ChevronRight } from "../../../icons/chevron.js";
 import { MailIcon } from "../../../icons/mail.js";
+import { ls } from "../../../strings.js";
```
rthomare, line 8: where is this used?

account-kit/react/src/components/auth/sections/InjectedProvidersAuth.tsx
```diff
@@ -4,7 +4,6 @@ import { useConnect } from "../../../hooks/useConnect.js";
 import { Button } from "../../button.js";
 import { useAuthContext } from "../context.js";
 
-// eslint-disable-next-line jsdoc/require-jsdoc
```
rthomare, line 7: Does this not apply anymore?

account-kit/react/src/components/button.tsx
```diff
@@ -34,44 +37,43 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
+    useLayoutEffect(() => {
+      if (!localRef) return;
+
+      const parent = localRef.parentElement;
+      if (!parent) return;
+      const siblings = parent.children;
+
+      // this assumes this element is in a button group
+      // in a 3 element button group all buttons have their text shown
+      if (siblings.length <= 3) return;
```
rthomare, line 51: Sorry a little confused what is the thing we are trying to achieve with the UI. To me this feels a little in-direct. If a child view is reaching out to understand parent context that feels like a code smell. Especially in a broad component like button.

account-kit/react/src/components/auth/sections/InjectedProvidersAuth.tsx
moldy530: now that all the exported methods are correctly being parsed, we shouldn't need to disable this lint rule anymore

account-kit/react/src/components/button.tsx
moldy530: I agree generally about not relying on your parents for styling your self, but here this was probably the easiest way to achieve this layout:
image

There's a couple wrinkles:
1. since a button group is max 2 rows, I did this by having a flex wrap + setting the first child at max width and the rest of them having a flex grow and flex shrink. This is done on the `btn-group` class. Because the children are flex grow and shrink and the container is flex wrap, the buttons will shrink only until the text is not cut off. So we can't do something in the button like `if (width < text.width + icon.width) hideText` because that will never be the case and the buttons will overflow into the next row.
2. If we made a button group component, the component itself wouldn't be able to hold this logic because it would just receive children and AFAIK it's not straightforward to check what kind of child it is and pass in new props
3. the button group is just the high level container for each auth section. Almost all of the sections are buttons so that's chill, but you can also have email as a section. The email section is typically rendered in the first section and someone could add other sections along side email. In that case, you would want the remaining sections (which are all buttons) to render as the second row of a button group.

This ultimately felt like the easiest thing to do to match the layout. Buttons aren't actually exported to be consumed externally, so this is all internal to the SDK which feels like an ok place to make these kinds of one offs

account-kit/react/src/components/auth/sections/EmailAuth.tsx
moldy530: In the component itself, this moved because I ran the linter which re-ordered imports

account-kit/react/src/components/button.tsx
dphilipson: Hm, this is pretty tricky. So if I'm understanding correctly, the goal of this code is to apply different styling if the button has more than three siblings (counting itself)?

I'm okay with this solution, but did you look into using tricky CSS selectors to apply based on sibling count? I think it might be possible to do something like
```css
.parent:has(> :nth-child(4)) > * {
    /* ... */
}
```
where that rule would match elements that are a child of a `.parent` element if that `.parent` has at least four children. If that works, it might be a little simpler.

account-kit/react/src/tailwind/components/buttons.ts
```diff
@@ -1,7 +1,17 @@
 import type { ComponentDef } from "../types";
 
 export const buttonComponents: ComponentDef = {
+  ".btn-group": {
```
dphilipson, line 4: For my own understanding, how are these CSS rules localized to the account-kit UI component? I'm assuming this isn't changing global stylings for common classes like `.btn`, is it?

account-kit/react/src/components/button.tsx
moldy530: oh that's cool, lemme try that

account-kit/react/src/tailwind/components/buttons.ts
moldy530: I need to double check, I think they might conflict with global styles actually. I need to double check, but I think we might need to configure our plugin to use a prefix to avoid the clashing

account-kit/react/src/components/button.tsx
moldy530: ok sweet! check it out now? I moved it into css, though I think the smell still exists here because it does require specific set up between btn and btn groups, but I think this is ok. At somepoint, we will need to support the case where there are more than 4 buttons in the second row. What we will have to do there is: if there are 4 buttons, they get condensed into a `...` button that opens some other view containing the overflow. To handle that we'll have to use a `ref` that checks either sibling counts or children, so we might not be out of the woods fully. That said I like the css solution more because it keeps all of this btn and btn group styling in the same place, so it's a bit more discoverable

account-kit/react/src/tailwind/components/buttons.ts
moldy530: this probably is a problem, but will have to deal with it in a separate PR

account-kit/react/src/components/button.tsx
dphilipson: That sort of thing may still be possible with CSS if the "..." button is always in the DOM but hidden by CSS unless there are more than four buttons in the second row. But we can cross that bridge when we come to it. The new version looks good!

dphilipson (dismissed):
Looks good with the latest change, but presumably needs some kind of rebase. The changeset is so large that GitHub’s UI is lagging as I type this.

rthomare:
Also LGTM but let me know if you need a review after the rebase

PR #942, opened 2024-08-27
feat: add support for social login

dphilipson:
# Pull Request Checklist

---

## PR-Codex overview
This PR introduces OAuth enhancements and popup functionality to the Alchemy Signer, including new methods and type assertions.

### Detailed summary
- Added `oauth` and `oauthReturn` types to `OauthParams`
- Implemented `assertNever` function for type checking
- Introduced `preparePopupOauth` method for popup OAuth login
- Updated session handling for OAuth connections

> The following files were skipped due to too many changes: `account-kit/signer/src/session/manager.ts`, `account-kit/signer/src/client/base.ts`, `account-kit/signer/src/base.ts`, `account-kit/signer/src/client/index.ts`


account-kit/signer/src/base.ts
```diff
@@ -604,6 +641,30 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
     return user;
   };
 
+  private authenticateWithOauth = async (
+    args: Extract<AuthParams, { type: "oauth" }>
+  ): Promise<User> => {
+    const params: OauthParams = {
+      ...args,
+      expirationSeconds: this.sessionManager.expirationTimeMs,
```
moldy530, line 649: does this need to be 
```suggestion
      expirationSeconds: Math.floor(this.sessionManager.expirationTimeMs / 1000),
```

account-kit/signer/src/client/index.ts
```diff
@@ -205,19 +229,21 @@ export class AlchemySignerWebClient extends BaseSignerClient<ExportWalletParams>
...
     orgId,
+    connectedEventName,
   }: {
     bundle: string;
     orgId: string;
-  }) => {
+    connectedEventName: keyof AlchemySignerClientEvents;
+  }): Promise<User> => {
     this.eventEmitter.emit("authenticating");
     await this.initIframeStamper();
```
moldy530, line 248: maybe we should rename this to `completeBundleAuth` or something since passkey auth doesn't use the iframe stamper? I have a branch somewhere that tries to add passkey sessions, but I hit a roadblock in the signer-service which needed a change to the chain configs so the branch is private.

This method is great for that branch so thanks for doing this refactor!

account-kit/signer/src/client/index.ts
```diff
@@ -423,4 +573,35 @@ export class AlchemySignerWebClient extends BaseSignerClient<ExportWalletParams>
...
+    } else {
+      throw new Error(
+        "initOauth must be called before using popup-based OAuth login"
+      );
+    }
+  };
+}
+
+function resolveRelativeUrl(url: string): string {
+  // Funny trick.
```
moldy530, line 599: neat

account-kit/signer/src/client/index.ts
```diff
@@ -423,4 +573,35 @@ export class AlchemySignerWebClient extends BaseSignerClient<ExportWalletParams>
...
+  const a = document.createElement("a");
+  a.href = url;
+  return a.href;
+}
+
+function base64urlEncode(s: string): string {
```
moldy530, line 605: nit: we already have this in `utils/base64UrlEncode.ts`

account-kit/core/src/types.ts
```diff
@@ -103,6 +103,15 @@ export type CreateConfigProps = RpcConnectionConfig & {
...
   connectors?: CreateConnectorFn[];
+
+  /**
+   * If set, calls `preparePopupOauth` immediately upon initializing the signer.
+   * If you intend to use popup-based OAuth login, you must either set this
+   * option to true or manually ensure that you call
+   * `signer.preparePopupOauth()` at some point before the user interaction that
+   * triggers the OAuth authentication flow.
+   */
+  preparePopupOauthOnStartup?: boolean;
```
moldy530, line 114: nit: should we call this something like `usePopupOauth` or `usePopup` or even just `popup`, so we  don't leak impl details to devs?

account-kit/react/src/hooks/useAuthenticate.ts
```diff
@@ -67,6 +77,7 @@ export function useAuthenticate(
   );
 
   return {
+    preparePopupOauth,
```
moldy530, line 80: this is here in case a developer wants to manually prepare the popup? doesn't the config handle this for you now?

account-kit/signer/src/base.ts
```diff
@@ -170,11 +201,16 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
...
-      return this.authenticateWithEmail(params);
+    switch (params.type) {
+      case "email":
+        return this.authenticateWithEmail(params);
+      case "passkey":
+        return this.authenticateWithPasskey(params);
+      case "oauth":
+        return this.authenticateWithOauth(params);
+      case "oauthReturn":
+        return this.handleOauthReturn(params);
```
moldy530, line 212: nit: should we add a `default` which throws? Actually maybe not because I remember something similar in Rust where we only handle the cases that exist so that the compiler throws an error when something isn't handled right?

account-kit/signer/src/client/base.ts
```diff
@@ -57,6 +59,15 @@ export abstract class BaseSignerClient<TExportWalletParams = unknown> {
     );
   }
 
+  /**
+   * Asynchronously fetches and sets the OAuth configuration.
+   *
+   * @returns {Promise<OauthConfig>} A promise that resolves to the OAuth configuration
+   */ public initOauth = async (): Promise<OauthConfig> => {
```
moldy530, line 66: sometimes the linter does this and I haven't tracked down why

```suggestion
   */ 
   public initOauth = async (): Promise<OauthConfig> => {
```

account-kit/signer/src/client/base.ts
```diff
@@ -309,7 +331,7 @@ export abstract class BaseSignerClient<TExportWalletParams = unknown> {
     route: R,
     body: SignerBody<R>
   ): Promise<SignerResponse<R>> => {
-    const url = this.connectionConfig.rpcUrl ?? "https://api.g.alchemy.com";
+    const url = "http://api.g.localtest.me:8081";
```
moldy530, line 334: was this left over from local testing?

account-kit/signer/src/base.ts
dphilipson: Yea, when it's written like this, if a new option is added for `params.type` or if we forgot to handle one of them, then this will become a compiler error until we add handling for the new type.

Something I've seen done before that we might consider, which maintains this typechecking but also throws a descriptive runtime error, is to define an `assertNever` utility function. It would look something like
```ts
function assertNever(x: never, makeMessage: (x: unknown) => string): never {
  throw new Error(makeMessage(x as unknown));
}
```
and used
```ts
switch (params.type) {
  case "email":
    // ...
  default:
    assertNever(params.type, (x) => `Unknown auth type: ${x}`);
}
```
You don't need that `makeMessage` as a callback, you could have it just take a string, but then it's annoying because the typesystem doesn't like you interpolating a `never` value into a string so you have to cast it to `any` or `unknown`.

account-kit/signer/src/client/base.ts
dphilipson: Oh yeah it was, thank you.

account-kit/signer/src/base.ts
dphilipson: Added `assertNever` in a bunch of places, should be nice.

account-kit/core/src/types.ts
dphilipson: I think you're right, I'll keep it simple. If an advanced dev wants to customize this behavior, they can `useSigner` and call `preparePopupOauth` on the signer, but we don't need to go out of our way to expose this.

account-kit/signer/src/base.ts
moldy530: nice, yea I like this a lot

package.json
```diff
@@ -86,5 +86,6 @@
     "**/src/**/*.{tsx,jsx,ts,js}": [
       "yarn lint:write"
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
```
moldy530, line 90: I wonder if this is what's causing the breaks in vercel?

moldy530 (approved):
LGTM if the vercel apps deploy successfully

package.json
dphilipson: Ohh, yea, whoops, let me take that out. My corepack config adds that automatically, and I usually remember not to commit it but not this time.

PR #963, opened 2024-09-13
fix: light account 1271 signing

adamegyed:
The EIP-712 Domain has been incorrect for both versions of LightAccount. This PR fixes it, and updates the tests to actually verify the signatures rather than just string-matching on the output.

# Pull Request Checklist

---

## PR-Codex overview
The focus of this PR is to update the `signWith1271Wrapper` function in the `base.ts` file and add support for version-based signing methods. 

### Detailed summary
- Updated `signWith1271Wrapper` function in `base.ts` to accept a `version` parameter.
- Modified signing methods based on the provided version.
- Added support for signing with different versions in `createLightAccountBase` function.
- Updated client setup and message verification in test files.

> The following files were skipped due to too many changes: `account-kit/smart-contracts/src/light-account/clients/client.test.ts`


account-kit/smart-contracts/src/light-account/clients/client.test.ts
```diff
@@ -65,32 +65,37 @@ describe("Light Account Tests", () => {
...
       case "v2.0.0":
-        expect(await account.signMessage({ message })).toBe(
-          "0x00823ee659d8f86abaee06bb36c4d84bfefa2938af2835ff62564e9559bc671d0c0e8603a5aa3ed9603b28a137eaf491b64c55ed73007c12efebcfd9a82cb181221b"
-        );
+        {
+          const signature = await account.signMessageWith6492({
+            message,
+          });
+
+          // We must use a public client, rather than an account client, to verify the message, because AA-SDK incorrectly attaches the account address as a "from" field to all actions taken by that client, including the `eth_call` used internally by viem's signature verifier logic. Per EIP-684, contract creation reverts on non-zero nonce, and the `eth_call`'s from field implicitly increases the nonce of the account contract, causing the contract creation to revert.
```
Zer0dot, line 90: Is this a known thing that we're working on fixing?

Zer0dot (approved):
Thanks for this. Small question, there are unchanged files which have failing assertions-- is that relevant?

Regardless, think it's worth pushing this change, it's a pretty big feature.

account-kit/smart-contracts/src/light-account/clients/client.test.ts
adamegyed: TBD from the clients team. We've raised this as an issue in the past, but I'm not sure what the plan for this is.

adamegyed:
> there are unchanged files which have failing assertions

Looks like there were a few other places in the codebase just string-matching on signatures. Currently working on updating these, ideally we can fix all of the test cases.

Zer0dot (dismissed):

aa-sdk/ethers/src/__tests__/provider-adapter.test.ts
```diff
@@ -19,13 +26,26 @@ describe("Simple Account Tests", async () => {
 
   it("should correctly sign the message", async () => {
     const provider = await givenConnectedProvider({ alchemyProvider, signer });
+    const message = "hello world";
+    // todo: why does this provider's version of signMessage return a string, instead of Hex?
```
dphilipson, line 30: Seems to be because this provider is an ethers provider rather than viem because we're in the ethers unit tests, and ethers doesn't distinguish hex strings.

aa-sdk/ethers/src/__tests__/provider-adapter.test.ts
```diff
@@ -19,13 +26,26 @@ describe("Simple Account Tests", async () => {
...
+    });
+
     expect(
-      await provider.signMessage(
-        "0xa70d0af2ebb03a44dcd0714a8724f622e3ab876d0aa312f0ee04823285d6fb1b"
-      )
-    ).toBe(
-      "0x007ecc361d63ab82d89faeecfb79d40127f376c1ef3d545aeec3578eadce9d0c405a4d1ae6177bdebdc8413065014f735ee98da9643cc0e25c07a7423b694f8ae71b"
-    );
+      await publicClient.verifyMessage({
```
dphilipson, line 40: I wrote this comment before realizing that the codebase is already doing things the way you did everywhere, so no need to change. But since I already wrote the comment, you get it anyways as a free FYI.

Nit: it's more idiomatic to write this assertion as
```ts
await expect(publicClient.verifyMessage({ ... })).resolves.toBe(true)
```
That way gives a clearer error message in the case where the assertion fails.

Likewise throughout.

dphilipson (approved):

aa-sdk/ethers/src/__tests__/provider-adapter.test.ts
adamegyed: makes sense, removed comment.

dphilipson (approved):

account-kit/smart-contracts/src/light-account/clients/client.test.ts
moldy530: this is a bug in viem AFAIK, we don't do anything special in the SDK that causes this to be appended

PR #981, opened 2024-09-24
feat: add auth0connection param

dphilipson:
Allow users to set the `connection` query param when using Auth0, which allows them to link directly to one particular auth method in Auth0 rather than a selection screen.


rthomare (dismissed):

account-kit/signer/src/signer.ts
```diff
@@ -27,19 +27,30 @@ export type AuthParams =
...
-    } & RedirectConfig)
+    } & OauthProviderConfig &
+      OauthRedirectConfig)
   | { type: "oauthReturn"; bundle: string; orgId: string };
 
-export type OauthMode = "redirect" | "popup";
+export type OauthProviderConfig =
+  | {
+      authProviderId: "auth0";
+      isCustomProvider?: false;
```
moldy530, line 39: isn't this always true for `auth0`?

account-kit/signer/src/signer.ts
```diff
@@ -27,19 +27,30 @@ export type AuthParams =
...
+      auth0Connection?: string;
+    }
+  | {
+      authProviderId: string;
```
moldy530, line 43: should this be a union type of the providers we support? like the `keyof DEFAULT_SCOPE_AND_CLAIMS` above?

account-kit/signer/src/signer.ts
dphilipson: No, custom providers are a feature we haven't yet implemented in the dashboard where the developer brings their own OIDC provider that's not known by Alchemy. So for example if Petco wanted to add a "Log in with your Petco account" button.

account-kit/signer/src/signer.ts
dphilipson: I believe providers here can be arbitrary ids that someone has set up in their Auth0 dashboard. E.g. if someone's Auth0 has methods for SMS, Okta, Petco account, each of those would be a different id and we can't necessarily know them ahead of time.

I could maybe do some TypeScripty thing so it will have autocomplete for common ids like "github" but we do need to accept arbitrary strings here.

account-kit/signer/src/signer.ts
moldy530: ahh that makes sense, wasn't sure what a `customProvider` was here

account-kit/signer/src/signer.ts
moldy530: ah gotcha, so this is specifically for `auth0` or does it apply to other auth providers like 1st party google and fb?

dphilipson: Oh wait sorry, I thought you were talking about the `auth0connection` param, my bad.

This provider id at the moment will always be one of "google", "facebook", "apple", "auth0", but once we add custom auth providers it can be anything set from the user. Actually, I could make it so if `isCustomProvider` is false it has to be one of the known ones, that's probably a good idea.

PR #991, opened 2024-09-27
feat: toast

RobChangCA:
# Pull Request Checklist

# Description
Adds a toast component to display a success / error when transferring and NFT.
Screenshot 2024-09-27 at 6 04 12 PM

# Testing steps
Attempt to transfer an NFT, on success you should see a green toast appear on top of the white nav bar.
You can click the X next to the toast to dismiss it early.  It will naturally dismiss after 4 seconds.
This toast is not themeable as the nav bar is always white.

---

## PR-Codex overview
This PR introduces a `Toast` notification system using `@radix-ui/react-toast`. It adds a `ToastProvider`, updates the UI components to utilize the toast functionality, and implements animations for toast visibility.

### Detailed summary
- Added `@radix-ui/react-toast` dependency.
- Created `ToastProvider` in `ToastProvider.tsx`.
- Implemented `useToast` hook in `useToast.tsx`.
- Added `Toast` component for displaying notifications.
- Integrated toast notifications in `MintCard` for success/error messages.
- Added toast animations in `globals.css`.
- Created `XIcon` component for closing the toast.

### DIFF
--- a/examples/ui-demo/src/components/shared/MintCard.tsx
+++ b/examples/ui-demo/src/components/shared/MintCard.tsx
@@ -15,6 +15,7 @@ import { AccountKitNftMinterABI, nftContractAddress } from "@/utils/config";
 import { encodeFunctionData } from "viem";
 import { useConfig } from "@/app/state";
 import { useQuery } from "@tanstack/react-query";
+import { Toast } from "./Toast";
 
 type NFTLoadingState = "loading" | "success";
 
@@ -32,8 +33,11 @@ type MintStatus = {
 
 export const MintCard = () => {
   const [status, setStatus] = useState<MintStatus>(initialState);
-  // To be wired into the toast pr
-  const [hasError, setHasError] = useState(false);
+  const [toastContent, setToastContent] = useState<{
+    type: "success" | "error";
+    text: string;
+  } | null>(null);
+
   const { nftTransfered, setNFTTransfered } = useConfig();
 
   const handleSuccess = () => {
@@ -42,11 +46,18 @@ export const MintCard = () => {
       gas: "success",
       signing: "success",
     }));
+    setToastContent({
+      type: "success",
+      text: "You've successfully collected your NFT! Refresh to mint again.",
+    });
     setNFTTransfered(true);
   };
   const handleError = () => {
     setStatus(initialState);
-    setHasError(true);
+    setToastContent({
+      type: "error",
+      text: "There was a problem with that action",
+    });
   };
 
   const { client } = useSmartAccountClient({ type: "LightAccount" });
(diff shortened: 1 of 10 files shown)


examples/ui-demo/src/components/shared/MintCard.tsx
```diff
@@ -96,13 +107,19 @@ export const MintCard = () => {
...
       console.log("uri", uri);
       return uri;
-
     },
     enabled: !!client && !!client?.readContract,
   });
 
   return (
     <div className="flex bg-bg-surface-default radius-1 border-btn-secondary border-2 overflow-hidden h-[532px]">
+      <Toast
```
rthomare, line 116: Do we plan on using this toast in multiple places. In the way it is currently implemented every component needs to instantiate and open + close their own toast, and handle the logic. 

Can we do better here? Could we wrap this all up in a toast provider component that exposes a useOpenToast function that takes in some parameters simply open and then auto close the toast globally? 

That way when we want to use it for errors, or something else, it's a bit easier for downstream components?

moldy530: looks like this is using a ToastProvider already, but it is weird that we can't just render a global Toast component and show a `Toast` via some kind of `showToast(...)` method. 

Maybe we could do something like this? https://www.radix-ui.com/primitives/docs/components/toast#imperative-api

the `useImperativeHandle` hook which contains a `publish` method is interesting here. I think we could use that to be able to:
1. render a single Toast instance in the root of the app
2. add to the context of the ui-demo app a ref or wrapper around a ref to the toast and export a method like `showToast({ content, type })`

Then we can have a `useToast` hook or something that lets us access that `showToast` method

RobChangCA: The intention was to have an explicit interface to reduce abstraction, I will refactor to use a hook implementation.

examples/ui-demo/src/hooks/useToast.tsx
```diff
@@ -0,0 +1,9 @@
+"use client";
+
+import { ToastContext } from "@/contexts/ToastProvider";
+import { useContext } from "react";
+
+export const useToast = () => {
+  const { toast, setToast, closeToast } = useContext(ToastContext);
```
rthomare, line 7: Can we throw an error if the context comes back as undefined from `useContext`?

Something like 
```
if (!context) { 
   throw Error("useToast can only be used by components that are descendants of the ToastProvider`");
}
```

rthomare: And then you can just return context no need to break it down and then send it a new object.

examples/ui-demo/src/components/shared/Toast.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <span
+            className={`${
+              type === "success" ? "bg-[#16A34A]" : "bg-[#DC2626]"
+            } align-middle px-2 py-1 text-fg-invert text-xs font-semibold rounded mr-2`}
+          >
+            {type === "success" ? "Success" : "Error"}
+          </span>
+          {text}
+        </p>
+        <Close>
```
rthomare, line 64: Should this be hooked up to an action for closing the toast?

examples/ui-demo/src/components/shared/Toast.tsx
```diff
@@ -0,0 +1,71 @@
...
+  return (
+    <>
+      <Root
+        id="toast"
+        open={open}
+        onOpenChange={closeToast}
+        className={`${getBGColor()} align-middle rounded-lg shadow-lg px-3 py-2 flex justify-center items-center`}
+      >
+        <p className={`${getTextColor()} text-center pr-2 align-middle`}>
+          <span
```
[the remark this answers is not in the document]
RobChangCA: Great catch, missed that.  Will hard code the value

rthomare (approved):

PR #997, opened 2024-09-30
refactor: use alchemy transport in createConfig

moldy530:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on updating the configuration and transport mechanism for the `Alchemy` integration within the `account-kit` library. It replaces the previous API key method with a transport object, enhancing the way connections are managed.

### Detailed summary
- Added `localStorage.clear()` in `setupTests.ts` before each test.
- Updated `createConfig` to use `transport: alchemy({ apiKey: "your-api-key" })` instead of `apiKey`.
- Modified various components and tests to accommodate the new transport structure.
- Introduced `storeReviver` and `storeReplacer` for handling serialization of `transport`.
- Updated `vite.config.ts` to include `setupTests.ts`.
- Adjusted imports in multiple files to include `alchemy`.
- Enhanced the `watchBundlerClient` and `getBundlerClient` actions to utilize the new transport.
- Revised the test cases to ensure they validate the new transport mechanism correctly.
- Updated documentation to reflect changes in the transport handling.


account-kit/core/src/store/types.ts
```diff
@@ -95,11 +92,10 @@ export type StoreState = {
       >;
     }>;
   };
+  bundlerClient: ClientWithAlchemyMethods;
   // serializable state
   // NOTE: in some cases this can be serialized to cookie storage
   // be mindful of how big this gets. cookie limit 4KB
-  transport: AlchemyTransport;
```
moldy530, line 101: transport has been removed because it's not serializable in NextJS (makes sense it's a function). so instead we store connections as `transport.config` so that we can serialize it and create a transport when we need it. The `bundlerClient` is still fine to pass along it seems though

site/pages/reference/account-kit/core/functions/createConfig.mdx
```diff
@@ -27,7 +27,7 @@ import { sepolia } from "@account-kit/infra";
 
 const config = createConfig({
   chain: sepolia,
-  apiKey: "your-api-key",
```
moldy530, line 30: the basic usage of this for a devs doesn't change much, just requires passing in a `transport` for a given chain.

For multi chain, that's also been made better where you can provide one global `transport` or you can pass a `chains` array which contains `transports` for each chain

account-kit/core/src/store/types.ts
rthomare: Is there any reason to serialize the whole bundler client? Does it actually get recovered correctly from deserialization? I'm actually not aware of the state and stuff inside of it, but if there is it might make sense to be explicit about it?

rthomare: Ignore me saw the test above.

rthomare: Do you have concerns on us accidentally adding something to the bundler client that we need that will make this unserializable? I guess we have the E2E test and can block future updates and revisit then.

rthomare (approved):

account-kit/core/src/store/types.ts
moldy530: the bundler client itself isn't being serialized anymore to storage, BUT i am concerned about the nextjs thing that I ran into. I need to figure out how best to test that the `config` returned from `createConfig` is serializable in the way that next js expects it to be. 

I'm gonna dig around a little bit and see if I can write a test that covers that (maybe nextjs exposes their serialization/deserialization methods and we can write a test explicitly for that so this doesn't become an issue in the future)

moldy530: update: the issue wasn't on the actual body of the config, but it's in the result of the deserialized cookie returned from `cookieToInitialState`

it means that I CAN have transport in the config, I just can't deserialize it to a function... I'll have to come back to that later

PR #1013, opened 2024-10-03
feat: add logger schemas to infra, react, core

moldy530:
# Pull Request Checklist

---

## PR-Codex overview
This PR primarily focuses on updating the `@account-kit/logging` package to version `4.0.0-beta.9`, enhancing logging functionality across various components, and introducing performance tracking for key methods.

### Detailed summary
- Updated `VERSION` in `version.ts` to `4.0.0-beta.9`.
- Added `@account-kit/logging` dependency in multiple `package.json` files.
- Enhanced logging with `InnerLogger` and `ReactLogger`.
- Introduced performance tracking in several methods using `profiled`.
- Implemented `fetchRemoteWriteKey` to dynamically fetch the write key.
- Updated logging events for user authentication and operations.
- Refactored `createClientLogger` to handle remote write keys.
- Modified `BaseAlchemySigner` methods to include logging for authentication and transactions.


account-kit/logging/src/index.ts
```diff
@@ -1,15 +1,54 @@
...
-  return createClientLogger();
+export function createLogger(context: LoggerContext): EventLogger {
+  const innerLogger =
+    typeof window === "undefined"
+      ? createServerLogger(context)
+      : createClientLogger(context);
+
+  const logger: EventLogger = {
+    ...innerLogger,
+    profiled<TFunc extends (...args: any[]) => Promise<unknown> | unknown>(
```
moldy530, line 19: I wanna log performance metrics for certain methods and this makes it super easy to wrap existing functions with profiling

account-kit/react/src/hooks/useSendUserOperation.ts
```diff
@@ -204,8 +205,14 @@ export function useSendUserOperation<
...
-    sendUserOperation,
-    sendUserOperationAsync,
+    sendUserOperation: ReactLogger.profiled(
+      "sendUserOperation",
+      sendUserOperation
+    ),
+    sendUserOperationAsync: ReactLogger.profiled(
+      "sendUserOperationAsync",
+      sendUserOperationAsync
+    ),
```
moldy530, line 215: example of how the `profiled` method works

account-kit/core/src/metrics.ts
```diff
@@ -0,0 +1,40 @@
...
+          authType: "email" | "passkey_anon" | "passkey_email";
+        }
+      | { authType: "oauth"; provider: string };
+  },
+  {
+    EventName: "signer_sign_message";
+  },
+  {
+    EventName: "client_send_uo";
+  },
```
moldy530, line 27: I don't think we need this here actually if it's being added to the infra package

rthomare: Agreed it'll cause dupes and confusion down stream when we go to measure this. Also like the approach of having a single place to define the event schema.

account-kit/core/src/metrics.ts
```diff
@@ -0,0 +1,40 @@
...
+import { VERSION } from "./version.js";
+
+export type ConfigCreatedData = {
+  ssr: boolean;
+  chainIds: number[];
+};
+
+export type CoreEventsSchema = [
+  {
+    EventName: "config_created";
```
rthomare, line 11: is config created when we have a session id? Should we have an upstream event that is like page load?

account-kit/react/src/metrics.ts
```diff
@@ -0,0 +1,24 @@
+import { createLogger } from "@account-kit/logging";
+import type { AlchemyAccountsUIConfig } from "./types.js";
+import { VERSION } from "./version.js";
+
+export type ReactEventsSchema = readonly [
+  {
```
rthomare, line 6: is config_created a dupe of the one in core. I'd recommend we try to reduce the amount of duplicate events as much as possible. It'll make our lives easier when we analyze.

account-kit/core/src/metrics.ts
moldy530: Yea a session ID (anon ID) in the logger, is available with all event because it get set during initialization of the logger

account-kit/react/src/metrics.ts
moldy530: This one isn't a dupe exactly, the core config one only includes the core configuration details whereas this one includes the UI config settings. The number of `config_created` where packageName = `react` will be a subset of the total `config_created` events with name = "core"

account-kit/core/src/metrics.ts
moldy530: > Also like the approach of having a single place to define the event schema.

Per package right? I went with per package but central location because it helps keep a separation of concerns between packages while still maintaining high discoverability for the metric names

account-kit/react/src/metrics.ts
moldy530: I didn't wanna merge the events mainly because I didn't want to leak the `core` impl into `react` because of metrics

account-kit/core/src/metrics.ts
moldy530: wait... just realize the anonid is per page load. lemme write it to session storage too

account-kit/core/src/metrics.ts
rthomare: Yeah per package haha

account-kit/react/src/metrics.ts
rthomare: Gotcha ok we can keep it if they are distinct different events. But wouldn't they share an anonid so we'd be able to map it correctly? Anyway I'm not gonna block this we can always iterate to something better.

moldy530: Yea they will share anonId so we can merge these events on the backend

account-kit/logging/src/index.ts
```diff
@@ -1,15 +1,54 @@
...
+      name: string,
+      func: TFunc
+    ): TFunc {
+      return new Proxy(func, {
+        apply(target, thisArg, argArray) {
+          const start = Date.now();
+          const result = target.apply(thisArg, argArray);
+          if (result instanceof Promise) {
+            return result.then((res) => {
```
dphilipson, line 28: Are you interested in logging in the case where the promise rejects?

account-kit/logging/src/index.ts
```diff
@@ -1,15 +1,54 @@
...
+      ? createServerLogger(context)
+      : createClientLogger(context);
+
+  const logger: EventLogger = {
+    ...innerLogger,
+    profiled<TFunc extends (...args: any[]) => Promise<unknown> | unknown>(
+      name: string,
+      func: TFunc
+    ): TFunc {
+      return new Proxy(func, {
```
dphilipson, line 23: This is fine, but the proxy feels unnecessary here. I think you could achieve the same thing by doing
```ts
return function(...args) {
  const start = Date.now();
  const result = func.apply(this, args);
  // ...
}
```

dphilipson (approved):
Looks great, some inline comments.

account-kit/logging/src/index.ts
moldy530: Yea I was thinking about that and it's in the spec as an open question. I'm wondering if error timings are helpful at all or if that just ends up adding noise?

account-kit/logging/src/index.ts
moldy530: ah yea, so that's exactly how I had it before and typescript wasn't happy about the typing of that lemme see if I can get it to work without the proxy

dphilipson: The proxy is effectively just casting a bunch of stuff to `any` so if you had to you could do that with the nonproxy version and it wouldn't be any worse. But you might be able to do some clever typing to make it work, I dunno.

dphilipson: I tried playing with it, one thing that may work would be to declare the types like this:
```ts
function profiled<TArgs extends any[], TRet>(func: (...args: TArgs) => TRet): (...args: TArgs) => TRet {
  return function (...args: TArgs): TRet {
    return func.apply(this, args);
  }
}
```
i.e. have separate type parameters for the parameters and return type. I'm not sure why it has trouble with just one param for the function type: it seems that TypeScript can't recognize that the types `TFunc` and ```ts
(...args: Parameters<TFunc>) => ReturnType<TFunc>
```
are the same.

Comment (dphilipson) [reply account-kit/logging/src/index.ts:23]: Oh you know, actually thinking about it more that makes sense. `TFunc`, despite extending a function, doesn't have to be _just_ a function: it could have other fields on it too. So returning a plain function doesn't necessarily make it the same type as `TFunc` because we'd be missing those extra fields.

Writing it in terms of parameters and return type makes it explicit that what you're getting back is _only_ the function piece of what was passed in, which practically speaking is all we need.

File: account-kit/logging/src/index.ts (hunk 6 above)
Comment (dphilipson) [reply account-kit/logging/src/index.ts:28]: Yeah, maybe it would just make the data confusing if it were mixed in with the other timings.

File: account-kit/logging/src/index.ts (hunk 7 above)
Comment (moldy530) [reply account-kit/logging/src/index.ts:23]: lemme try this! thanks for looking into it

Comment (moldy530) [reply account-kit/logging/src/index.ts:23]: Ok so single TFunc and using `Parameters<TFunc>` with `ReturnType<TFunc>` didn't work as typescript says they could be instantiated with different things.

what did work was making the generics on the `TArgs` and `TRet`, but I do still need to typecast in the promise case:

```
profiled<TArgs extends any[], TRet>(
      name: string,
      func: (...args: TArgs) => TRet
    ): (...args: TArgs) => TRet {
      return function (...args: TArgs): TRet {
        const start = Date.now();
        const result = func(...args);
        if (result instanceof Promise) {
          return result.then((res) => {
            innerLogger.trackEvent({
              name: "performance",
              data: {
                executionTimeMs: Date.now() - start,
                functionName: name,
              },
            });
            return res;
          }) as TRet; // Typecast is required here
        }

        innerLogger.trackEvent({
          name: "performance",
          data: {
            executionTimeMs: Date.now() - start,
            functionName: name,
          },
        });
        return result;
      };
    },
```

File: account-kit/logging/src/index.ts [hunk 8]
```diff
@@ -1,15 +1,53 @@
...
+
+  const logger: EventLogger = {
+    ...innerLogger,
+    profiled<TArgs extends any[], TRet>(
+      name: string,
+      func: (...args: TArgs) => TRet
+    ): (...args: TArgs) => TRet {
+      return function (...args: TArgs): TRet {
+        const start = Date.now();
+        const result = func(...args);
```
dphilipson, line 25: Should do `func.apply(this, args)` instead of `func(...args)`, otherwise you lose the value of `this`.

account-kit/logging/src/index.ts
dphilipson: If you did want to keep a single `TFunc` param, this declaration worked for me with the code you just pasted:
```ts
function profiled<TFunc extends (...args: unknown[]) => unknown>(
  name: string,
  func: TFunc,
): (...args: Parameters<TFunc>) => ReturnType<TFunc> {
```
that is, declaring the return type of `profiled` to be that derived type rather than `TFunc`. But that turns out more verbose than what you have so I like it as is.

dphilipson: Also, I suspect that cast to `TRet` is needed for similar reasons to above: after the `instanceof Promise` check, we know it's a `Promise`, but it could also be more and have additional stuff on it (I've seen promises with `.cancel` methods on them, for example), whereas the thing we're returning is just a vanilla promise, so the type system can't assume that the return type is unchanged when you call `.then`.

account-kit/logging/src/index.ts
moldy530: should `this` here be the `func`?

I thought if we did `func.apply(this...` the `this` here is actually object we're returning which contains `profiled` and `trackEvent` in it?

dphilipson: It would do that if we were returning an arrow function e.g.
```ts
return (...args: TArgs): TRet => { /* ... */ }
```
because arrow functions capture the surrounding value of `this` which is more often what you want. But since we're returning a function declared with the `function` keyword, `this` will refer to the object that the returned function is ultimately "called on": that is, whatever value appears to the left of the dot when this function is called with method call syntax. So for example, in
```ts
const sender = {
    sendUserOperation: ReactLogger.profiled(
      "sendUserOperation",
      sendUserOperation
    ),
};

sender.sendUserOperation(op);
```
in the call to `sender.sendUserOperation(op)`, `this` will be `sender`.

dphilipson (approved):
Looks great, thanks for the updates!

PR #1019, opened 2024-10-06
feat: expose signer error messages

dphilipson:
Adds a new field to the signer store state, `error`, containing error information, and populates it on errors thrown by `authenticate` as well as during signer creation if there are error query params in the URL. Adds a similar field to the `signerStatus` field of the main store state. Finally, removes SDK-related params from the URL after reading them, both the new error params as well as the existing bundle params.

# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing error handling in the `AlchemySigner` implementation and updating several related components to accommodate the new error management features.

### Detailed summary
- Added `error` property to `SignerStatus` and `AlchemySignerStore`.
- Introduced `ErrorInfo` interface for structured error information.
- Updated `convertSignerStatusToState` to accept and handle errors.
- Modified state updates to retain error information during status changes.
- Enhanced error handling in the `authenticate` method.
- Updated tests and documentation to reflect error handling improvements.

### DIFF
--- a/account-kit/core/src/store/store.ts
+++ b/account-kit/core/src/store/store.ts
@@ -4,7 +4,11 @@ import {
   createAlchemyPublicRpcClient,
   type AlchemyTransportConfig,
 } from "@account-kit/infra";
-import { AlchemySignerStatus, AlchemyWebSigner } from "@account-kit/signer";
+import {
+  AlchemySignerStatus,
+  AlchemyWebSigner,
+  type ErrorInfo,
+} from "@account-kit/signer";
 import type { Chain } from "viem";
 import { deepEqual } from "wagmi";
 import {
@@ -145,7 +149,10 @@ const createInitialStoreState = (
     connections: connectionMap,
     accountConfigs,
     config: { client, sessionConfig },
-    signerStatus: convertSignerStatusToState(AlchemySignerStatus.INITIALIZING),
+    signerStatus: convertSignerStatusToState(
+      AlchemySignerStatus.INITIALIZING,
+      undefined
+    ),
     smartAccountClients: createEmptySmartAccountClientState(chains),
   };
 
@@ -196,17 +203,6 @@ export const createSigner = (params: ClientStoreConfig) => {
     sessionConfig,
   });
 
-  const search = new URLSearchParams(window.location.search);
-  if (search.has("bundle")) {
-    signer.authenticate({ type: "email", bundle: search.get("bundle")! });
-  } else if (search.has("alchemy-bundle")) {
-    signer.authenticate({
-      type: "oauthReturn",
-      bundle: search.get("alchemy-bundle")!,
-      orgId: search.get("alchemy-org-id")!,
-    });
-  }
-
   if (client.enablePopupOauth) {
     signer.preparePopupOauth();
   }
@@ -218,12 +214,15 @@ export const createSigner = (params: ClientStoreConfig) => {
  * Converts the AlchemySigner's status to a more readable object
  *
  * @param {AlchemySignerStatus} alchemySignerStatus Enum value of the AlchemySigner's status to convert
+ * @param {ErrorInfo | undefined} error the current signer error, if present
  * @returns {SignerStatus} an object containing the original status as well as booleans to check the current state
  */
 export const convertSignerStatusToState = (
-  alchemySignerStatus: AlchemySignerStatus
+  alchemySignerStatus: AlchemySignerStatus,
+  error: ErrorInfo | undefined
 ): SignerStatus => ({
   status: alchemySignerStatus,
+  error,
   isInitializing: alchemySignerStatus === AlchemySignerStatus.INITIALIZING,
   isAuthenticating:
     alchemySignerStatus === AlchemySignerStatus.AUTHENTICATING ||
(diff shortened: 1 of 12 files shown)


account-kit/signer/src/base.ts
```diff
@@ -66,14 +69,19 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
    * @param {TClient} param0.client The client instance to be used internally
    * @param {SessionConfig} param0.sessionConfig Configuration for managing sessions
```
moldy530, line 70: this might need to be updated to include the `param0.initialError` now

account-kit/signer/src/base.ts
```diff
@@ -84,15 +92,6 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
       ...sessionConfig,
       client: this.inner,
     });
-    this.store = createStore(
```
moldy530, line 87: oh wow, good catch

account-kit/core/src/store/store.ts
```diff
@@ -265,7 +275,16 @@ const addClientSideStoreListeners = (store: Store) => {
...
+        const shouldClearError =
+          status === AlchemySignerStatus.AUTHENTICATING ||
+          status === AlchemySignerStatus.AWAITING_EMAIL_AUTH ||
+          status === AlchemySignerStatus.CONNECTED;
+        store.setState((state) => ({
+          signerStatus: convertSignerStatusToState(
+            status,
+            shouldClearError ? undefined : state.signerStatus.error
+          ),
+        }));
```
moldy530, line 287: can we add tests for this in `store.test.ts`?

dphilipson: Simplified the logic so it shouldn't need more testing than the rest of the listeners now.

account-kit/core/src/store/store.ts
```diff
@@ -196,14 +203,20 @@ export const createSigner = (params: ClientStoreConfig) => {
...
   });
 
-  const search = new URLSearchParams(window.location.search);
-  if (search.has("bundle")) {
-    signer.authenticate({ type: "email", bundle: search.get("bundle")! });
-  } else if (search.has("alchemy-bundle")) {
+  const { emailBundle, oauthBundle, oauthOrgId } = takeQueryParams({
+    emailBundle: "bundle",
+    // We don't need this, but we still want to remove it from the URL.
+    emailOrgId: "orgId",
```
dphilipson, line 209: @moldy530 Is this right? I didn't see anywhere we're using the "orgId" query param.

account-kit/core/src/store/store.ts
```diff
@@ -311,3 +341,29 @@ export const createEmptySmartAccountClientState = (chains: Chain[]) => {
...
+
+/**
+ * Reads and removes the specified query params from the URL.
+ *
+ * @param {T} keys object whose values are the query parameter keys to read and
+ * remove
+ * @returns {{ [K in keyof T]: string | undefined }} object with the same keys
+ * as the input whose values are the values of the query params.
+ */
+function takeQueryParams<T extends Record<string, string>>(
```
rthomare, line 353: nit the function name at first felt like a read only function, should we make this more explicit? We split it into two functions? Rename it to getAndUpdateQueryUrlParam?

Not going to block on this btw.

rthomare: We also have a very similar function in account kit signer. Should we consolidate somewhere?

moldy530: if we consolidate, probs best to put it in signer since it's pretty signer specific

account-kit/core/src/store/store.ts
```diff
@@ -218,12 +231,15 @@ export const createSigner = (params: ClientStoreConfig) => {
  * Converts the AlchemySigner's status to a more readable object
  *
  * @param {AlchemySignerStatus} alchemySignerStatus Enum value of the AlchemySigner's status to convert
+ * @param {ErrorInfo | undefined} error the current signer error, if present
  * @returns {SignerStatus} an object containing the original status as well as booleans to check the current state
  */
 export const convertSignerStatusToState = (
-  alchemySignerStatus: AlchemySignerStatus
+  alchemySignerStatus: AlchemySignerStatus,
+  error: ErrorInfo | undefined
```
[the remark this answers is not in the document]
dphilipson: The way this function is being used, it would most likely be a mistake to omit the error parameter. Passing `undefined` here clears the error state, which should be a deliberate choice and not the default behavior of leaving out a parameter.

account-kit/core/src/store/store.ts
dphilipson: You're right, I should make the name more clear. I was following the Rust naming convention of take meaning to remove the current value and return it, e.g. [Option::take](https://doc.rust-lang.org/std/option/enum.Option.html#method.take), but that's unlikely to be familiar to readers here.

account-kit/core/src/store/store.ts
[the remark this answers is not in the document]
rthomare: Yeah that is very fair.

account-kit/core/src/store/store.ts
dphilipson: Just pushed a version that consolidates all query param reading into the signer constructor.

rthomare (approved):

PR #1059, opened 2024-10-14
fix: open the auth modal upon email return

dphilipson:
On mobile, if the user clicks an email magic link, the modal should open immediately to display that they are completing email login.

This has the side effect of opening the modal for OAuth redirect login too, because at the moment email and OAuth share a loading state when returning.

---

## PR-Codex overview
This PR focuses on modifying the handling of headers in API routes and enhancing the authentication flow in the `AuthCardContent` component.

### Detailed summary
- In `route.ts`, replaced `...req.headers` with `...headers` for better header management.
- In `route.ts`, added `openAuthModal` to the destructured values from `useAuthModal`.
- In `AuthCardContent`, called `openAuthModal()` when the authentication step is not initial and the user is authenticating.

### DIFF
--- a/account-kit/react/src/components/auth/card/index.tsx
+++ b/account-kit/react/src/components/auth/card/index.tsx
@@ -56,7 +56,7 @@ export const AuthCardContent = ({
   className?: string;
   showClose?: boolean;
 }) => {
-  const { closeAuthModal } = useAuthModal();
+  const { openAuthModal, closeAuthModal } = useAuthModal();
   const { status, isAuthenticating } = useSignerStatus();
   const { authStep, setAuthStep } = useAuthContext();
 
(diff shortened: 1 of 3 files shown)


rthomare (approved):

account-kit/react/src/components/auth/card/index.tsx
```diff
@@ -114,6 +114,7 @@ export const AuthCardContent = ({
     } else if (authStep.type !== "initial") {
       didGoBack.current = false;
     } else if (!didGoBack.current && isAuthenticating) {
+      openAuthModal();
```
moldy530, line 117: should this step always be `email_completing`? how is the oauth step with redirect handled?

dphilipson: I was confused about that too- I think at the moment, oauth with redirect also puts it in an "email_completing" state. That's a bug because it will have the wrong text, but it doesn't affect the demo app. I'll make a ticket for it.

PR #1087, opened 2024-10-16
fix: don't show error view when cancelling oauth

dphilipson:
This is an intentional action from the user and not an error. Instead, just return to the main screen.

---

## PR-Codex overview
This PR focuses on enhancing the OAuth flow in the `CompletingOAuth` component by adding error handling for cancelled OAuth processes and updating the dependencies.

### Detailed summary
- Added `@account-kit/signer` as a new dependency.
- Imported `OauthCancelledError` from `@account-kit/signer`.
- Introduced a check for `oauthWasCancelled` to handle cancelled OAuth scenarios.
- Updated the dependency array in the `useEffect` hook to include `oauthWasCancelled`.
- Modified the error handling to exclude cancelled OAuth errors from triggering the `ConnectionError` component.

### DIFF
--- a/account-kit/react/src/components/auth/card/loading/oauth.tsx
+++ b/account-kit/react/src/components/auth/card/loading/oauth.tsx
@@ -5,11 +5,13 @@ import { capitalize } from "../../../../utils.js";
 import { useAuthContext } from "../../context.js";
 import { useOAuthVerify } from "../../hooks/useOAuthVerify.js";
 import { ConnectionError } from "../error/connection-error.js";
+import { OauthCancelledError } from "@account-kit/signer";
 
 export const CompletingOAuth = () => {
   const { isConnected } = useSignerStatus();
   const { setAuthStep, authStep } = useAuthContext("oauth_completing");
   const { authenticate } = useOAuthVerify({ config: authStep.config });
+  const oauthWasCancelled = authStep.error instanceof OauthCancelledError;
 
   useEffect(() => {
     if (isConnected) {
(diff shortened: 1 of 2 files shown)


account-kit/react/src/components/auth/card/loading/oauth.tsx
```diff
@@ -10,14 +10,17 @@ export const CompletingOAuth = () => {
   const { isConnected } = useSignerStatus();
   const { setAuthStep, authStep } = useAuthContext("oauth_completing");
   const { authenticate } = useOAuthVerify({ config: authStep.config });
+  const oauthWasCancelled = authStep.error?.name === "OauthCancelledError";
```
rthomare, line 13: Dumb question is this the same across all Oauth providers? 

Is this an error that we propagate down? Should we store it in a const file somewhere?

rthomare (commented):
Generally looks good to me, just a flag to think about storing the error string in a const file in case gets changed it later.

account-kit/react/src/components/auth/card/loading/oauth.tsx
dphilipson: Actually, you're right, let me use a constant here. It's the name of a class in the SDK.

moldy530: would the error be `instanceof OauthCancelledError` here?

dphilipson: Yes it would, but `OauthCancelledError` is exported from `@account-kit/signer`, but not from any of the direct dependencies `@account-kit/react` (core, infra, and logging). So I need to either expose it from `core` or add `signer` as a dependency to `react`.

Exposing it from `core` makes a lot of sense. One question is whether it should be exposed from `react` as well.

moldy530: react should have implicit access to `@account-kit/signer` because of its dependency on core right? I think we should just add signer as a dependency to `react` potentially if we don't want it to be implicit. because of the monorepo setup and the fact that we keep versions tied together, it should be fine because core `vX` will always depend on the same signer `vX` as react

moldy530 (dismissed):

moldy530:
hmmm aight so the site is failing to deploy consistently which makes me wonder something funky is happening in main. will take a look

PR #1090, opened 2024-10-17
fix: fix bug when trying to re-try EOA connections from auth card

iykazrji:
# Pull Request Checklist

# Testing plan
- Simulate an EOA connection failure (WalletConnect, Metamask, CoinbaseWallet)
- Hit on `Try Again` from the Auth Card
- This should re-start the attempted EOA connection process.

---

## PR-Codex overview
This PR focuses on enhancing the authentication process by introducing a new hook `useConnectEOA`, improving error handling for Wallet Connect, and optimizing the connection flow for EOA and Wallet Connect in the authentication cards.

### Detailed summary
- Added `useConnectEOA` hook for EOA connection management.
- Introduced `clearWalletConnectStore` function to handle Wallet Connect state.
- Updated `ConnectionError` component to disconnect on error.
- Refactored `AuthCardContent` to use `disconnect` instead of `signer.disconnect`.
- Enhanced `EoaConnectCard` and `WalletConnectCard` to retry connections.
- Removed redundant connection logic in `EoaPickCard`.


account-kit/react/src/components/auth/card/error/connection-error.tsx
```diff
@@ -86,3 +103,23 @@ export const ConnectionError = ({
...
+    const store = txn.objectStore("keyvaluestorage");
+
+    // Clear Store
+    store.clear();
+  };
+
+  dbOpenRequest.onerror = () => {
+    console.error("Error opening Wallet Connect DB. Cannot clear store.");
+  };
+};
```
iykazrji, line 125: Adding this function to clear the IndexedDB store wallet connect  populates on attempted authentication. This is to address the fact that we could get into continuous error loops when an auth attempt fails.

rthomare (dismissed):

account-kit/react/src/components/auth/card/error/connection-error.tsx
moldy530: oh added a comment on the other PR but we might want to add this to the `disconnect` action from `account-kit/core`

moldy530: actually... looking at the definition for wallet connect providers, https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/providers/universal-provider/src/UniversalProvider.ts#L546

the session should not be getting used on retry if we're using the `disconnect` action from core which calls the `wagmi_disconnect` method. That method then calls the connectors disconnect which eventually clears the wallet connect session.

iykazrji: From my tests, this doesn't seem to be the case. I'd be sending a video recording to explain the behaviour calling `disconnect` from `@account-kit/core`

PR #1108, opened 2024-10-26
feat: add exit animation to mobile dialog drawer

iykazrji:
# Pull Request Checklist

## Screenshots
Before:
https://www.loom.com/share/df896258fff140468e6b809e343efbb7?sid=7120c059-5567-49aa-a4fc-9ecf4052de68

After:
https://www.loom.com/share/81015edf28384c5aad5edd3137ea271b?sid=58f49f6a-b930-4a55-8f23-ddd97f08a3ae

### DIFF
--- a/account-kit/react/src/components/dialog/dialog.tsx
+++ b/account-kit/react/src/components/dialog/dialog.tsx
@@ -4,6 +4,7 @@ import { useCallback, useEffect, useState, type ReactNode } from "react";
 import { createPortal } from "react-dom";
 import { RemoveScroll } from "react-remove-scroll";
 import { FocusTrap } from "./focustrap.js";
+import { Transition } from "react-transition-group";
 
 type DialogProps = {
   isOpen: boolean;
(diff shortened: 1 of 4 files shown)


moldy530 (commented):
looks good, but can we avoid using the 3rd party library?

using the `viewTransition` api directly might be overkill. Can we instead either, 
1. remove the `isOpen ? ...` piece, always render the portal, and just selectively add the slide up or down animation based on the isOpen property
2. useEffect triggered by changes on the on `isOpen`. This will in turn start the animation (up or down). If the animation is down, then on animation finished, remove the portal if we don't want it always rendered. Probs would need some internal `show` state as well here to handle this.

account-kit/react/src/components/dialog/dialog.tsx
```diff
@@ -19,11 +19,22 @@ type DialogProps = {
...
 
+  useEffect(() => {
+    if (isOpen) {
+      setShowBackdrop(true);
+    } else {
+      setTimeout(() => {
+        setShowBackdrop(false);
+      }, 200);
+    }
+  }, [isOpen]);
```
moldy530, line 36: this feels slightly janky because there's a small, but noticeable pause before the backdrop is hidden

I'd suggest adding a `useRef<HTMLDivElement>()` and capturing a ref to the `div` with the `animate-slide-{up|down}` class below

Then this `useEffect` could do something like
 
```tsx
useLayoutEffect(() => {
  if (!ref.current) return;
  if (isOpen) {
    setRenderPortal(true);
    return;
  }

  const listener = () => { setRenderPortal(false); };
  ref.current.addEventListener("animationend", listener);

  return () => { ref.current.removeEventListener("animationend", listener); }
}, [isOpen])
```

Then below we can:

```
return (isOpen || renderPortal) ? createPortal(...) : null;
```

This removes the need to add a custom class based on `showBackdrop` and avoids creating portal if we don't need it

account-kit/react/src/components/dialog/dialog.tsx
```diff
@@ -40,27 +51,31 @@ export const Dialog = ({ isOpen, onClose, children }: DialogProps) => {
...
+  return createPortal(
+    <RemoveScroll enabled={isScrollLocked}>
+      {/* Overlay */}
+      <div
+        aria-modal
+        role="dialog"
+        className={`fixed inset-0 bg-black/80 flex items-end md:items-center justify-center z-[999999] ${
+          !showBackdrop
+            ? "animate-fade-out invisible"
+            : "animate-fade-in visible"
```
moldy530, line 63: I actually am not seeing this animation run at all

account-kit/react/src/components/dialog/dialog.tsx
iykazrji: This would mean the backdrop would not have an exit animation, are we okay with that? We could also make the timeout to match the animation duration of the slideout animation (currently 150ms)

account-kit/react/src/components/dialog/dialog.tsx
iykazrji: Yeah! I'd fix this! Seen what the issue is

account-kit/react/src/components/dialog/dialog.tsx
moldy530: it should still be animated during the same animation as the slide down, but gets removed from the page once the slide down is done.

yea alternatively, just tweaking the animations to have them match so that they don't seem disconnected should work

account-kit/react/src/components/dialog/dialog.tsx
```diff
@@ -40,19 +69,24 @@ export const Dialog = ({ isOpen, onClose, children }: DialogProps) => {
...
-  return isOpen
+  return renderPortal
     ? createPortal(
         <RemoveScroll enabled={isScrollLocked}>
           {/* Overlay */}
           <div
             aria-modal
             role="dialog"
-            className="fixed inset-0 bg-black/80 flex items-end md:items-center justify-center z-[999999] animate-fade-in"
+            ref={backdropRef}
```
moldy530, line 79: I think we want to listen to the animation ending on the `slide-down` below not the backdrop

moldy530 (approved):

PR #1112, opened 2024-10-29
fix(signer): expire sessions on focus and on timer after init

moldy530:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing session management in the `SessionManager` class by implementing a mechanism to automatically clear expired sessions.

### Detailed summary
- Added a new property `clearSessionHandle` to track session expiration timers.
- Modified `clearSession` method to clear the timeout if it exists.
- Introduced `registerSessionExpirationHandler` method to set a timeout for session expiration.
- Updated `setSession` method to utilize `registerSessionExpirationHandler` for new sessions.

### DIFF
--- a/account-kit/signer/src/session/manager.ts
+++ b/account-kit/signer/src/session/manager.ts
@@ -45,6 +45,7 @@ export class SessionManager {
   private eventEmitter: EventEmitter<SessionManagerEvents>;
   readonly expirationTimeMs: number;
   private store: Store;
+  private clearSessionHandle: NodeJS.Timeout | null = null;
 
   constructor(params: SessionManagerParams) {
     const {
@@ -125,6 +126,10 @@ export class SessionManager {
 
   public clearSession = () => {
     this.store.setState({ session: null });
+
+    if (this.clearSessionHandle) {
+      clearTimeout(this.clearSessionHandle);
+    }
   };
 
   public setTemporarySession = (session: { orgId: string }) => {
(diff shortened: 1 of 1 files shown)


account-kit/signer/src/session/manager.ts
```diff
@@ -170,10 +170,14 @@ export class SessionManager {
...
      */
     if (session.expirationDateMs < Date.now()) {
-      this.store.setState({ session: null });
+      this.clearSession();
       return null;
     }
 
+    setTimeout(() => {
+      this.clearSession();
+    }, session.expirationDateMs - Date.now());
```
dphilipson, line 179: Slightly odd in that if `getSession` is called multiple times, then it will set this timeout multiple times, although I don't know if that actually causes any problems. Still, I would consider storing the timeout id and clearing the old timeout when setting a new one.

dphilipson (dismissed):
Thanks for being so quick! Left a comment.

account-kit/signer/src/session/manager.ts
dphilipson: Also: I don't know if this would be an issue, but there's a funny thing where if you set a timeout longer than 2^31, or around 25 days, then it triggers immediately. It may be worth putting in a check for that, in case someone asks for like 30 day session expirations.

moldy530: oh wow... TIL about setTimeout max being 24.8 days. yea I'll make sure to check that actually

re: getSession being called multiple times, yea good call. I'll add a global handler for this

dphilipson (approved):
Looks great, thanks for the changes!

PR #1120, opened 2024-10-31
feat: added dev:build script to improve dev experience

RobChangCA:
# Pull Request Checklist

When you run dev:build from the account-kit/react package lerna will watch for changes and run a build allowing consuming apps such as the demo app to automatically update with your changes instead of manually running build.

---

## PR-Codex overview
This PR introduces a new command for building the `@account-kit/react` package in development mode using `lerna`. It also updates the `package.json` to streamline the build process.

### Detailed summary
- Added a new script `build:dev` to `package.json` for development builds.
- The `build:dev` script uses `lerna watch` to monitor changes and run builds for the `@account-kit/react` package.

### DIFF
--- a/account-kit/react/package.json
+++ b/account-kit/react/package.json
@@ -37,8 +37,8 @@
   "scripts": {
     "prebuild": "npx tsx ./inject-version.ts",
     "dev": "storybook dev",
-    "dev:build": "lerna watch --scope='@account-kit/react' --include-dependencies -- lerna run build --scope=@account-kit/react",
     "build": "yarn clean && yarn build:esm && yarn build:types",
+    "build:dev": "lerna watch --scope='@account-kit/react' --include-dependencies -- lerna run build --scope=@account-kit/react",
     "build:esm": "tsc --project tsconfig.build.json --outDir ./dist/esm",
     "build:types": "tsc --project tsconfig.build.json --declarationDir ./dist/types --emitDeclarationOnly --declaration --declarationMap",
     "docs:gen": "npx ak-docgen generate --in ./src/index.ts --out ../../site/pages/reference/account-kit/react",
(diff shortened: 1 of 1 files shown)


account-kit/react/package.json
```diff
@@ -37,6 +37,7 @@
   "scripts": {
     "prebuild": "npx tsx ./inject-version.ts",
     "dev": "storybook dev",
+    "dev:build": "lerna watch --scope='@account-kit/react' --include-dependencies -- lerna run build --scope=@account-kit/react",
```
moldy530, line 40: what cases does this cover over the `yarn dev` script? I remember testing this some before and I didn't have to build when I made changes to files in `@account-kit/react` because storybook was using the local TS files instead of the built output of the lib

moldy530: also nit, if we do need the script can we make it `build:dev`?

RobChangCA: I wrote a more detailed description in the channel, should have posted it here as well my bad. 

I found that I was missing regressions when switching between branches because I forgot to run build on our account kit library resulting in the wrong version rendering in the demo app locally.  This script dev:build will alleviate that issue by watching changes in the account-kit/react package and rebuild when necessary.

moldy530 (approved):

account-kit/react/package.json
moldy530: should we just include this in the `dev` script then? can we have it run `storybook dev` and the `lerna` script?

in the past i've used `stmux` to start these watcher tasks in split windows. Something like this:
```
"dev": "stmux -w always -M -- [ -t storybook \"storybook dev\" .. -t build \"yarn build:dev\" ]"
```

RobChangCA: IMO we should keep them separate as it will significantly slow down build speeds if we only intend to be working in storybook

PR #1149, opened 2024-11-13
fix: add check to ensure WC indexedDB store had been previously initialized on disconnect

iykazrji:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing the `clearWalletConnectStore` function to handle the scenario where the `WALLET_CONNECT_V2_INDEXED_DB` may not exist, ensuring the database is deleted if it is outdated.

### Detailed summary
- Added a check for `walletConnectDBExists` to determine if the database exists.
- Implemented an `onupgradeneeded` event to delete the database if the version is outdated.
- Wrapped the database operations in a try-catch block to handle errors gracefully.


account-kit/core/src/actions/disconnect.ts
```diff
@@ -46,17 +46,36 @@ export async function disconnect(config: AlchemyAccountsConfig): Promise<void> {
 // Persistence of Wallet Connect connection state on error.
 const clearWalletConnectStore = () => {
   // Open Wallet Connect Indexed DB
+  let walletConnectDBExists = true;
```
moldy530, line 49: we should probably start with the assumption that it isn't initialized

account-kit/core/src/actions/disconnect.ts
```diff
@@ -46,17 +46,36 @@ export async function disconnect(config: AlchemyAccountsConfig): Promise<void> {
...
   const dbOpenRequest = indexedDB.open("WALLET_CONNECT_V2_INDEXED_DB");
 
+  dbOpenRequest.onupgradeneeded = () => {
+    if (dbOpenRequest.result.version === 1) {
+      walletConnectDBExists = false;
+
+      // Remove the Database created from the indexedDB.open() call above.
+      indexedDB.deleteDatabase("WALLET_CONNECT_V2_INDEXED_DB");
```
moldy530, line 57: wouldn't this throw if this state means it's not connected?

account-kit/core/src/actions/disconnect.ts
```diff
@@ -46,17 +46,36 @@ export async function disconnect(config: AlchemyAccountsConfig): Promise<void> {
 // Persistence of Wallet Connect connection state on error.
 const clearWalletConnectStore = () => {
   // Open Wallet Connect Indexed DB
+  let walletConnectDBExists = true;
   const dbOpenRequest = indexedDB.open("WALLET_CONNECT_V2_INDEXED_DB");
```
moldy530, line 50: do we need to do any other checks here? does `indexDb.open` ever throw?

account-kit/core/src/actions/disconnect.ts
iykazrji: Actually, by this point the DB would actually already have been created by the previous `indexedDB.open()` call made above. It's the behaviour for IDB to create a database if it doesn't exist when we call `indexedDB.open()`.

account-kit/core/src/actions/disconnect.ts
iykazrji: Yeah, but we're already handling the error case for that with the `dbOpenRequest.onerror()` handler below

account-kit/core/src/actions/disconnect.ts
iykazrji: Actually on further testing, it is better we start with the assumption that it is in fact initialized. This is because `dbRequest.onUpgradeneeded()` isn't actually called on every request, only when there is a DB version change. This means when we call `disconnect()` after walletConnect has created the IDB store, we wouldn't be running that event handler, and thus if we default to false, we wouldn't have a way to change the value to true.

account-kit/core/src/actions/disconnect.ts
moldy530: I see, so lemme run this back to see if I understand

`db.open` will create a db if it doesn't exist or open the existing one if it does.

if it doesn't exist and thus gets created, that fires the upgrade event here. We know it was created because version is 1 indexed (not 0 indexed?)

Since this only gets called on disconnect, there's no risk that this gets fired when WC first creates the db

moldy530 (approved):
LGTM, other sanity check - we don't need to remove the handlers right? they only get attached to this specific open request, not globally? Wanna make sure that, if you call disconnect (previously logged in as email) and then try to connect with WC for the first time, then you won't end up in a state where this handler deletes the DB because this handler is still registered

iykazrji:
Yes! The handlers are only attached to this `dbOpenRequest` called only when the `disconnect()` call happens.

PR #1209, opened 2024-12-12
feat: add entity id nonce logic

howydev:
## PR-Codex overview
This PR primarily focuses on updating the versions of various packages and components in the project, along with adding a new React Native example application. It includes improvements to the Android and iOS configurations, as well as updates to the smart contracts and utilities.

### Detailed summary
- Bumped version numbers across multiple packages to `4.6.1`.
- Added a new React Native example app with configuration files.
- Updated `AndroidManifest.xml` and `AppDelegate` for proper app setup.
- Introduced new utility functions and modules in smart contracts.
- Enhanced project structure and build configurations for better performance.

> The following files were skipped due to too many changes: `examples/react-native-bare-example/ios/ReactNativeBareExample/LaunchScreen.storyboard`, `examples/react-native-bare-example/src/screens/Home.tsx`, `account-kit/smart-contracts/src/ma-v2/account/account.ts`, `account-kit/smart-contracts/src/ma-v2/modules/time-range-module/abis/timeRangeModule.ts`, `examples/react-native-bare-example/android/app/build.gradle`, `account-kit/smart-contracts/src/ma-v2/modules/single-signer-validation/abis/singleSignerValidationModule.ts`, `account-kit/smart-contracts/src/ma-v2/modules/webauthn-validation/abis/webauthnValidation.ts`, `account-kit/smart-contracts/src/ma-v2/modules/native-token-limit-module/abis/nativeTokenLimitModule.ts`, `examples/react-native-bare-example/android/gradlew`, `account-kit/smart-contracts/src/ma-v2/abis/accountFactoryAbi.ts`, `account-kit/smart-contracts/src/ma-v2/modules/allowlist-module/abis/allowlistModule.ts`, `account-kit/smart-contracts/src/ma-v2/abis/modularAccountAbi.ts`, `account-kit/smart-contracts/src/ma-v2/abis/semiModularAccountBytecodeAbi.ts`, `account-kit/smart-contracts/src/ma-v2/abis/semiModularAccountStorageAbi.ts`, `examples/react-native-bare-example/ios/ReactNativeBareExample.xcodeproj/project.pbxproj`, `yarn.lock`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #1220, opened 2024-12-14
feat: basic ma v2 sdk integrations

howydev:
## PR-Codex overview
This PR focuses on adding new error classes and enhancing existing modules in the `aa-sdk` and `account-kit` projects, particularly around entity ID and nonce key validation. It also updates various modules to support these changes.

### Detailed summary
- Added error classes: 
  - `InvalidEntityIdError`
  - `InvalidNonceKeyError`
  - `EntityIdOverrideError`
- Updated `forkBlockNumber` in `local070Instance`.
- Enhanced `SingleSignerValidationModule` with new encoding functions.
- Introduced documentation for new error classes.
- Updated multiple modules to incorporate new error handling and validation logic.

> The following files were skipped due to too many changes: `account-kit/smart-contracts/src/ma-v2/abis/accountFactoryAbi.ts`, `account-kit/smart-contracts/src/ma-v2/modules/allowlist-module/abis/allowlistModuleAbi.ts`, `account-kit/smart-contracts/src/ma-v2/abis/modularAccountAbi.ts`, `account-kit/smart-contracts/src/ma-v2/abis/semiModularAccountBytecodeAbi.ts`, `account-kit/smart-contracts/src/ma-v2/abis/semiModularAccountStorageAbi.ts`, `account-kit/smart-contracts/src/ma-v2/client/client.test.ts`


.vitest/src/instances.ts
```diff
@@ -10,6 +10,17 @@ import { split } from "../../aa-sdk/core/src/transport/split";
...
+export const local070InstanceOptSep = defineInstance({
+  chain: localhost,
+  forkBlockNumber: 21153995,
+  forkUrl:
+    process.env.VITEST_OPT_SEPOLIA_FORK_URL ??
+    "https://opt-sepolia-rpc.publicnode.com",
+  entryPointVersion: "0.7.0",
+  anvilPort: 8345,
+  bundlerPort: 8445,
+});
```
moldy530, line 22: why do we need opt sepolia here?

account-kit/smart-contracts/src/ma-v2/account/account.ts
moldy530, line 1: nit: could we call this file smav2 instead?

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+export const DEFAULT_OWNER_ENTITY_ID = 0;
+
+export type SMAV2Account<
+  TSigner extends SmartAccountSigner = SmartAccountSigner
+> = SmartContractAccountWithSigner<"SMAV2Account", TSigner, "0.7.0">;
+
+export type CreateSMAV2AccountParams<
+  TTransport extends Transport = Transport,
+  TSigner extends SmartAccountSigner = SmartAccountSigner,
+  TEntryPointVersion extends "0.7.0" = "0.7.0"
```
moldy530, line 40: if we only plan on supporting 0.7 then I don't think you need this here

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+> = Pick<
+  ToSmartContractAccountParams<
+    "SMAV2Account",
+    TTransport,
+    Chain,
+    TEntryPointVersion
```
moldy530, line 46: you can just hard code the `0.7.0` here

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+  >,
+  "transport" | "chain"
+> & {
+  signer: TSigner;
+  salt?: bigint;
+  factoryAddress?: Address;
+  initCode?: Hex;
+  initialOwner?: Address;
+  accountAddress?: Address;
```
moldy530, line 55: you can get this from the `Pick` above as well

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+  entryPoint?: EntryPointDef<TEntryPointVersion, Chain>;
+  isGlobalValidation?: boolean;
+  entityId?: bigint;
```
moldy530, line 58: I saw a thread from @Blu-J as well around this, but think that PR got closed and comments lost. Where was that thread, there was some interesting discussion in there?

I'm curious what the relationship is between entityId and globalValidation? can you pass in an `entityId != 0` and `isGlobalValidation = true`?

moldy530: Is global validation and entityId always referring to the signer? If so, maybe it makes sense to move some of that logic into a special `Signer` type for MAv2 that wraps an underlying `SmartAccountSigner` type.

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+    Chain,
+    TEntryPointVersion
+  >,
+  "transport" | "chain"
+> & {
+  signer: TSigner;
+  salt?: bigint;
+  factoryAddress?: Address;
+  initCode?: Hex;
+  initialOwner?: Address;
```
moldy530, line 54: how is this used? if it's for deferred execution, we could just require that `initCode` be required when using defered execution.

It might be helpful to understand what params need to be passed in for:
1. creating an instance for the root owner
2. creating an instance for a session key signer

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+    initialOwner,
+    accountAddress,
+    entryPoint = getEntryPoint(chain, { version: "0.7.0" }),
+    isGlobalValidation = true,
+    entityId = 0n,
+  } = config;
+
+  if (entityId >= 2n ** 32n) {
+    throw new Error("Entity ID must be less than uint32.max");
+  }
```
moldy530: think I saw a suggestion from @jaypaik that viem exports this as a constant as well, so we can just check against that

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+    initCode,
+    initialOwner,
+    accountAddress,
+    entryPoint = getEntryPoint(chain, { version: "0.7.0" }),
+    isGlobalValidation = true,
+    entityId = 0n,
+  } = config;
+
+  if (entityId >= 2n ** 32n) {
+    throw new Error("Entity ID must be less than uint32.max");
```
moldy530: also saw a comment about using custom errors. similar to how we have an `errors.ts` file in the msca namespace

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    // If an initial owner is not provided, use the signer's address
+    const ownerAddress = initialOwner || (await signer.getAddress());
+
+    return concatHex([
+      factoryAddress,
+      encodeFunctionData({
+        abi: accountFactoryAbi,
+        functionName: "createAccount",
+        args: [ownerAddress, salt, DEFAULT_OWNER_ENTITY_ID],
```
moldy530, line 107: looking at this, this kinda answers one of my earlier questions: it sounds like we can have some stricter typing around passing in an `entityId` and `initCode` override. 

For example, if entityId > 0, then initCode is required as an override and then we don't have to pass in the initial `owner` here

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+    chain,
+  });
+
+  const getAccountInitCode = async () => {
+    if (initCode) {
+      return initCode;
+    }
+
+    // If an initial owner is not provided, use the signer's address
+    const ownerAddress = initialOwner || (await signer.getAddress());
```
moldy530, line 100: nit: 
```suggestion
    const ownerAddress = initialOwner ?? (await signer.getAddress());
```

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+      }),
+    ]);
+  };
+
+  const _accountAddress = await getAccountAddress({
+    client,
+    entryPoint,
+    accountAddress,
+    getAccountInitCode,
+  });
```
moldy530, line 117: to smart account already handles doing this for you

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  const baseAccount = await toSmartContractAccount({
+    transport,
+    chain,
+    entryPoint,
+    accountAddress: _accountAddress,
```
moldy530, line 123: given above comment, I would just pass in the `accountAddress` as it comes through the params

account-kit/smart-contracts/src/ma-v2/modules/single-signer-validation/signer.ts
```diff
@@ -0,0 +1,74 @@
...
+    // TODO: we can't implement these methods yet, because the RI at `alpha.0` doesn't have a wrapping type,
+    // and viem doesn't support raw signing, only via EIP-191 or EIP-712.
+    // When we do implement this, we need to prefix the data with the validation module address & entityId.
+
+    signMessage({
+      message,
+    }: {
+      message: SignableMessage;
+    }): Promise<`0x${string}`> {
+      return signer.signMessage({ raw: hashMessage(message) });
```
moldy530, line 62: is it expected that the result should be doubly hashed and prefixed? `signer.signMessage` will also do `191` over this

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+    source: `SMAV2Account`,
+    getAccountInitCode,
+    ...standardExecutor,
+    ...singleSignerMessageSigner(signer),
+  });
+
+  // TODO: add deferred action flag
```
moldy530, line 130: this likely won't work since `getAccountNonce` will have to match the signature of the base smart contract account impl as well

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,158 @@
...
+  const getAccountNonce = async (nonceKey?: bigint): Promise<bigint> => {
+    // uint32 entityId + (bytes1 options) makes a uint40
+    const nonceKeySuffix: bigint =
+      entityId * 256n + (isGlobalValidation ? 1n : 0n);
```
moldy530: Is this the only place that `isGlobalValidation` gets used? If so, then I don't think we need to override the `getAccountNonce` method. instead we should provide a utility to `buildAccountNonce(nonceKey, isGlobalValidation, isDeferredExecute)` and you can then use that to craft the `nonceKey` when building a UO

we could also leverage a custom middleware on clients as well since passing an override all the time is annoying

account-kit/smart-contracts/src/ma-v2/client/client.ts
```diff
@@ -0,0 +1,73 @@
+import {
+  createSmartAccountClient,
+  type SmartAccountClient,
+  type SmartAccountSigner,
+} from "@aa-sdk/core";
+import { type Chain, type CustomTransport, type Transport } from "viem";
+
+import type { SmartAccountClientConfig } from "@aa-sdk/core";
```
moldy530, line 8: nit: you could merge this import with the one at the top of the file

account-kit/smart-contracts/src/ma-v2/modules/single-signer-validation/signer.ts
moldy530: basically: you should be able to just do `signer.signMessage(message)`

account-kit/smart-contracts/src/ma-v2/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  modularAccountImpl:
+    "0x99090abd2700E24Cc70E3c486A89F7af876fFA33" as `0x${string}`,
+  semiModularAccountBytecodeImpl:
+    "0xDCBb5d4639428B18703801f7Cd7230add729E4b0" as `0x${string}`,
+  semiModularAccountStorageOnlyImpl:
+    "0x563ea43Ba0CD9150466B054Ce47FB0Fb45B234E2" as `0x${string}`,
+  accountFactory: "0xA5FC7aDc6d2c838443dF5bb826152c58b918f2c8" as `0x${string}`,
+  accountFactoryOwner:
+    "0x3BD786ac7Dec96eF9d06ebbCFa4e0E0947A2c303" as `0x${string}`,
+};
```
moldy530, line 40: would love to have these live more closely to the definition. this big list will become a pain to maintain

account-kit/smart-contracts/src/ma-v2/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  validationSignature: Hex;
+};
+
+// Signature packing utility
+export const packSignature = ({
+  // orderedHookData, TO DO: integrate in next iteration of MAv2 sdk
+  validationSignature,
+}: PackSignatureParams): Hex => {
+  return concat(["0xFF", "0x00", validationSignature]);
+};
```
moldy530, line 14: do all signature modules do this? or just `single-signer-validation`?

account-kit/smart-contracts/src/ma-v2/account/account.ts
moldy530: how much of this code is the same for other MAv2 types?

.vitest/src/instances.ts
howydev: our temp deployments are only on opt sepo. we're in the process of salt mining to deploy more efficient ones, those will land next monday

moldy530: oh also this is gonna cause clashes with the other instances because it uses the same port as the `local070instance`. you should be fine to just use the existing 070 instance tbh

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: yeah they're 2 diff vars, but they'll both be used in the nonce. nonceKey has to contain `entityId << 8 | isGlobalValidation`

howydev: +1 on moving to signer, thought that might make sense. if a dapp collects a sig to add a session key then swaps the signer on the account, they would need to replace the signer + entityId + isGlobalValidation, so I think we can pair it up. Will think a little more if this holds across the 3 MA variants

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: Addresses are derived from 1 of: account address, or factory + initcode, or factory + owner + salt + entityid. initcode is a transformation of owner + salt + entityId, so assuming a dapp chooses a fixed salt/entityId for users they'll be equivalent. I guess that means we can deprecate the use of ownerAddress?

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: yep, was in the process of finding comments from earlier

account-kit/smart-contracts/src/ma-v2/client/client.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import { createSMAV2AccountClient } from "./client.js";
+
+import { local070InstanceOptSep } from "~test/instances.js";
+import { setBalance } from "viem/actions";
+import { accounts } from "~test/constants.js";
+
+describe("MA v2 Tests", async () => {
+  const instance = local070InstanceOptSep;
+  const client = instance.getClient().extend(publicActions);
```
moldy530, line 13: I wouldn't do this here actually. the existing LA tests I think do something they shouldn't be doing.

MultiOwnerLA tests do this at the top:
```
const instance = local070Instance;
  let client: ReturnType<typeof instance.getClient>;

  beforeAll(async () => {
    client = instance.getClient();
  });
```

and then the `givenConnectedProvider` passes in a `transport(client)`. Think I ran into some cases where not doing this sometimes ended up with a different instance in the pool to be used

.vitest/src/instances.ts
moldy530: oh I see, then you'll need to update the `setupTests` file as well to account for this instance. or just use the `sepolia` instance in this PR and on monday when we have deployments the tests will pass (I would opt for this so we don't have to manage so many instances in CI and locally)

account-kit/smart-contracts/src/ma-v2/modules/single-signer-validation/signer.ts
howydev: this isn't the exact implementation yet, but we do double hash signatures. the inner one is a 712 containing the account domain

howydev: we should probably be doing a `signTypedData` here instead of the `signMessage`

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: sry can u explain? IIUC we override the base smart account's implementation of `getAccountNonce` when we return it in this constructor

when `buildUserOp`/`initUserOp` happens it calls account.getAccountNonce and it goes here instead of `getNonce` in baseAccount

account-kit/smart-contracts/src/ma-v2/utils.ts
howydev: this is a watered down implementation - when accounts have signature hooks that expect data, that data would be added here

account-kit/smart-contracts/src/ma-v2/client/client.test.ts
howydev: the same happened to me earlier lol, wasn't sure what was happening. glad to know the RCA, will change to this pattern

account-kit/smart-contracts/src/ma-v2/account/account.ts
```diff
@@ -0,0 +1,150 @@
+import type {
+  EntryPointDef,
+  SmartAccountSigner,
+  SmartContractAccountWithSigner,
+  ToSmartContractAccountParams,
+} from "@aa-sdk/core";
+import {
+  createBundlerClient,
+  getAccountAddress,
```
moldy530, line 9: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'getAccountAddress' is defined but never used.

.vitest/src/instances.ts
howydev: hmm since we need this for now, will keep this in + keep this comment up is till Monday/when contracts are deployed, then will remove this instance

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: closed those as graphite was doing wonky stuff. comments: 
1. https://github.com/alchemyplatform/aa-sdk/pull/1209#discussion_r1883012331
2. https://github.com/alchemyplatform/aa-sdk/pull/1209#discussion_r1883016912
3. https://github.com/alchemyplatform/aa-sdk/pull/1209#discussion_r1883022802

I especially like the phantom types one to restricting uint32s to get compile time safety instead of runtime safety. I think we can punt on this for the upcoming sprint till Tuesday. I'll take a look + learn more about phantom types after next week when things are less hectic

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: We default to `entityId = 0` if no entityId is specified. It's not good to do this, but it's possible for someone to create an account with the owner associated to some `entityId != 0`. Because of this, I think we don't want to apply union types here

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: Thinking a little more - we default to entityId = 0 if no entityId is specified, but it's possible for someone to create an account with the owner associated to some entityId != 0 (not the best idea though). But because of that, we probably can't apply stricter typing around that

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: discussed offline, this approach is fine for now

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: discussed offline, current implementation is fine for now

account-kit/smart-contracts/src/ma-v2/utils.ts
howydev: The account parses + validates the prefix in `validationSignature` (signature hooks that aren't here yet + the `FF` byte). So `packSignature` changes depending on the v0.8 ERC6900 account used

The module uses the `00` byte + `validationSignature` - this will change depending on the validation module used

howydev:
Graphite bugged out for me, so I consolidated all the changes for Milestone 1 of basic SDK here

@moldy530's is asking really good q's and there's some MA v2 knowledge transfer happening in the convos here. Tagging @Blu-J @linnall as it might be helpful!

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: whoops - just realized this the above is for MA and not SMA, can disregard

account-kit/smart-contracts/src/ma-v2/account/account.ts
howydev: Differences:
1. both SMA variants default to `entityId = 0` as the owner's entity id
2. cannot upgrade to SMABytecode, only SMAStorage or MA. SMABytecode can upgrade to anything
3. SMAs must block usage of entity id = 0 when installing validations (adding owners, or adding session keys)

howydev:
TODO: another diff between MA and SMA is that 1271/712 is different between the two. We should probably make a `signer` that's SMA specific to account for the 2 possibilities

account-kit/smart-contracts/src/ma-v2/account/semiModularAccountV2.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  // TODO: add deferred action flag
+  const getAccountNonce = async (nonceKey?: bigint): Promise<bigint> => {
+    const nonceKeySuffix: Hex = `${toHex(entityId, { size: 4 })}${
+      isGlobalValidation ? "01" : "00"
+    }`;
+
+    if (nonceKey && toHex(nonceKey, { size: 5 }) !== nonceKeySuffix) {
+      throw new InvalidNonceKeyError(nonceKey, hexToBigInt(nonceKeySuffix));
+    }
```
dphilipson, line 130: This is fine, but something that may be a little simpler would be is doing everything with bigints instead of converting everything to hex strings, comparing them, then converting back, which I found a bit confusing especially with the sizes. It might look like
```ts
const nonceKeySuffix = (entityId << 8n)
  + (isGlobalValidation ? 1n : 0n);
if (nonceKey !== nonceKeySuffix) {
  // ...
```
and leaving out all the `hexToBigInt` below.

Actually, I'm confused here: the variable is called `nonceKeySuffix`, but you're requiring `nonceKey` to be equal to `nonceKeySuffix`, not be suffixed by it. If you were intending to compare suffixes via those `size` parameters, that won't work: the `size` param doesn't truncate, it throws if the input is too long. Is this logic correct?

account-kit/smart-contracts/src/ma-v2/client/client.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    const result = await provider.sendUserOperation({
+      uo: {
+        target: target,
+        value: sendAmount,
+        data: "0x",
+      },
+    });
+
+    const txnHash1 = provider.waitForUserOperationTransaction(result);
+    await expect(txnHash1).resolves.not.toThrowError();
```
dphilipson, line 48: Nit: you could simplify these lines as just
```ts
await provider.waitForUserOperationTransaction(result);
```
which will already fail the text if it results in an error.

account-kit/smart-contracts/src/ma-v2/index.ts
dphilipson, line 1: This file is empty, is it supposed to be here?

account-kit/smart-contracts/src/ma-v2/modules/single-signer-validation/module.ts
```diff
@@ -0,0 +1,38 @@
+import { encodeAbiParameters, type Address, type Hex } from "viem";
+
+import { singleSignerValidationModuleAbi } from "./abis/singleSignerValidationModuleAbi.js";
+
+// Todo: some unified type for ERC-6900 v0.8 modules. I couldn't figure out how to parameterize the class itself over the abi type parameters for onInstall and onUninstall.
+export const SingleSignerValidationModule = {
+  abi: singleSignerValidationModuleAbi,
```
dphilipson, line 7: This parameter `abi` seems unused.

account-kit/smart-contracts/src/ma-v2/modules/single-signer-validation/signer.ts
```diff
@@ -0,0 +1,74 @@
...
+      const dummyEcdsaSignature =
+        "0xfffffffffffffffffffffffffffffff0000000000000000000000000000000007aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa1c";
+
+      return packSignature({
+        // orderedHookData: [],
+        validationSignature: dummyEcdsaSignature,
+      });
+    },
+
+    signUserOperationHash: (uoHash: `0x${string}`): Promise<`0x${string}`> => {
```
dphilipson, line 44: Nit: use the type `Hex` from viem over `0x${string}`.

dphilipson (commented):
Looks good! Some questions and nitpicky comments inline

account-kit/smart-contracts/src/ma-v2/account/semiModularAccountV2.ts
```diff
@@ -0,0 +1,149 @@
...
+    factoryAddress = getDefaultMAV2FactoryAddress(chain),
+    initCode,
+    initialOwner,
+    accountAddress,
+    entryPoint = getEntryPoint(chain, { version: "0.7.0" }),
+    isGlobalValidation = true,
+    entityId = 0,
+  } = config;
+
+  if (entityId >= Number(maxUint32)) {
```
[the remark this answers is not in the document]
howydev: good catch! I confused 2**32 and maxUint32, this should be an `>`

account-kit/smart-contracts/src/ma-v2/account/semiModularAccountV2.ts
howydev: > The size param doesn't truncate; it throws if the input is too long
TIL, thought it was like a `slice` haha

During review earlier, aligned with @moldy530 @adam-alchemy to move this to a `uint152 parallelNonce` 
input, appending the `uint40` suffix and using that as the `uint192` nonceKey to the EP. This abstracts the suffix complexity away from devs and they can use smaller numbers such as `0`, `1` for the nonce keys

account-kit/smart-contracts/src/ma-v2/index.ts
howydev: will remove here and populate in a diff PR

account-kit/smart-contracts/src/ma-v2/account/semiModularAccountV2.ts
```diff
@@ -0,0 +1,151 @@
...
+  createBundlerClient,
+  getEntryPoint,
+  toSmartContractAccount,
+} from "@aa-sdk/core";
+import {
+  concatHex,
+  encodeFunctionData,
+  getContract,
+  maxUint32,
+  toHex,
```
moldy530, line 17: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'toHex' is defined but never used.

account-kit/smart-contracts/src/ma-v2/account/semiModularAccountV2.ts
```diff
@@ -0,0 +1,151 @@
...
+  type Address,
+  type Chain,
+  type Hex,
+  type Transport,
+  hexToBigInt,
```
moldy530, line 22: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'hexToBigInt' is defined but never used.

account-kit/smart-contracts/src/ma-v2/modules/single-signer-validation/signer.ts
howydev: addressing in: https://github.com/alchemyplatform/aa-sdk/pull/1246

account-kit/smart-contracts/src/ma-v2/modules/single-signer-validation/module.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const SingleSignerValidationModule = {
+  encodeOnInstallData: (args: { entityId: number; signer: Address }): Hex => {
+    const { entityId, signer } = args;
+
+    return encodeAbiParameters(
+      [
+        {
+          type: "uint32",
+          value: entityId,
```
dphilipson, line 11: This `value` field isn't part of the API of `encodeAbiParameters`. Normally you would be passing a `name` and `type` field for each parameter (where I think the name is only used for error messages). See [Viem's docs](https://viem.sh/docs/abi/encodeAbiParameters).

I'm surprised this wasn't a type error, not sure why Viem defined the signature of `encodeAbiParameters` to allow `unknown` params.

account-kit/smart-contracts/src/ma-v2/modules/single-signer-validation/signer.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * const messageSigner = singleSignerMessageSigner(signer, chain);
+ * ```
+ *
+ * @param {TSigner} signer the signer to use for signing operations
+ * @returns {object} an object with methods for signing operations and managing signatures
+ */
+export const singleSignerMessageSigner = <TSigner extends SmartAccountSigner>(
+  signer: TSigner
+) => {
```
dphilipson, line 32: What's the type of the object being returned here? It's a `SmartAccountSigner<TSigner>`, right, but just missing `signerType` and `inner` fields? If so, we should explicitly state the return type in the function signature for better type-checking and easier readability and add the missing fields.

If you don't intend to annotate the return type and include an `inner` field, then you can instead simplify by not using a generic `TSigner`. The code under review is equivalent to the simpler code
```ts
export const singleSignerMessageSigner = (
  signer: SmartAccountSigner,
) => {
  // ...
}
```

account-kit/smart-contracts/src/ma-v2/modules/single-signer-validation/signer.ts
dphilipson: Are you saying this implementation is incorrect? If so, it would be better to throw a "not implemented" error than to return an incorrect value.

dphilipson (commented):
This looks pretty good! Some comments added.

account-kit/smart-contracts/src/ma-v2/account/semiModularAccountV2.ts
```diff
@@ -0,0 +1,154 @@
...
+
+export type SignerEntity = {
+  isGlobalValidation: boolean;
+  entityId: number;
+};
+
+export type SMAV2Account<
+  TSigner extends SmartAccountSigner = SmartAccountSigner
+> = SmartContractAccountWithSigner<"SMAV2Account", TSigner, "0.7.0"> &
+  SignerEntity;
```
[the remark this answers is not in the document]
howydev: That makes sense!

account-kit/smart-contracts/src/ma-v2/account/semiModularAccountV2.ts
```diff
@@ -0,0 +1,154 @@
...
+        entityId?: never;
+      }
+  );
+
+export async function createSMAV2Account<
+  TTransport extends Transport = Transport,
+  TSigner extends SmartAccountSigner = SmartAccountSigner
+>(
+  config: CreateSMAV2AccountParams<TTransport, TSigner>
+): Promise<SMAV2Account<TSigner>>;
```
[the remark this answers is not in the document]
howydev: yeah, seems like a convention - [example](https://github.com/alchemyplatform/aa-sdk/blob/main/account-kit/smart-contracts/src/msca/account/multiOwnerAccount.ts#L58-L88)

account-kit/smart-contracts/src/ma-v2/client/client.ts
```diff
@@ -0,0 +1,72 @@
...
+ *  transport,
+ * });
+ * ```
+ *
+ * @param {CreateSMAV2AccountClientParams} config The configuration parameters required to create the MAv2 account client
+ * @returns {Promise<SmartAccountClient>} A promise that resolves to a `SmartAccountClient` instance
+ */
+export async function createSMAV2AccountClient({
+  ...config
+}: CreateSMAV2AccountClientParams): Promise<SmartAccountClient> {
```
dphilipson, line 63: No need for that `{ ...config }`, this can just be
```ts
export async function createSMAV2AccountClient(
  config: CreateSMAV2AccountClientParams,
): Promise<SmartAccountClient> {
  // ...
```

account-kit/smart-contracts/src/ma-v2/client/client.ts
```diff
@@ -0,0 +1,72 @@
...
+  const maV2Account = await createSMAV2Account({
+    ...config,
+  });
```
dphilipson, line 66: Likewise, no need for this `{ ...config }`, can just be
```ts
const maV2Account = await createSMAV2Account(config);
```

dphilipson (dismissed):
Looks great! Minor style comments inline.

linnall (approved):
LGTM

account-kit/smart-contracts/src/ma-v2/client/client.test.ts
```diff
@@ -0,0 +1,196 @@
+import { custom, parseEther, publicActions } from "viem";
+import { LocalAccountSigner, type SmartAccountSigner } from "@aa-sdk/core";
+import { createSMAV2AccountClient, type SMAV2AccountClient } from "./client.js";
```
moldy530, line 3: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'SMAV2AccountClient' is defined but never used.

account-kit/smart-contracts/src/ma-v2/actions/install-validation/installValidation.ts
```diff
@@ -0,0 +1,160 @@
+import {
+  AccountNotFoundError,
+  IncompatibleClientError,
+  isSmartAccountClient,
+  EntityIdOverrideError,
```
moldy530, line 5: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'EntityIdOverrideError' is defined but never used.

account-kit/smart-contracts/src/ma-v2/actions/install-validation/installValidation.ts
```diff
@@ -0,0 +1,160 @@
...
+import type { HookConfig, ValidationConfig } from "../common/types.js";
+import {
+  serializeValidationConfig,
+  serializeHookConfig,
+  serializeModuleEntity,
+} from "../common/utils.js";
+
+import { type SMAV2AccountClient } from "../../client/client.js";
+import { type SMAV2Account } from "../../account/semiModularAccountV2.js";
+import { DEFAULT_OWNER_ENTITY_ID } from "../../utils.js";
```
moldy530, line 23: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'DEFAULT_OWNER_ENTITY_ID' is defined but never used.

Zer0dot (approved):

linnall (approved):

linnall (approved):

PR #1248, opened 2024-12-20
test(test-d): add test-d files for vitest typechecking:

Blu-J:
# Pull Request Checklist

---

## PR-Codex overview
This PR primarily focuses on enhancing type-checking and testing configurations across various packages, while also removing unnecessary code related to `AlchemyTransport`. It introduces new test files and updates existing tests to ensure better type safety and code quality.

### Detailed summary
- Added `test-d.ts` files for improved type-checking.
- Updated `package.json` files to exclude `.test.ts` and include `.test-d.ts`.
- Modified CI workflow to include type-checking tests.
- Enhanced `CONTRIBUTING.md` to reflect new testing practices.
- Removed `AlchemyTransport` related code from `client.test.ts`.
- Added new tests in `client.test-d.ts` for type assertions.


.github/workflows/on-pull-request.yml
```diff
@@ -102,3 +102,6 @@ jobs:
       - name: Unit Test
         run: yarn test:ci
 
+      - name: Typecheck Test
```
Blu-J, line 105: We want to make sure that during the run we test our new function

image

.vitest/vitest.shared.ts
```diff
@@ -3,6 +3,11 @@ import { configDefaults, defineConfig } from "vitest/config";
 
 export const sharedConfig = defineConfig({
   test: {
+    typecheck: {
+      enabled: process.env.TYPECHECK === "true",
```
Blu-J, line 7: So using a ENV because I should have been able to pass this as an argument to the vitest. 
After hours of it not working, I did this env and it just works. So 🤷‍♀️

.vitest/vitest.shared.ts
```diff
@@ -3,6 +3,11 @@ import { configDefaults, defineConfig } from "vitest/config";
...
+      only: process.env.TYPECHECK === "true",
+      ignoreSourceErrors: true,
```
Blu-J, line 9: When testing, files that just don't typecheck, ie the multisig account
Will fail this step without this set as true
image

.github/workflows/on-pull-request.yml
Blu-J: image

.vitest/vitest.shared.ts
moldy530: Ah we should just exclude this file from the test sets since later we'll delete this file. That way the other tests can at least get caught as well. We've had cases where code built no prob but it ruined some branch of the types and we didn't notice cuz the tests don't get type checked

moldy530 (dismissed):

.vitest/vitest.shared.ts
Blu-J: They already are excluded in the tests btw, just the typecheck will check non test files that are excluded.

moldy530 (approved):

PR #1251, opened 2024-12-20
fix: using enum instead of strings for auth step types

jcaleb4:
# Pull Request Checklist

---

## PR-Codex overview
This PR primarily focuses on refactoring the `authStep` handling in the authentication components by replacing string literals with an `AuthStepType` enum for improved type safety and maintainability.

### Detailed summary
- Introduced `AuthStepType` enum to replace string literals in `setAuthStep` calls.
- Updated multiple components to use `AuthStepType` for various authentication steps.
- Ensured consistent handling of `authStep` types across the application.


account-kit/react/src/components/auth/context.ts
```diff
@@ -4,34 +4,49 @@ import type { Connector } from "@wagmi/core";
...
+  passkey_verify = "passkey_verify",
+  passkey_create = "passkey_create",
+  passkey_create_success = "passkey_create_success",
+  email_completing = "email_completing",
+  oauth_completing = "oauth_completing",
+  initial = "initial",
+  complete = "complete",
+  eoa_connect = "eoa_connect",
+  wallet_connect = "wallet_connect",
+  pick_eoa = "pick_eoa",
```
moldy530, line 19: nit: for the enum name, can we do capital case? ie `EmailVerify`

moldy530 (commented):
What's the motivation for this change? we use string consts in other places of the code base and actually use enums fairly sparingly

account-kit/react/src/components/auth/context.ts
jcaleb4: change is done

jcaleb4:
@moldy530 my motivation for this PR is just that while working around the auth flow I felt the need to unify the source of the auth steps, I know it may be an opinionated change, but one of the advantages of using TS is that we can centralize these kind of things.

In my personal experience, using strings (specially in big teams) gets to be really messy at some point and takes out several advantages of using TS with enums:
- Type safety
- readability and self-documentation 
- maintainability
- intelisense support
- consistency 

I know there are several other places where you're using strings instead of enums, I just wanted to propose making this change because it makes sense to me; this is not required or anything, it is just me thinking that enums in this particular case would be a better use

moldy530:
I think I'd lean towards the string literal union types over enums. I've always found enums to be super annoying for one of the reasons called out [here](https://devsparks.goooseman.dev/hacks/20240715-use-string-literals-not-enums/), and that is the fact that once you want to use an enum somewhere else, you now have to import the enum and type as well. Whereas something like 
```ts
type AuthStep = "one" | "two" | "three"
```

will still get the same autocomplete and typesafety of the enum with a much better devex cuz I don't have to remember to export the enum as well.

moldy530:
closing this out as it's pretty stale and would rather we stay consistent using string literals

PR #1273, opened 2025-01-10
feat: add oauth support to react native signer

iykazrji:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing the OAuth functionality in the `account-kit` library, adding new error handling, utility functions, and improving the overall structure of the codebase for better maintainability and usability.

### Detailed summary
- Introduced `OAuthProvidersError` and `InAppBrowserUnavailableError` for error handling.
- Added `resolveRelativeUrl` utility function.
- Enhanced `oauthWithRedirect` and `oauthWithPopup` methods in `RNSignerClient`.
- Updated `RNAlchemySigner` to export `RNAlchemySignerType`.
- Added `parseSearchParams` utility function.
- Improved `base64UrlEncode` utility function.
- Added `addOpenIdIfAbsent` utility function to ensure "openid" scope.
- Updated `getOauthProviderUrl` method in `BaseSignerClient` for better OAuth provider handling.
- Refactored `MagicLinkAuthScreen` and `OTPAuthScreen` to use the new signer instance.
- Introduced `OAuthScreen` for OAuth authentication.
- Updated `tsconfig.json` to allow importing TypeScript extensions.


iykazrji:
* **#1273** <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/1273?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/alchemyplatform/aa-sdk/1273?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

## How to use the Graphite Merge Queue

Add the label <kbd>graphite-merge-queue</kbd> to this PR to add it to the merge queue.

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/alchemyplatform?ref=merge-queue-instructions-in-stack-comment&prId=1273).

<h2></h2>
<sub>

An organization admin has enabled the [Graphite Merge Queue](https://graphite.dev/docs/graphite-merge-queue) in this repository.

</sub>

<sub>

Please do not merge from GitHub as this will restart CI on PRs being processed by the merge queue.

</sub>

<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

account-kit/rn-signer/src/oauth.ts
moldy530, line 1: I think these are already defined in account-kit/signer, we should just re-export them from there

account-kit/rn-signer/src/errors.ts
moldy530, line 1: are any of these already defined in account-kit/signer?

account-kit/rn-signer/src/client.ts
```diff
@@ -116,15 +145,61 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
...
-    throw new Error("Method not implemented.");
-  }
+  ): Promise<User> => {
+    try {
+      const providerUrl = await this.getOauthProviderUrl(_args);
+      const redirectUrl = _args.redirectUrl;
+
+      return new Promise(async (resolve, reject) => {
+        if (await InAppBrowser.isAvailable()) {
+          const res = await InAppBrowser.openAuth(providerUrl, redirectUrl);
```
iykazrji, line 157: We actually can await the result of the redirect operation and use it to complete the auth instead of calling `authenticate(type: "oauthReturn")`.

moldy530: woah...

account-kit/signer/src/client/base.ts
```diff
@@ -139,7 +139,7 @@ export abstract class BaseSignerClient<TExportWalletParams = unknown> {
 
   public abstract oauthWithRedirect(
     args: Extract<OauthParams, { mode: "redirect" }>
-  ): Promise<never>;
+  ): Promise<User | never>;
```
moldy530, line 142: discussed offline, but we should avoid this if possible

account-kit/rn-signer/src/client.ts
moldy530: ah ok this function makes a lot of sense! what I'd do though, is put this impl in the `oauthWithPopup` method in this file (and still throw undefined here for now maybe?)

In the `signer.ts` file, I would just support "pop-up" auth. That way the flow for doing oauth on mobile matches doing oauth on web (when pop-up you don't have to do oauth return, and when it's redirect you do). Since we don't have to do the oauth return, then oauth on mobile is actually a pop-up flow

account-kit/rn-signer/src/client.ts
```diff
@@ -1,40 +1,75 @@
...
+    const { connection, rootOrgId, oauthCallbackUrl } =
+      RNSignerClientParamsSchema.parse(params);
 
     super({
       stamper: NativeTEKStamper,
       rootOrgId: rootOrgId ?? "24c1acf5-810f-41e0-a503-d5d13fa8e830",
       connection,
     });
+
+    this.oauthCallbackUrl = oauthCallbackUrl;
```
moldy530, line 72: how is this used? This is generally configured on the server, so I'm curious how this works here.

account-kit/rn-signer/src/client.ts
```diff
@@ -116,15 +148,61 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
     this.eventEmitter.emit(params.connectedEventName, user, params.bundle);
     return user;
   }
-  override oauthWithRedirect(
+  override oauthWithRedirect = async (
     _args: Extract<OauthParams, { mode: "redirect" }>
```
moldy530, line 152: nit: `_args` implies the param isn't used. Since now it is, let's change it to `args`

account-kit/rn-signer/src/client.ts
```diff
@@ -116,15 +148,61 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
...
+              }
+              resolve(user);
+            }, reject);
+          } else if (error) {
+            reject(new OauthFailedError(error));
+          }
+        } else {
+          throw new Error(
+            "Couldn't perfom authentication operation. Browser not available."
+          );
```
moldy530, line 194: nit: let's add an `errors.ts` file in `rn-signer/src` and in it define strongly typed errors that extends `BaseError` exported from `aa-sdk/core`.

That will allow us to link to specific docs based on the error being thrown

account-kit/rn-signer/src/client.ts
```diff
@@ -138,13 +216,130 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
...
-  }
+
   protected override getWebAuthnAttestation(
     _options: CredentialCreationOptions,
     _userDetails?: { username: string }
   ): Promise<GetWebAuthnAttestationResult> {
     throw new Error("Method not implemented.");
   }
+
+  public getOauthProviderUrl = async (args: OauthParams): Promise<string> => {
```
moldy530, line 227: @dphilipson can you take a pass at this too?

AFAIK a lot of this is actually done by the signer service right?

moldy530 (commented):
This is lookin great!

Had some comments about code organization (basically oauth pop-up seems more like what the flow happening here is vs redirect)

Also would be good to get @dphilipson's eyes on the `getOauthProviderUrl` method since I think we do a lot of this elsewhere

account-kit/rn-signer/src/client.ts
iykazrji: @moldy530 on this, the signer service implementation relies on some browser apis that aren't supported on mobile. This is why we're re-implementing this for mobile

iykazrji:
> ah ok this function makes a lot of sense! what I'd do though, is put this impl in the oauthWithPopup method in this file (and still throw undefined here for now maybe?)

In the signer.ts file, I would just support "pop-up" auth. That way the flow for doing oauth on mobile matches doing oauth on web (when pop-up you don't have to do oauth return, and when it's redirect you do). Since we don't have to do the oauth return, then oauth on mobile is actually a pop-up flow

@moldy530 so the issue with doing this is the 'pop-up' auth mode doesn't accept the `redirect_url` field which is necessary to complete the authentication when calling `.authenticate()`. Will we be willing to adjust the type to accept a `redirect_url` to deal with this?

account-kit/rn-signer/src/client.ts
moldy530: ohhh ok, just found it. can we supply polyfills for those methods as part of our library? or does that not work?

basically just in the `index.ts` of the rn-signer assigning stuff like `globals.thingOnWeb = rnImpl`?

moldy530:
@iykazrji for the redirect_url, how does that work exactly with the `await InAppBrowser`? is the InAppBrowser configured to return when the page redirects to deeplink in the app?

iykazrji:
@moldy530 Yeah so after auth is complete, the auth provider sends the user back the specified `redirect_url`. The simply InAppBrowser, awaits this return and processes the response for us to then use to complete the authentication

account-kit/rn-signer/src/client.ts
moldy530: nevermind, found it!

moldy530:
ok cool, I think I see it now! yea so redirect makes sense here and the change to `User | never` is on the client impl not the signer impl, so shouldn't have upstream effects since the build still work on the web code which restricts the return type to `never` anways

moldy530:
aight yea this is lookin good to me! only open question is if we can consolidate the impl for `getOauthProviderUrl` so that we can minimize the amount of duplication

iykazrji:
Gotcha, would be looking into that!

account-kit/rn-signer/src/utils/getDeepLink.ts
```diff
@@ -0,0 +1,7 @@
+import { Platform } from "react-native";
+export const getDeepLink = (path = "") => {
+  const scheme = "rn-signer-demo";
+  const prefix =
+    Platform.OS === "android" ? `${scheme}://my-host/` : `${scheme}://`;
```
dphilipson, line 5: Is it intentional to be hardcoding "my-host" into this link?

account-kit/signer/src/client/types.ts
```diff
@@ -177,3 +181,21 @@ export type GetWebAuthnAttestationResult = {
...
+  redirectUrl?: string;
+  openerOrigin?: string;
+};
+
+export type GetOauthProviderUrlArgs = {
+  oauthParams: OauthParams;
+  turnkeyPublicKey: string;
+  oauthCallbackUrl: string;
+  oauthConfig?: OauthConfig;
+  usesRelativeUrl?: boolean;
```
dphilipson, line 200: It feels to me like we shouldn't need this to be passed explicitly. Seems like we should be able to tell if a url is relative by looking at it (for example, by checking if it contains "://").

dphilipson (commented):
This looks good! Left some initial thoughts.

account-kit/rn-signer/src/utils/getDeepLink.ts
iykazrji: Thanks for catching this! This function actually isn't being used and was meant to be removed.

account-kit/signer/src/client/types.ts
iykazrji: So the reason we're explicitly specifying this is to adapt the implementation to work in RN. It seems we are converting the redirect `url` to a relative url on web (using the `resolveRelativeURL()` in the `getOauthProviderUrl()` method), which we don't want in RN as we wouldn't be able to deep link properly.

account-kit/signer/src/client/base.ts
```diff
@@ -471,6 +479,158 @@ export abstract class BaseSignerClient<TExportWalletParams = unknown> {
...
+   *  turnkeyPublicKey,
+   *  oauthCallbackUrl
+   * })
+   *
+   * ```
+   * @param {GetOauthProviderUrlArgs} args Required. The Oauth provider's auth parameters
+   *
+   * @returns {Promise<string>} returns the Oauth provider's url
+   */
+  public getOauthProviderUrl = async (
```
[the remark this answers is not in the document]
iykazrji: We should be good using `protected` here tbh. I'd make the update!

account-kit/rn-signer/src/index.tsx
```diff
@@ -1,2 +1,3 @@
-// eslint-disable-next-line import/extensions
+/* eslint-disable import/extensions */
```
moldy530, line 1: we shouldn't be ignoring these. I'm actually shocked it's building with these 🤔 

does it not work if these are `... from "./signer.js"`

moldy530 (commented):
couple small comments, but generally LGTM

account-kit/rn-signer/src/index.tsx
iykazrji: Yeah, so specifying the extension seems to throw Metro off and prevents the example apps from building, hence we removed the extension from this export.

site/pages/reference/account-kit/infra/functions/alchemyGasManagerMiddleware.mdx
```diff
@@ -18,13 +18,13 @@ import { alchemyGasManagerMiddleware } from "@account-kit/infra";
...
 
 ```ts
-import { sepolia, alchemyGasManagerMiddleware } from "@account-kit/infra";
+import { sepolia, alchemyErc7677Middleware } from "@account-kit/infra";
 import { http } from "viem";
 
 const client = createSmartAccountClient({
   transport: http("rpc-url"),
   chain: sepolia,
-  ...alchemyGasManagerMiddleware("policyId"),
```
moldy530, line 27: odd... this probably shouldn't have changed, I wonder what happened here?

oos for this though

moldy530 (dismissed):

account-kit/rn-signer/src/index.tsx
iykazrji: Using `.ts` works actually! Pushing an update to address this. Thanks for the heads up.

moldy530 (approved):

PR #1316, opened 2025-01-31
 fix the link for SmartAccountSigner.

Noisyxl:
This PR updates the documentation to fix the link for SmartAccountSigner.
Changed the link to point to the current documentation in smart-account-signer.mdx.
This ensures users have easier access to information about SmartAccountSigner exported from @aa-sdk/core.

---

## PR-Codex overview
This PR updates the documentation for using a third-party library as a Signer with Account Kit, specifically correcting the link to the `SmartAccountSigner` implementation.

### Detailed summary
- Corrected the link to `SmartAccountSigner` implementation from `.../main/packages/core/src/...` to `.../main/aa-sdk/core/src/...` in the documentation.

### DIFF
--- a/site/pages/third-party/signers.mdx
+++ b/site/pages/third-party/signers.mdx
@@ -8,7 +8,7 @@ description: Learn how to use a 3rd party Signer with Account Kit
 Account Kit is designed to be flexible and allow you to use any Signer you want. If you choose not use our signer, you can either:
 
 1. Use a 3rd party library as a Signer that integrates with Account Kit.
-1. Implement [`SmartAccountSigner`]( https://github.com/alchemyplatform/aa-sdk/blob/main/aa-sdk/core/src/signer/types.ts) (exported in `@aa-sdk/core`).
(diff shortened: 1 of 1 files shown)


moldy530 (changes requested):
What's the motivation for this change? If we're pointing folks to .mdx files that means we should just link to that portion of the site. Linking to the type in code shows the exact interface to implement. 

If the concepts doc already does this, then we should update this link to match the url route on the site and not send people to GitHub 

EDIT: oh I see the code reference this doc links to is broken because it points to a wrong package location. The link should be updated to point to https://github.com/alchemyplatform/aa-sdk/blob/main/aa-sdk/core/src/signer/types.ts

Noisyxl:
> What's the motivation for this change? If we're pointing folks to .mdx files that means we should just link to that portion of the site. Linking to the type in code shows the exact interface to implement.
> 
> If the concepts doc already does this, then we should update this link to match the url route on the site and not send people to GitHub
> 
> EDIT: oh I see the code reference this doc links to is broken because it points to a wrong package location. The link should be updated to point to https://github.com/alchemyplatform/aa-sdk/blob/main/aa-sdk/core/src/signer/types.ts

done

PR #1329, opened 2025-02-06
fix: Fix MAv2 React Hook client creation for 7702

jakehobbs:
# Pull Request Checklist


account-kit/core/src/actions/createAccount.ts
```diff
@@ -45,9 +45,11 @@ export type AccountConfig<TAccount extends SupportedAccountTypes> =
     ? OmitSignerTransportChain<
         CreateMultiOwnerModularAccountParams<Transport, AlchemyWebSigner>
       >
-    : OmitSignerTransportChain<
+    : TAccount extends "ModularAccountV2"
+    ? OmitSignerTransportChain<
         CreateModularAccountV2Params<Transport, AlchemyWebSigner>
-      >;
+      >
+    : never;
```
adamegyed, line 52: Just so I understand the old behavior: previously, it would fall into the MAv2 branch as a default branch, and now we're asserting that the user has to provide "ModularAccountV2" as a mode?

account-kit/core/src/actions/getSmartAccountClient.ts
```diff
@@ -218,12 +220,23 @@ export function getSmartAccountClient(
           isLoadingClient: false,
         };
       case "ModularAccountV2":
+        const mode = (accountParams as AccountConfig<"ModularAccountV2">).mode;
```
adamegyed, line 223: If we avoid destructuring `accountParams` earlier, can we avoid the `as` casting here? It should propagate the type inference as being of `"ModularAccountV2:` thought the generic args.

adamegyed (commented):
Looks good! Had a few small questions.

Also, since this PR is not touching code in the `examples/ui-demo/` folder, could you set the base as `adam/7702-core`? That way we can merge it in before demo liveness.

account-kit/core/src/actions/createAccount.ts
jakehobbs: This change isn't necessary. I was just trying to make it more explicit.

adamegyed: It makes sense! Just wanted to clarify

account-kit/core/src/actions/getSmartAccountClient.ts
jakehobbs: been fighting this for a bit but cannot seem to make it infer properly

jakehobbs: pushed a change to make it a bit safer at least

jakehobbs:
> Looks good! Had a few small questions.
> 
> Also, since this PR is not touching code in the `examples/ui-demo/` folder, could you set the base as `adam/7702-core`? That way we can merge it in before demo liveness.

i changed the base, also fixed an issue w/ MAv2 accounts not reconnecting after page refresh.

jakehobbs:
@adamegyed just fixed docs too fyi

adamegyed (approved):
Looks good!

PR #1338, opened 2025-02-09
fix: make type really optional, add defaults for return type

howydev:
# Pull Request Checklist

---

## PR-Codex overview
This PR modifies the `useSmartAccountClient` hook in the `account-kit/react/src/hooks/useSmartAccountClient.ts` file to improve type handling and default values for account types.

### Detailed summary
- Added `OptionalFields` to `GetSmartAccountClientParams`.
- Changed the default type of `TAccount` to `"ModularAccountV2"`.
- Updated the `type` parameter in the function signature to default to `"ModularAccountV2"`.

### DIFF
--- a/account-kit/react/src/hooks/useSmartAccountClient.ts
+++ b/account-kit/react/src/hooks/useSmartAccountClient.ts
@@ -1,5 +1,5 @@
 "use client";
-
+import { type OptionalFields } from "@aa-sdk/core";
 import type {
   GetSmartAccountClientParams,
   GetSmartAccountClientResult,
@@ -21,9 +21,12 @@ export type UseSmartAccountClientProps<
   TAccount extends SupportedAccountTypes | undefined =
     | SupportedAccountTypes
     | undefined
-> = GetSmartAccountClientParams<
-  TChain,
-  TAccount extends undefined ? "ModularAccountV2" : TAccount
+> = OptionalFields<
+  GetSmartAccountClientParams<
+    TChain,
+    TAccount extends undefined ? "ModularAccountV2" : TAccount
+  >,
+  "type"
 >;
 
 export type UseSmartAccountClientResult<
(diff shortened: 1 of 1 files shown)


account-kit/react/src/hooks/useSmartAccountClient.ts
```diff
@@ -21,9 +21,12 @@ export type UseSmartAccountClientProps<
...
     | undefined
-> = GetSmartAccountClientParams<
-  TChain,
-  TAccount extends undefined ? "ModularAccountV2" : TAccount
+> = OptionalFields<
+  GetSmartAccountClientParams<
+    TChain,
+    TAccount extends undefined ? "ModularAccountV2" : TAccount
+  >,
+  "type"
```
adamegyed, line 29: should `"type"` here be `mode`? There's actually another inner parameter to a `client` constructor called `type` that's just an inner string field, which makes any uses of `type` after the switch to `mode` still compile.

adamegyed: related: if `GetSmartAccountClientParams` already has `mode` as optional from https://github.com/alchemyplatform/aa-sdk/pull/1335/files, then I think we can skip using `OptionalFields<...>` here

howydev: we're doing both optional - the client level `type` defaults to MAv2 and the account level `mode` defaults to default

howydev: with this, `useSmartAccountClient({})` becomes valid syntax

adamegyed: Should we be overriding `type`? I thought we needed the client type to be `alchemy` for the alchemy gas estimator logic to get correctly used?

howydev: related: https://github.com/alchemyplatform/aa-sdk/pull/1328/files

adamegyed: oh nevermind, I think that's for `.type` within the transport passed to the client.

howydev: yeah, that seems to be `transport.config.type`

adamegyed (approved):
Ah, I think I just understood the relationship between the default in the generic and the default-initialized param - this looks good!

account-kit/react/src/hooks/useSmartAccountClient.ts
howydev: im so glad we changed the account `type` to `mode` lol

PR #1356, opened 2025-02-12
fix: only reconnect accounts for active chain during rehydration

jakehobbs:
# Bug when switching chains after refreshing page

## Explanation
- When the page is refreshed, the store is rehydrated using the cookie data.
- The `hydrate` function calls `hydrateAccountState`, which reads all existing accounts from the store and sets them to "RECONNECTING" on **all chains**.
- The `hydrate` function then calls the `reconnect` function, which calls `createAccount` for each account. **But this only happens on the active chain**, so any other chain's accounts will get stuck in the "RECONNECTING" state but never actually reconnect.

## Solutions
- The easiest solution is to not set accounts to "RECONNECTING" if they are not on the active chain during hydration. We can just set them to "DISCONNECTED", then they will naturally reconnect if they are attempted to be used again.
- An alternative solution would be to alter the `createAccount` function to accept an option `chain` param instead of only being able to reconnect accounts on the active chain.

# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing the account state management by introducing an `activeChainId` that determines if the current chain is active, affecting the reconnection logic for various account types.

### Detailed summary
- Added `activeChainId` to track the current active chain.
- Introduced `isActiveChain` to check if the current chain matches `activeChainId`.
- Updated the reconnection logic for `LightAccount`, `MultiOwnerModularAccount`, `MultiOwnerLightAccount`, and `ModularAccountV2` to consider `shouldReconnect` based on the active chain.

### DIFF
--- a/account-kit/core/src/hydrate.ts
+++ b/account-kit/core/src/hydrate.ts
@@ -110,25 +110,27 @@ const hydrateAccountState = (
     ([, cnx]) => cnx.chain
   );
   const initialState = createDefaultAccountState(chains);
+  const activeChainId = config.store.getState().chain.id;
 
   return Object.entries(accountConfigs).reduce((acc, [chainKey, config]) => {
     const chainId = Number(chainKey);
+    const isActiveChain = chainId === activeChainId;
+    const shouldReconnect = shouldReconnectAccounts && isActiveChain;
     acc[chainId] = {
       LightAccount:
-        config.LightAccount?.accountAddress && shouldReconnectAccounts
+        shouldReconnect && config.LightAccount?.accountAddress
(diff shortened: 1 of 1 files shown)


adamegyed (approved):
Good job digging through the stack to find this!

Do you know if `setChain` from `useChain` will automatically start the re-connect process? https://accountkit.alchemy.com/react/multi-chain-apps#change-chains

jakehobbs:
> Good job digging through the stack to find this!
> 
> Do you know if `setChain` from `useChain` will automatically start the re-connect process? https://accountkit.alchemy.com/react/multi-chain-apps#change-chains

it does not. the only place `reconnect` is called is during rehydration.

PR #1401, opened 2025-02-28
feat: improve call efficiency by removing execute wrapper for calls to the account

howydev:
Currently, all calls go through `sendUserOperation` that turns the call into an `execute`. This results in a redundant wrapping when calls are made to the account directly. We change `encodeExecute` thats used in `sendUserOperation` to format the calls to not add the wrapping when the call target is the account

# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on simplifying the `sendUserOperation` function call in the `installValidation` method by removing the `account` parameter from the request.

### Detailed summary
- Removed the `account` parameter from the `client.sendUserOperation` call in the `installValidation` method.
- Kept the `uo` and `overrides` parameters in the `sendUserOperation` call.

### DIFF
--- a/account-kit/smart-contracts/src/ma-v2/actions/install-validation/installValidation.ts
+++ b/account-kit/smart-contracts/src/ma-v2/actions/install-validation/installValidation.ts
@@ -219,11 +219,7 @@ export const installValidationActions: <
         account,
       });
 
-      return client.sendUserOperation({
-        uo: callData,
-        account,
-        overrides,
-      });
+      return client.sendUserOperation({ uo: callData, overrides });
     },
 
     uninstallValidation: async ({
@@ -242,11 +238,7 @@ export const installValidationActions: <
         account,
       });
 
-      return client.sendUserOperation({
-        uo: callData,
-        account,
-        overrides,
-      });
+      return client.sendUserOperation({ uo: callData, overrides });
     },
   };
 };


adamegyed (commented):
I _think_ this makes sense to do, but I'm not totally sure. It is already possible for an account to specify the raw calldata on an account, because the parameters to `sendUserOperation` are either:
- raw calldata (toward the account, from the EP)
- a single instance of `{ target, value, data }`
- an array of `{ target, value, data }`.

Also, for MAv2, in the case where someone tries to use the non-pure-calldata form of `execute` towards self, the call will fail with `SelfCallRecursionDepthExceeded`: https://github.com/alchemyplatform/modular-account/blob/v2.0.0/src/account/ModularAccountBase.sol#L890-L894

So this change would allow people to specify a call as `{ target: ≤self>, data: callData, value?: <some_value_ignored> }`, instead of just as `<callData>`. This might help people with debugging, but it could also be unexpected.

Wdyt?

howydev:
oh TIL on the type - missed it. sendUserOp calls buildUserOp that calls _initUserOp that does this with the calldata: https://github.com/alchemyplatform/aa-sdk/blob/main/aa-sdk/core/src/actions/smartAccount/internal/initUserOperation.ts#L60

so this change is not needed for execute, but would be needed for the `installValidation` calls instead

PR #1441, opened 2025-03-12
feat: pull email mode from server

florrdv:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on deprecating the `emailMode` option in the authentication configuration, simplifying the email authentication process, and introducing a new configuration management for the signer.

### Detailed summary
- Removed `emailMode` from email authentication sections.
- Deprecated `emailMode` in various types with a warning for future removal.
- Introduced `SignerConfig` and `EmailConfig` types for better configuration management.
- Updated `EmailAuth` component to handle new configuration logic.
- Added methods for fetching signer configuration.


jakehobbs (commented):
lgtm, is the signer service ready to handle this in prod?

account-kit/signer/src/base.ts
```diff
@@ -75,6 +83,7 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
   inner: TClient;
   private sessionManager: SessionManager;
   private store: InternalStore;
+  private config: SignerConfig | undefined;
```
moldy530, line 86: should this instead just be `Promise<SignerConfig>`? we await it anyways in the auth components

moldy530: that way it's never null since that doesn't seem like a state it should be in anyways

account-kit/react/src/components/auth/sections/EmailAuth.tsx
```diff
@@ -28,8 +27,9 @@ export const EmailAuth = memo(
     const signer = useSigner();
     const { authenticateAsync, isPending } = useAuthenticate({
       onMutate: async (params) => {
+        const cfg = await signer?.getConfig();
         if (params.type === "email" && "email" in params) {
-          if (params.emailMode === "magicLink") {
+          if (cfg?.email.mode === "MAGIC_LINK") {
```
moldy530, line 32: we might want to fallback to the user provided one if it exists here still, at least for backwards compat. 

If it's set in the dashboard, then we should ignore the user provided one here (and elsewhere)

moldy530 (commented):
LGTM mostly nits. I'd approve if it wasn't gonna get dismissed when you rebase lol

account-kit/react/src/components/auth/sections/EmailAuth.tsx
florrdv: Can do! We can remove all that logic next major version then. Let me make those changes

florrdv:
> lgtm, is the signer service ready to handle this in prod? also what about the switch b/w the two email modes on the demo app?

pending signer upgrade to deploy. discussed with will & we won't be adding support to demo app to avoid confusion for now at least! have another PR up to add those changes if we ever want to go back on that

account-kit/signer/src/base.ts
florrdv: fixed!

account-kit/signer/src/base.ts
```diff
@@ -113,6 +122,7 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
     this.registerListeners();
     // then initialize so that we can catch those events
     this.sessionManager.initialize();
+    this.config = this.fetchConfig();
```
florrdv, line 125: TS can't infer that we set config here if we use initConfig sadly

account-kit/react/src/components/auth/sections/EmailAuth.tsx
```diff
@@ -28,8 +28,15 @@ export const EmailAuth = memo(
...
     const { authenticateAsync, isPending } = useAuthenticate({
       onMutate: async (params) => {
+        const cfg = await signer?.getConfig();
         if (params.type === "email" && "email" in params) {
-          if (params.emailMode === "magicLink") {
+          const emailMode = legacyEmailMode
+            ? legacyEmailMode === "magicLink"
+              ? "MAGIC_LINK"
+              : "OTP"
+            : cfg?.email.mode;
```
moldy530, line 37: nit: I would probably swap the order of priority here to this since if its present on the server and legacyMode doesn't match you're gonna get an error anyways when you call the server

```suggestion
          const emailMode = cfg?.email.mode ? cfg.email.mode : legacyEmailMode
            ? legacyEmailMode === "magicLink"
              ? "MAGIC_LINK"
              : "OTP";
```

account-kit/signer/src/base.ts
moldy530: oh interesting

jakehobbs (approved):

PR #1443, opened 2025-03-13
feat: add twitch integration

dphilipson:
Also allow custom parameters to be passed to the auth URL as well as setting default custom parameters for certain providers, which is used for Twitch to set `force_verify=true`.

---

## PR-Codex overview
This PR introduces support for the `Twitch` authentication provider across various components and configurations, enhancing the authentication options available in the application.

### Detailed summary
- Added `twitch` support in `examples/ui-demo/src/app/config.tsx`.
- Imported and exported `TwitchIcon` in `auth-icons/index.ts`.
- Integrated `TwitchIcon` in `OAuth.tsx` for authentication buttons.
- Updated `KnownAuthProvider` to include `twitch`.
- Enhanced `ContinueWithOAuth` to display `TwitchIcon`.
- Created `TwitchIcon` component in `icons/twitch.tsx`.
- Defined `setAddTwitchAuth` function in `Authentication.tsx` for toggling Twitch auth.
- Updated `DEFAULT_PROVIDER_CUSTOMIZATION` to include `twitch` settings.
- Refactored `getDefaultProviderCustomization` to return customization parameters.
- Adjusted `BaseSignerClient` to handle `otherParameters` for `twitch`.


account-kit/signer/src/oauth.ts
```diff
@@ -1,27 +1,36 @@
...
 
-export type ScopeAndClaims = {
+export type AuthProviderCustomization = {
   scope: string;
   claims?: string;
+  otherParameters?: Record<string, string>;
 };
 
-const DEFAULT_SCOPE_AND_CLAIMS: Record<KnownAuthProvider, ScopeAndClaims> = {
+const DEFAULT_PROVIDER_CUSTOMIZATION: Record<
```
jakehobbs, line 9: should we move the facebook `joey` thing into here now that it supports other parameters than just scope & claims? i'm fairly certain that it's safe to always include it even if using `redirect` mode instead of `popup`.

account-kit/react/src/icons/auth-icons/twitch.tsx
```diff
@@ -0,0 +1,19 @@
+import type { SVGProps } from "react";
+
+export const TwitchIcon = (
```
jakehobbs, line 3: nit: should we have an icon that is filled in w/ white instead of being transparent so it looks a little nicer in dark mode?

how this icon looks:

Screenshot 2025-03-12 at 5 21 57 PM

example from twitch's website in dark mode:

Screenshot 2025-03-12 at 5 21 48 PM

jakehobbs (dismissed):

account-kit/signer/src/oauth.ts
dphilipson: Oh yea, that's a good idea. I had no idea what "joey" meant so I didn't want to touch it.

jakehobbs: you don't want to know 😆

jakehobbs: tl;dr facebook's oauth flow sets a COOP header that wipes out `window.opener` on mobile, so our auth flow fails. setting this `joey` param causes FB to use an older version of their oauth flow on mobile that is more well-behaved... i discovered this by snooping around w/ how FB's own js sdk initiates the oauth flow.

account-kit/react/src/icons/auth-icons/twitch.tsx
dphilipson: Looks like this now:
image

jakehobbs (approved):
lgtm!

moldy530 (approved):

PR #1483, opened 2025-03-25
DO NOT MERGE: new auth method

blakecduncan:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing the authentication flow by introducing support for multi-factor authentication (MFA) and refining the handling of different authentication steps in the `BaseAlchemySigner` class.

### Detailed summary
- Added `AuthStepResult` type to represent various authentication states.
- Extended `AuthParams` type to include MFA parameters.
- Implemented `authenticateStep` method to handle different auth types, including MFA.
- Created `authenticateWithOtpNew` and `authenticateWithEmailNew` methods for improved OTP and email handling.
- Updated error handling for unsupported auth step types.
- Modified `validateMultiFactors` to return `AuthStepResult` instead of `User`.

# Goal:

We’ve updated our API to improve multi-factor authentication (MFA) flows by adding a new validate-mfa endpoint. Previously, you had to submit all MFA credentials (e.g., TOTP) with your primary credentials (like email + otp) in a single call, which meant you couldn’t provide incremental feedback to the user if one step failed.

Now that we have a new endpoint for MFA, it makes sense to refine the SDK’s authentication flow by introducing a authenticateStep method that returns a status after each step—rather than always returning a final User or throwing an error. This is especially important for multi-step flows like email + otp + totp.

Currently, the existing authenticate method returns a User or throws an error. This approach can lead to unresolved Promises across multiple steps in an MFA flow, forcing developers to rely on events to figure out which step the user is on.

## Possible MFA flow with the current `authenticate` method

```typescript
// 1) Returns a promise that won’t resolve until the user eventually enters the OTP and TOTP:
const user1 = signer.authenticate({ type: "email", email });
// The developer must listen for an event to know we’re “awaiting OTP.” The promise remains unresolved.

// 2) Similarly, returns a promise that only resolves after TOTP is submitted:
const user2 = signer.authenticate({ type: "otp", otpCode });
// Developer must again rely on events to detect “awaiting MFA.”

// 3) Finally resolves with a user:
const user3 = signer.authenticate({ type: "mfa", mode: "totp", multiFactorCode: "asdf" });
// At the end, user1, user2, and user3 all resolve to the same final user.
```
This sequence can be confusing because each call to authenticate returns a promise that waits for future steps, rather than returning partial information at each step.

## Proposed Flow with `authenticateStep`

```typescript
// 1) Immediate return with `AWAITING_EMAIL_AUTH` but no final user:
const { status, user } = signer.authenticateStep({
  type: "email",
  email,
});
// status = AWAITING_EMAIL_AUTH
// user = undefined or "partial" if you choose

// 2) Next step returns `AWAITING_MFA_AUTH` along with the allowed MFA factors:
const { status, user, multiFactors } = signer.authenticateStep({
  type: "otp",
  otpCode,
});
// status = AWAITING_MFA_AUTH
// multiFactors = e.g., totp
// user = undefined or partial

// 3) Finally returns a fully authenticated user:
const { status, user } = signer.authenticateStep({
  type: "mfa",
  mode: "totp",
  multiFactorCode: "asdf",
});
// status = CONNECTED
// user = the final, fully authenticated user
```
Each call returns immediate feedback about the current status (AWAITING_EMAIL_AUTH, AWAITING_MFA_AUTH, etc.) rather than suspending the promise until all steps are complete. This means you can handle errors or partial progress more cleanly.

Note: We won’t break backward compatibility. The existing authenticate method continues to work as before, but we recommend switching to authenticateStep for multi-step flows.

### DIFF
--- a/account-kit/signer/src/base.ts
+++ b/account-kit/signer/src/base.ts
@@ -287,6 +287,10 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
             return this.handleOauthReturn(params);
           case "otp":
             return this.authenticateWithOtp(params);
+          case "mfa":
+            throw new Error(
+              `Unsupported auth step type: ${type}, please use the authenticateStep method.`
+            );
           default:
             assertNever(type, `Unknown auth type: ${type}`);
         }
@@ -323,10 +327,14 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
               return this.authenticateWithEmailNew(params);
             case "otp":
               return this.authenticateWithOtpNew(params);
+            case "mfa":
+              return this.validateMultiFactors(params);
             case "passkey":
             case "oauth":
             case "oauthReturn":
-              throw new Error(`Unsupported auth step type: ${type}`);
+              throw new Error(
+                `Unsupported auth step type: ${type}, please use the authenticate method.`
+              );
             default:
               assertNever(type, `Unknown auth step type: ${type}`);
           }
@@ -382,6 +390,8 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
           name: "signer_authnticate",
           data: { authType: "otp" },
         });
+        break;
+      case "mfa":
         break;
       default:
         assertNever(type, `Unknown auth type: ${type}`);
@@ -1475,7 +1485,7 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
    */
   public async validateMultiFactors(
     params: ValidateMultiFactorsParams
-  ): Promise<User> {
+  ): Promise<AuthStepResult> {
     // Get MFA context from temporary session
     const tempSession = this.sessionManager.getTemporarySession();
     if (
@@ -1530,7 +1540,7 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
       },
     });
 
-    return user;
+    return { user, status: AlchemySignerStatus.CONNECTED };
   }
 
   protected initConfig = async (): Promise<SignerConfig> => {
--- a/account-kit/signer/src/base.ts
+++ b/account-kit/signer/src/base.ts
@@ -40,7 +40,7 @@ import {
   type SessionManagerParams,
 } from "./session/manager.js";
 import type { SessionManagerEvents } from "./session/types";
-import type { AuthParams } from "./signer";
+import type { AuthParams, AuthStepResult } from "./signer";
 import { SolanaSigner } from "./solanaSigner.js";
 import {
   AlchemySignerStatus,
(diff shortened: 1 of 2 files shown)


account-kit/signer/src/base.ts
```diff
@@ -836,6 +864,53 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
...
+
+    const { orgId, otpId, multiFactors, isNewUser } =
+      await this.initOrCreateEmailUser(
+        params.email,
+        params.emailMode ?? "otp",
+        params.multiFactors,
+        params.redirectParams
+      );
+
+    const isMfaRequired = multiFactors ? multiFactors?.length > 0 : false;
```
dphilipson, line 890: (Very nit) No need for the `?.`, we know `multiFactors` is present in that branch.

account-kit/signer/src/base.ts
```diff
@@ -312,6 +312,34 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
...
+      async (params) => {
+        const { type } = params;
+        const result = (() => {
+          switch (type) {
+            case "email":
+              return this.authenticateWithEmailNew(params);
+            case "otp":
+              return this.authenticateWithOtpNew(params);
+            default:
+              throw new Error("type not implemented");
```
dphilipson, line 327: (Minor) I'd recommend using `assertNever` to check that all types are handled, which also means explicitly writing out the other cases. That way, if we add a new type of auth step the type system will make us explicitly acknowledge it here. I'm thinking something like:
```ts
          switch (type) {
            case "email":
              return this.authenticateWithEmailNew(params);
            case "otp":
              return this.authenticateWithOtpNew(params);
            case "passkey":
            case "oauth":
            case "oauthReturn":
              throw new Error(`Unsupported auth step type: ${type}`);
            default:
              assertNever(type, `Unknown auth step type: ${type}`);
          }
```
Over time we've had a number of issues in the sdk where we added a new union arm and forgot to add handling for it somewhere, so IMO it's worth the extra hassle.

(I also suggest including the type in the error message to make debugging easier if this happens.)

account-kit/signer/src/base.ts
```diff
@@ -836,6 +864,53 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
...
+  ): Promise<AuthStepResult> => {
+    if ("bundle" in params) {
+      const user = await this.completeEmailAuth(params);
+      return {
+        status: AlchemySignerStatus.CONNECTED,
+        user,
+      };
+    }
+
+    if (!("email" in params)) {
```
dphilipson, line 878: This check isn't strictly necessary because the type system should supposedly enforce that `email` is present in this branch, but doesn't hurt I suppose.

dphilipson (commented):
This looks great! Left some nitpicky comments.

account-kit/signer/src/base.ts
blakecduncan: oh yeah good call, will updated 👍

account-kit/signer/src/base.ts
blakecduncan: Oh yeah good catch

account-kit/signer/src/base.ts
```diff
@@ -312,6 +316,42 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
     }
   );
 
+  authenticateStep: (params: AuthParams) => Promise<AuthStepResult> =
```
rthomare, line 319: nit because this also starts and progresses the flow should the name be `handleAuth`, `authenticateOrAdvance`, or `progressAuth`? 

Should we break up starting and progressing authentication into two calls? Such as `login` and `verify`?

blakecduncan:
Closing as I plan to further investigate the different approaches here

PR #1542, opened 2025-04-16
feat: add mfa methods to rn-signer

blakecduncan:
Tested these changes on a separate branch in the example expo app.  See the PR [here](https://github.com/alchemyplatform/aa-sdk/pull/1544)

# Pull Request Checklist

---

## PR-Codex overview
This PR primarily focuses on refactoring the multi-factor authentication (MFA) functionality within the codebase, including renaming parameters and types, removing unused methods, and enhancing error handling.

### Detailed summary
- Deleted several unused `.mdx` files related to MFA methods.
- Renamed `EnableMfaParams` to `AddMfaParams` and `EnableMfaResult` to `AddMfaResult`.
- Updated return types for methods to reflect the new naming.
- Added a new `parseMfaError` function for better error handling.
- Implemented MFA methods in `BaseSignerClient` for listing, adding, verifying, and removing MFA factors.
- Removed deprecated methods from `RNSignerClient` and `AlchemySignerWebClient`.


account-kit/rn-signer/src/client.ts
```diff
@@ -28,6 +30,15 @@ import { z } from "zod";
...
 import { parseSearchParams } from "./utils/parseUrlParams";
+import { parseMfaError } from "./utils/parseMfaError";
+
+const MFA_PAYLOAD = {
+  GET: "get_mfa",
+  ADD: "add_mfa",
+  DELETE: "delete_mfas",
+  VERIFY: "verify_mfa",
+  LIST: "list_mfas",
+};
```
Blu-J, line 41:
```suggestion
} as const;
```

account-kit/rn-signer/src/utils/parseMfaError.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function parseMfaError(error: unknown): MfaFactor[] | null {
+  if (error instanceof Error) {
+    try {
+      const parsed = JSON.parse(error.message);
+      if (parsed?.data?.multiFactors) {
+        return parsed.data.multiFactors;
+      }
+    } catch {
+      // ignore JSON parse failures
```
Blu-J, line 11: ❓ Why are we ignoring it, I guess if we want to debug this will be findable in the network right?

blakecduncan: Yeah good question, json.parse throws an error if you pass it invalid json data.  I am assuming here if we are passing invalid json, it's not an mfa error.  In that case, we should return null and let the calling function handle the error

account-kit/rn-signer/src/client.ts
```diff
@@ -244,52 +284,162 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
...
    * Retrieves the list of MFA factors configured for the current user.
    *
-   * @throws {Error} This method is not implemented in RNSignerClient
+   * @returns {Promise<{ multiFactors: MfaFactor[] }>} A promise that resolves to an array of configured MFA factors
+   * @throws {Error} If no user is authenticated
    */
-  public override getMfaFactors(): Promise<{ multiFactors: MfaFactor[] }> {
-    throw new Error("getMfaFactors is not implemented in RNSignerClient");
-  }
+  public override getMfaFactors = async (): Promise<{
```
moldy530, line 290: shouldn't the base implementation just work here?

account-kit/rn-signer/src/client.ts
```diff
@@ -244,52 +284,162 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
...
-   * @param {EnableMfaParams} _params The parameters required to enable a new MFA factor
-   * @throws {Error} This method is not implemented in RNSignerClient
+   * @param {EnableMfaParams} params The parameters required to enable a new MFA factor
+   * @returns {Promise<EnableMfaResult>} A promise that resolves to the factor setup information
+   * @throws {Error} If no user is authenticated or if an unsupported factor type is provided
    */
-  public override addMfa(_params: EnableMfaParams): Promise<EnableMfaResult> {
-    throw new Error("enableMfa is not implemented in RNSignerClient");
-  }
+  public override addMfa = async (
```
moldy530, line 321: same here, shouldn't this just be handled by the base implementation?

account-kit/rn-signer/src/client.ts
```diff
@@ -244,52 +284,162 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
...
+   * @param {VerifyMfaParams} params The parameters required to verify the MFA factor
+   * @returns {Promise<{ multiFactors: MfaFactor[] }>} A promise that resolves to the updated list of MFA factors
+   * @throws {Error} If no user is authenticated
    */
-  public override verifyMfa(_params: VerifyMfaParams): Promise<{
-    multiFactors: MfaFactor[];
-  }> {
-    throw new Error("verifyMfa is not implemented in RNSignerClient");
-  }
+  public override verifyMfa = async (
```
moldy530, line 360: here as well, I can't recall but `this.turnkeyClient` should be available from the base impl which means that all of the code in here isn't specific to the environment that it's running in

moldy530 (commented):
general comment is that it feels like most of the MFA methods don't depend on environment specific things, so they should be defined in the base class rather than per client definition

account-kit/rn-signer/src/client.ts
jakehobbs: i had this question while working on my "api key signer" client impl as well... the MFA methods are all abstract on the `BaseSignerClient`, unlike many of the other methods (whoami, signRawMessage, etc.)

moldy530: oh yea I just went to check that too. that seems unexpected. there's nothing in these definitions that's environment specific from what I can see

blakecduncan: Yeah that is a good point, I didn't think about that.  When we first got started with MFA I was only implementing it for the web.  I don't think it ever clicked for me that all these methods can just live in the base client.

I can update that as part of this PR

jakehobbs: sounds good, thanks!

moldy530 (approved):

moldy530 (approved):

PR #1549, opened 2025-04-17
feat: add server-side api key signer client

jakehobbs:
# Pull Request Checklist

---

## PR-Codex overview
This PR introduces significant enhancements related to API key management in the `account-kit/signer` package. It adds functionality for creating and handling API keys, including new types, methods, and documentation updates.

### Detailed summary
- Deleted `experimental_createApiKey` method and replaced it with `addApiKey`.
- Introduced `CreateApiKeySigner` and `AlchemyApiKeySignerClient`.
- Added `unpackSignRawMessageBytes` utility function.
- Updated `MockAlchemySignerWebClient` to mock API key behavior.
- Enhanced documentation for new API key-related methods.
- Modified types to include API key parameters and responses.


account-kit/signer/src/client/alchemy.ts
```diff
@@ -0,0 +1,91 @@
+import type { SignerBody, SignerResponse, SignerRoutes } from "./types.js";
+import { VERSION } from "../version.js";
+import type { ConnectionConfig } from "@aa-sdk/core";
+import type { TSignedRequest } from "@turnkey/http";
+
+export class AlchemySignerClient {
```
adamegyed, line 6: Clean organization!

account-kit/signer/package.json
```diff
@@ -53,6 +53,7 @@
     "@aa-sdk/core": "^4.26.0",
     "@account-kit/logging": "^4.26.0",
     "@solana/web3.js": "^1.98.0",
+    "@turnkey/api-key-stamper": "^0.4.4",
```
florrdv, line 56: :question: wondering if we want to pull this out to a separate package? feels a little counterintuitive to have this all in one but it also makes the lift a lot bigger

account-kit/signer/src/apiKeySigner.ts
```diff
@@ -0,0 +1,176 @@
...
+   */
+  signTypedData = async <
+    const TTypedData extends TypedData | Record<string, unknown>,
+    TPrimaryType extends keyof TTypedData | "EIP712Domain" = keyof TTypedData
+  >(
+    params: TypedDataDefinition<TTypedData, TPrimaryType>
+  ): Promise<Hex> => {
+    const messageHash = hashTypedData(params);
+    return this.inner.signRawMessage(messageHash);
+  };
```
adamegyed, line 175: Could you also implement `signAuthorization` here? It's optional in the interface, and would allow us to use this key w/ 7702.

Side note: it would be nice to create a utility that goes from a "raw signer" to a `SmartAccountSigner`, since this is logically just a copy of `AlchemyWebSigner` and much of `LocalAccountSigner`.

account-kit/signer/src/apiKeySigner.ts
```diff
@@ -0,0 +1,176 @@
...
+  connection: ConnectionConfig;
+}
+
+/**
+ * ApiKeySignerClient is a client that uses an API key to sign messages.
+ */
+class ApiKeySignerClient {
+  protected turnkeyClient: TurnkeyClient;
+  protected alchemyClient: AlchemySignerClient;
```
florrdv, line 31: I would probably prefer not to call this alchemyClient cause I would expect that to be an sdk instance

adamegyed (approved):
Looks good, 1 small request

account-kit/signer/src/client/base.ts
```diff
@@ -397,20 +401,28 @@ export abstract class BaseSignerClient<TExportWalletParams = unknown> {
...
    * @param {string} params.name Name of the API key.
-   * @param {string} params.publicKey Public key to be used for the API key.
-   * @param {number} params.expirationSec Number of seconds until the API key expires.
-   * @throws {Error} If there is no authenticated user or the API key creation fails.
+   * @param {string} params.publicKey Public key to be used.
+   * @param {number} params.expirationSec Number of seconds until the key expires.
+   * @throws {Error} If there is no authenticated user or the request fails.
    */
-  public experimental_createApiKey = async (
-    params: experimental_CreateApiKeyParams
```
florrdv, line 410: :bangbang: we should deprecate this instead because Gensyn is actively using this in prod

account-kit/signer/src/client/types.ts
```diff
@@ -327,8 +327,10 @@ export type SubmitOtpCodeResponse =
   | { bundle: string; mfaRequired: false }
   | { mfaRequired: true; encryptedPayload: string; multiFactors: MfaFactor[] };
 
-export type experimental_CreateApiKeyParams = {
+export type AddApiKeyParams = {
```
florrdv, line 330: :bangbang: same here prolly, let's deprecate just in case

examples/ui-demo/package.json
```diff
@@ -16,6 +16,7 @@
     "@account-kit/infra": "^4.2.0",
     "@account-kit/logging": "^4.2.0",
     "@account-kit/react": "^4.2.0",
+    "@account-kit/smart-contracts": "^4.2.0",
```
florrdv, line 19: :bangbang: remove this from testing

site/pages/reference/account-kit/signer/classes/BaseSignerClient/addApiKey.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+```ts
+import { BaseSignerClient } from "@account-kit/signer";
+```
+
+## Usage
+
+```ts
+// To generate a new keypair:
+const ecdh = createECDH("prime256v1");
```
florrdv, line 23: :sparkles: thanks for including this!

site/pages/reference/account-kit/signer/classes/BaseSignerClient/addApiKey.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+### params.publicKey
+
+`string`
+Public key to be used.
+
+### params.expirationSec
+
+`number`
+Number of seconds until the key expires.
```
florrdv, line 49: :question: does this show that this is optional?

account-kit/signer/src/apiKeySigner.ts
jakehobbs: will implement that

let me dig into that a bit more

account-kit/signer/package.json
jakehobbs: we have a ton of other turnkey pkgs in here already (including every other stamper pkg that they offer):
```
    "@turnkey/http": "^2.6.2",
    "@turnkey/iframe-stamper": "^1.0.0",
    "@turnkey/viem": "^0.4.8",
    "@turnkey/webauthn-stamper": "^0.4.3",
```

jakehobbs: i'm hesitant to create another pkg just for this, but curious what others think

account-kit/signer/src/apiKeySigner.ts
jakehobbs: hmm what would you call it, just smth like alchemySignerClient then? (i think the sdk clients are always referred to as "smart account clients" fwiw)

site/pages/reference/account-kit/signer/classes/BaseSignerClient/addApiKey.mdx
jakehobbs: it's not optional

account-kit/signer/src/client/base.ts
jakehobbs: don't feel strongly since it's experimental & they are pinned to a specific version, but will do

account-kit/signer/src/client/base.ts
```diff
@@ -69,7 +74,8 @@ export abstract class BaseSignerClient<TExportWalletParams = unknown> {
     const { stamper, connection, rootOrgId } = params;
     this.rootOrg = rootOrgId ?? "24c1acf5-810f-41e0-a503-d5d13fa8e830";
     this.eventEmitter = new EventEmitter<AlchemySignerClientEvents>();
-    this.connectionConfig = ConnectionConfigSchema.parse(connection);
+    const connectionConfig = ConnectionConfigSchema.parse(connection);
+    this.alchemyClient = new AlchemySignerClient(connectionConfig);
```
moldy530, line 78: Wait so now we have two different signer clients? One base one that platforms implement, but it's functionality lives in a different client?

Seems a bit odd to have the abstract class called client not actually implement the client logic

account-kit/signer/src/client/alchemy.ts
moldy530, line 6: This should implement all the other routes as well if it's going to exist (signer config, prepare oauth, etc) though I'm not super clear on why the base signer client no longer impls this stuff since that's the base functionality imo (I'm probably missing some context)

account-kit/signer/src/client/base.ts
moldy530: Experimental is probs fine to kill, it's why we label them that way so we can make breaking changes. Let's just give gensyn a heads up

account-kit/signer/src/apiKeySigner.ts
```diff
@@ -0,0 +1,192 @@
...
+    publicKey: string;
+    privateKey: string;
+  };
+  connection: ConnectionConfig;
+}
+
+/**
+ * ApiKeySignerClient is a client that uses an API key to sign messages.
+ */
+class ApiKeySignerClient {
```
moldy530, line 29: This seems like a bunch of duplication of the logic that exists in the base signer client already? Does it not work for this to just extend that?

account-kit/signer/src/apiKeySigner.ts
```diff
@@ -0,0 +1,192 @@
...
+ * });
+ *
+ * const client = createAlchemySmartAccountClient({
+ *   account,
+ *   transport,
+ *   chain,
+ * });
+ * ```
+ */
+export class AlchemyApiKeySigner implements SmartAccountSigner {
```
moldy530, line 143: Any reason for this to not just extend base alchemy signer? Again seems like quite a bit of duplication of code

account-kit/signer/src/apiKeySigner.ts
jakehobbs: there are a lot of abstract methods on `BaseSignerClient` that i don't need... should we just throw on some of them like we do in the react native client?

it almost feels like i need something i step lower than the `BaseSignerClient` b/c i don't care about a lot of methods it has involving auth, etc.

i'm not totally following why some of these methods are `BaseSignerClient` are even abstract to begin with, esp. the ones involving MFA.

jakehobbs: working on this on another branch for to compare

account-kit/signer/src/apiKeySigner.ts
```diff
@@ -0,0 +1,192 @@
...
+    privateKey: string;
+  };
+  connection: ConnectionConfig;
+}
+
+/**
+ * ApiKeySignerClient is a client that uses an API key to sign messages.
+ */
+class ApiKeySignerClient {
+  protected turnkeyClient: TurnkeyClient;
```
dphilipson, line 30: I'm open to other opinions, but marking these `protected` implies that this class is intended to be subclassed, and I have a strong aversion to subclassing classes which are not abstract. I would recommend (almost) never subclassing in general, but in particular I think subclassing non-abstract classes is a giant pile of worms. Prefer marking this as `private` or making the intent of how this class should be subclassed clear.

Happy to go into more detail about this: subclassing is one of my biggest pet peeves.

account-kit/signer/src/apiKeySigner.ts
```diff
@@ -0,0 +1,192 @@
...
+  constructor(params: ApiKeySignerClientParams) {
+    this.inner = new ApiKeySignerClient(params);
+  }
+
+  /**
+   * Retrieves the current user's address.
+   *
+   * @returns {Promise<Address>} A promise that resolves to the user's address
+   */
+  getAddress = async (): Promise<Address> => {
```
dphilipson, line 161: I'd suggest marking all these methods from the interface as `override`, to make it more obvious to the reader and also produce more helpful errors if the interface isn't correctly implemented.

account-kit/signer/src/apiKeySigner.ts
jakehobbs: i def think that makes sense. i tried doing that initially but i believe i ran into some issues where we were trying to access local storage, which isn't available on the server. double checking though.

account-kit/signer/src/apiKeySigner.ts
jakehobbs: thanks, will keep that in mind & change it if we end up keeping this. i'm pretty new to abstract classes tbh. (most of the typescript i've written has followed more functional patterns rather than OOP.)

i'm about to put up another pr that handles things a bit differently though.

account-kit/signer/src/apiKeySigner.ts
jakehobbs: actually `BaseAlchemySigner` is what requires localstorage, due to it creating a new `SessionManage` in the constructor, which fails in a non-browser environment.

account-kit/signer/src/apiKeySigner.ts
jakehobbs: i see what you mean now, i am adding that

account-kit/signer/src/client/base.ts
jakehobbs: this will be removed in upcoming commit

account-kit/signer/src/client/alchemy.ts
jakehobbs: this will be removed in upcoming commit

account-kit/signer/src/apiKeySigner.ts
jakehobbs: will be extending it in an upcoming commit

account-kit/signer/src/apiKeySigner.ts
jakehobbs: hmmm the `LocalAccountSigner` doesn't actually use a raw signer though, as the local account actually implements a lot.

account-kit/signer/src/apiKeySigner.ts
```diff
@@ -0,0 +1,246 @@
...
+    );
+  }
+
+  override getOauthConfig = async (): Promise<OauthConfig> => {
+    throw new Error(
+      "OAuth methods are not supported by AlchemyApiKeySignerClient"
+    );
+  };
+
+  // TODO(jh): this will be moved to the base client, see https://github.com/alchemyplatform/aa-sdk/pull/1542#discussion_r2052673181
```
jakehobbs, line 172: will remove the MFA methods once the other pr lands

account-kit/signer/src/smartAccountSigner.ts
```diff
@@ -0,0 +1,91 @@
...
+  unpackSignRawMessageBytes,
+  type SmartAccountSigner,
+} from "@aa-sdk/core";
+import { hashAuthorization, type Authorization } from "viem/experimental";
+
+/**
+ * SmartAccountSignerFromClient is a class that implements the SmartAccountSigner interface.
+ * It wraps a BaseSignerClient and provides methods for signing messages, typed data, and authorizations.
+ */
+export class SmartAccountSignerFromClient implements SmartAccountSigner {
```
jakehobbs, line 21: naming is hard

account-kit/signer/src/apiKeySigner.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  /**
+   * Sets the user for the client
+   *
+   * @param {User} user The user object to set
+   */
+  public setUser = (user: User) => {
+    this.user = user;
+  };
```
moldy530, line 61: is this required? base has a `user` with a setter available already

account-kit/signer/src/apiKeySigner.ts
```diff
@@ -0,0 +1,246 @@
...
+ * @returns {Promise<SmartAccountSigner>} A promise that resolves to a SmartAccountSigner
+ * @throws {Error} If the API key is invalid for the given orgId
+ */
+export const createApiKeySigner = async (
+  clientParams: AlchemyApiKeySignerClientParams,
+  userOrgId: string
+): Promise<SmartAccountSigner> => {
+  const client = new AlchemyApiKeySignerClient(clientParams);
+  const user = await client.whoami(userOrgId);
+  client.setUser(user);
```
moldy530, line 244: client.whoami already sets `this.user`

account-kit/signer/src/smartAccountSigner.ts
moldy530: `BaseAlchemySigner` already implements `SmartAccountAuthenticator` which extends `SmartAccountSigner`

i need to see how this is used (still reviewing), but just a heads up in case that means anything here

account-kit/signer/src/base.ts
```diff
@@ -93,10 +87,12 @@ export type SignerConfig = {
  * Implements the `SmartAccountAuthenticator` interface and handles various signer events.
  */
 export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
+  extends SmartAccountSignerFromClient
```
moldy530, line 90: ah I see now

account-kit/signer/src/apiKeySigner.ts
```diff
@@ -0,0 +1,246 @@
...
+    privateKey: string;
+  };
+}
+
+/**
+ * AlchemyApiKeySignerClient is a client for signing messages using an API key.
+ * It extends the BaseSignerClient and uses the ApiKeyStamper for signing.
+ * Primarily intended to be used server-side.
+ */
+export class AlchemyApiKeySignerClient extends BaseSignerClient<undefined> {
```
moldy530, line 37: nit: can we put this in `src/client/apiKey` or something and then below have `createApiKeySigner` be the only thing exported from src/apiKeySigner?

account-kit/signer/src/apiKeySigner.ts
jakehobbs: i need a way to bubble it up from the `AlchemyApiKeySignerClient` to `createApiKeySigner`, and i'm avoiding doing it within the constructor since it depends on an async call.

jakehobbs: oh wait, nvm i forgot that `whoami` sets it already.

account-kit/signer/src/apiKeySigner.ts
jakehobbs: oh sick thx, forgot about that

account-kit/signer/src/apiKeySigner.ts
jakehobbs: sure thing

account-kit/signer/src/client/apiKey.ts
```diff
@@ -0,0 +1,193 @@
...
+  OauthConfig,
+  OauthParams,
+  OtpParams,
+  RemoveMfaParams,
+  SignupResponse,
+  SubmitOtpCodeResponse,
+  User,
+  VerifyMfaParams,
+} from "./types";
+import { BaseSignerClient } from "./base";
```
gpuliyar, line 21: 🚫 **[eslint]** <[import/extensions](https://github.com/import-js/eslint-plugin-import/blob/v2.29.0/docs/rules/extensions.md)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Missing file extension for "./base"

account-kit/signer/src/client/apiKey.ts
```diff
@@ -0,0 +1,165 @@
...
+  CreateAccountParams,
+  EnableMfaResult,
+  GetWebAuthnAttestationResult,
+  MfaFactor,
+  OauthConfig,
+  SignupResponse,
+  SubmitOtpCodeResponse,
+  User,
+} from "./types";
+import { BaseSignerClient } from "./base";
```
gpuliyar: 🚫 **[eslint]** <[import/extensions](https://github.com/import-js/eslint-plugin-import/blob/v2.29.0/docs/rules/extensions.md)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Missing file extension for "./base"

jakehobbs:
## change to signer service to accept api key must be released first

account-kit/signer/src/base.ts
dphilipson, line 90: Wait, this is an abstract class extending a concrete class? I think this is really asking for trouble. I'm going to go into my screed against subclassing, just a moment.

jakehobbs: tbh feeling like adding this `SmartAccountSignerFromClient` might just be an over-abstraction.

i was trying to avoid some code duplication for handling the signMessage/signTypedData/signAuthorization, but doing all of this honestly results in more code, b/c those methods are only a couple of lines anyways.

pushing some changes to simplify.

jakehobbs:
Keeping this `experimental` for now, w/ the possibility of exploring further later.

PR #1555, opened 2025-04-21
feat(wip): implement viem compatible SCA impls for LA

moldy530:


account-kit/infra/src/client/viem.ts
```diff
@@ -0,0 +1,33 @@
+import { createBundlerClient } from "viem/account-abstraction";
+import type { AlchemySmartAccountClientConfig } from "./smartAccountClient";
+
+// Yea this isn't great or going to work well, we should just keep using our smart account client, but change the account type to match the viem one
```
Blu-J, line 4: 💬 What about the idea of a transform then from the current client into this new client. Like rust's `Into` traits
Or, is it more worth it for the changing of the account type?

account-kit/infra/src/client/viem.ts
```diff
@@ -0,0 +1,33 @@
...
+        throw new Error(
+          "this is the biggest issue with viem's approach. we can't skip gas estimation during paymaster flows..."
+        );
+      },
+      async getPaymasterData(parameters) {
+        throw new Error("port over the paymaster data middleware");
+      },
+    },
+    userOperation: {
+      // This doesn't have the ability to skip gas estimation AFAICT
```
Blu-J, line 25: 🙀

account-kit/infra/src/client/viem.ts
```diff
@@ -0,0 +1,33 @@
...
+    chain: params.chain,
+    paymasterContext: { policyId: params.policyId },
+    paymaster: {
+      async getPaymasterStubData(parameters) {
+        if (!("policyId" in parameters.context)) {
+          throw new Error("policyId is required");
+        }
+
+        throw new Error(
+          "this is the biggest issue with viem's approach. we can't skip gas estimation during paymaster flows..."
```
Blu-J, line 17: 💬 If we do some hacky stuff, like stub it out, it might be possible. But would is it worth the cost to our souls.
Is there a line of communication to the viem team on this at least?

PR #1565, opened 2025-04-25
feat: viem wallet client conversion

florrdv:
# Pull Request Checklist

---

## PR-Codex overview
This PR introduces a new method, `toViemWalletClient`, to the `BaseAlchemySigner` class, allowing for the adaptation of the signer to a `viem` `WalletClient`. It enhances interoperability with `viem` by providing options for transport and configuration.

### Detailed summary
- Added a new method `toViemWalletClient` to `BaseAlchemySigner`.
- Method accepts options for transport and returns a `WalletClient`.
- Implements a fallback transport for handling requests.
- Supports signing messages and typed data through the `viemAccount`.
- Provides example usage in the documentation comments.


account-kit/signer/src/base.ts
```diff
@@ -809,6 +817,74 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
...
+      ...opts,
+      account: viemAccount,
+      transport: custom({
+        async request({ method, params }) {
+          switch (method) {
+            case "personal_sign":
+              return viemAccount.signMessage(
+                isHex(params[0])
+                  ? { message: { raw: params[0] } }
+                  : { message: params[0] }
```
adamegyed, line 874: It should be technically possible to sign ascii/utf characters that represent hex characters, and this conversion prevents that by re-interpreting the characters to their raw value if they happen to be hex characters.

I think viem defaults to having the parameters to `personal_sign` be either a string or an object containing a field called `raw` - could we continue to use that type here?

account-kit/signer/src/base.ts
```diff
@@ -809,6 +817,74 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
...
+              );
+            case "eth_signTypedData_v4":
+              return viemAccount.signTypedData(params[1]);
```
adamegyed, line 877: Should we assert `params[0]` to be the account address?

account-kit/signer/src/base.ts
moldy530: That's what the check here is already doing no? Passing raw if hex so that it doesn't get treated as string

account-kit/signer/src/base.ts
moldy530: Yea +1

moldy530: And probs same above checking that params[1] is the signer too

account-kit/signer/src/base.ts
```diff
@@ -809,6 +817,74 @@ export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
...
+      chain: opts.chain,
+      pollingInterval: opts.pollingInterval ?? 4_000,
+    });
+
+    return createWalletClient({
+      ...opts,
+      account: viemAccount,
+      transport: custom({
+        async request({ method, params }) {
+          switch (method) {
```
moldy530, line 869: Also eth_sendTx should be handled here as well which should call sign tx then submit it

account-kit/signer/src/base.ts
dphilipson: No, Adam's saying that this makes it impossible to sign the string "0x1234" because it will get interpreted as bytes `0x12`, `0x34`. (Or not _impossible_, but someone would need to pass the hex encoding of the above string, i.e. "0x307831323334"). Adam's saying that the caller should be responsible for specifying whether their string should be interpreted as hex rather than us trying to infer it.

But actually I think it's wrong to accept the type with `raw` here too. This method is trying to emulate the `personal_sign` method of the eth-jsonrpc API, and therefore should _always_ interpret the input has hex ([Metamask's docs](https://docs.metamask.io/wallet/reference/json-rpc-methods/personal_sign/)). Getting non-hex input here should be an error.

account-kit/signer/src/base.ts
dphilipson, line 869: Question: do we also want to implement `eth_requestAccounts` ([Metamask's docs](https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_requestaccounts/))?

account-kit/signer/src/base.ts
adamegyed: `params[1]` is the typed data object, so I don't think there's anything we can assert over that.

account-kit/signer/src/base.ts
adamegyed: Ah, good catch. So the real fix here is to just throw an error instead of handling the case of `{ message: params[0] }`, right?

dphilipson: Yeah, I think so.

florrdv:
Closing this for now because there's no immediate need to create EIP1193 viem wallet clients.

PR #1566, opened 2025-04-25
feat: add erc20 to 7677 middleware

andysim3d:
# Pull Request Checklist

---

## PR-Codex overview
This PR introduces a new `PolicyToken` type and integrates it into the `erc7677Middleware`. It enhances the middleware to support ERC-20 paymaster interactions, allowing for token-based fee payments in user operations.

### Detailed summary
- Added `PolicyToken` type with properties: `tokenAddress`, `approvalMode`, `maxTokenAmount`, `erc20Name`, and `version`.
- Updated `ClientMiddlewareArgs` to include `policyToken`.
- Enhanced `erc7677Middleware` to calculate token amounts based on gas fees and `policyToken`.
- Implemented logic for requesting token conversion rates and creating permit data for ERC-20 tokens.

### DIFF
--- a/aa-sdk/core/src/middleware/erc7677middleware.ts
+++ b/aa-sdk/core/src/middleware/erc7677middleware.ts
@@ -54,6 +54,30 @@ export type Erc7677Client<
   TContext extends Record<string, any> = Record<string, any>
 > = Client<T, Chain, undefined, Erc7677RpcSchema<TContext>>;
 
+export type Erc20PaymasterClientRpcSchema = [
+  {
+    Method: "alchemy_requestPaymasterTokenQuote";
+    Parameters: [
+      {
+        request: {
+          tokenAddress: Hex;
+          policyId: string;
+        };
+      }
+    ];
+    ReturnType: {
+      ethToTokenConversionRate?: Address; // conversion rate of Eth to Erc20Token
+    };
+  }
+];
+
+export type Erc20PaymasterClient<T extends Transport = Transport> = Client<
+  T,
+  Chain,
+  undefined,
+  Erc20PaymasterClientRpcSchema
+>;
+
 export type Erc7677MiddlewareParams<
   TContext extends Record<string, any> | undefined =
     | Record<string, any>
@@ -70,6 +94,8 @@ export type Erc7677MiddlewareParams<
       ) => Promise<TContext>)
     | TContext;
 };
+
+const Eth2Wei: bigint = 1_000_000_000_000_000_000n;
 
 /**
  * Middleware function for interacting with ERC-7677 enabled clients. It supports resolving paymaster and data fields for user operations.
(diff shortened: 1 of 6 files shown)


aa-sdk/core/src/middleware/erc7677middleware.ts
```diff
@@ -54,6 +54,30 @@ export type Erc7677Client<
...
+    Parameters: [
+      {
+        request: {
+          tokenAddress: Hex;
+          policyId: string;
+        };
+      }
+    ];
+    ReturnType: {
+      ethToTokenConversionRate?: Address; // conversion rate of Eth to Erc20Token
```
jakehobbs, line 69: is this backwards?

```suggestion
        request: {
          tokenAddress: Address;
          policyId: string;
        };
      }
    ];
    ReturnType: {
      ethToTokenConversionRate?: Hex; // conversion rate of Eth to Erc20Token
```

aa-sdk/core/src/middleware/erc7677middleware.ts
```diff
@@ -71,6 +95,8 @@ export type Erc7677MiddlewareParams<
     | TContext;
 };
 
+const Eth2Wei: bigint = 1_000_000_000_000_000_000n;
```
jakehobbs, line 98: this may be of use to you: https://viem.sh/docs/utilities/parseEther.html

aa-sdk/core/src/middleware/erc7677middleware.ts
```diff
@@ -159,21 +185,143 @@ export function erc7677Middleware<
...
         : params?.context) ?? {};
+    const entrypoint = account.getEntryPoint();
 
     if (!client.chain) {
       throw new ChainNotFoundError();
     }
 
+    let max_token: BigInt | undefined = undefined;
+    if (policyToken !== undefined) {
+      let total_gas = BigInt(0);
```
jakehobbs, line 204: nit (style): we normally camel case variable names in ts

aa-sdk/core/src/middleware/erc7677middleware.ts
```diff
@@ -159,21 +185,143 @@ export function erc7677Middleware<
...
+      total_gas += uo.callGasLimit
+        ? BigInt(uo.callGasLimit.toLocaleString())
+        : BigInt(0);
```
dphilipson, line 207: This is fine, but fyi you have the option of writing `BigInt(0)` as `0n`.

aa-sdk/core/src/middleware/erc7677middleware.ts
```diff
@@ -159,21 +185,143 @@ export function erc7677Middleware<
...
+
+      total_gas += uo.preVerificationGas
+        ? BigInt(uo.preVerificationGas.toLocaleString())
+        : BigInt(0);
+
+      total_gas += uo.verificationGasLimit
+        ? BigInt(uo.verificationGasLimit.toLocaleString())
+        : BigInt(0);
+
+      let paymasterAddress = "0x";
```
dphilipson, line 217: (Nit) If this variable is always going to be assigned to later, then don't initialize it with an assignment. Instead do
```ts
let paymasterAddress: Hex;
```
This has two advantages:
* It gives it the more narrow type `Hex` instead of `string`, meaning it would be a type error if we accidentally assign a non-hex value to it later.
* It makes it a type error if we forget to assign to it in one of the branches below.

aa-sdk/core/src/middleware/erc7677middleware.ts
```diff
@@ -159,21 +185,143 @@ export function erc7677Middleware<
...
 
     if (!client.chain) {
       throw new ChainNotFoundError();
     }
 
+    let max_token: BigInt | undefined = undefined;
+    if (policyToken !== undefined) {
+      let total_gas = BigInt(0);
+      total_gas += uo.callGasLimit
+        ? BigInt(uo.callGasLimit.toLocaleString())
```
dphilipson, line 206: This doesn't work. `uo.callGasLimit` has type `PromiseOrValue<...>` meaning it is either an actual value or a promise that eventually resolves to a value. In the latter case, `.toLocaleString()` returns `"[object Promise]"` rather than anything representing the value.

You should read from `userOp` instead of `uo`, which represents the same operation with all fields awaited. Likewise throughout.

aa-sdk/core/src/middleware/erc7677middleware.ts
```diff
@@ -159,21 +185,143 @@ export function erc7677Middleware<
...
+        : BigInt(0);
+
+      total_gas += uo.verificationGasLimit
+        ? BigInt(uo.verificationGasLimit.toLocaleString())
+        : BigInt(0);
+
+      let paymasterAddress = "0x";
+
+      if (entrypoint.version === "0.7.0") {
+        const uo_07 = uo as UserOperationStruct<"0.7.0">;
```
dphilipson, line 220: Just to reiterate: `uo` is not a `UserOperationStruct<"0.7.0">`, it is a `PromiseOrValue<UserOperationStruct<"0.7.0">>`. You should read from `userOp` instead.

aa-sdk/core/src/middleware/erc7677middleware.ts
```diff
@@ -159,21 +185,143 @@ export function erc7677Middleware<
...
+          ? BigInt(uo_07.paymasterPostOpGasLimit.toLocaleString())
+          : BigInt(0);
+        total_gas += uo_07.paymasterVerificationGasLimit
+          ? BigInt(uo_07.paymasterVerificationGasLimit.toLocaleString())
+          : BigInt(0);
+      } else {
+        const uo_06 = uo as UserOperationStruct<"0.6.0">;
+        paymasterAddress = uo_06.paymasterAndData
+          ? uo_06.paymasterAndData.toString().slice(0, 42)
+          : "0x";
```
dphilipson, line 233: Do we not need to update `total_gas` in this branch?

aa-sdk/core/src/middleware/erc7677middleware.ts
dphilipson: Also if you do need a constant, TypeScript convention is upper snake-case, e.g. `ETH_2_WEI`, although I would probably name this something like `WEI_PER_ETH` which is a big more specific.

aa-sdk/core/src/middleware/erc7677middleware.ts
```diff
@@ -159,21 +185,143 @@ export function erc7677Middleware<
...
+        message: {
+          owner: account.address as Hex,
+          spender: paymasterAddress as Hex,
+          value: policyToken.maxTokenAmount,
+          nonce: (await account.getAccountNonce()) + BigInt(1),
+          deadline: BigInt(0xffffffffffffffffffffffffffffffff),
+        },
+      } as const;
+
+      context["erc20Context"] = {
```
dphilipson, line 317: (Nit) Can write this as `context.erc20Context`.

aa-sdk/core/src/middleware/erc7677middleware.ts
```diff
@@ -159,21 +185,143 @@ export function erc7677Middleware<
...
+        ],
+      });
+      max_token =
+        (max_gas_in_wei *
+          BigInt(
+            ratio.ethToTokenConversionRate ? ratio.ethToTokenConversionRate : 0
+          )) /
+        Eth2Wei;
+
+      const typed_permit_data = {
```
dphilipson, line 258: Looks unused. (Ignore if you just haven't written the part that uses it yet).

dphilipson (changes requested):

aa-sdk/core/src/middleware/erc7677middleware.ts
moldy530, line 1: does erc7677 define how erc20 should be handled or is the logic in this file specific to alchemy?

moldy530: if this isn't to a specific spec (it doesn't look like it since it looks similar to the logic for alchemy_requestGasAndPaymasterAndData), then we shouldn't it in here. Instead we should put it just in `gasManager.ts` in account-kit/infra

account-kit/infra/src/middleware/gasManager.ts
```diff
@@ -54,11 +54,20 @@ export function alchemyGasManagerMiddleware(
```
moldy530, line 54: regarding my comment about the logic in aa-sdk/core. I think that logic should go within the `alchemyGasManagerMiddleware`.

you can leverage the base middleware:
```
erc7677Middleware<{ policyId: string | string[] }>({
    context: { policyId: policyId },
  })
```
to make calls to the RPC and do all of the 7677 stuff, but since you also need to sign permit data based on mode you'll likely have to wrap the middleware here.

likely something like:
```ts
export function alchemyGasManagerMiddleware(
  policyId: string | string[]
): Pick<ClientMiddlewareConfig, "dummyPaymasterAndData" | "paymasterAndData"> {
  const baseMiddleware = erc7677Middleware<{ policyId: string | string[] }>({
    context: { policyId: policyId },
  });
  // don't actually do this with ...middlewareParams it's just cuz I'm being too lazy to scroll to what should actually be here
  return {
    dummyPaymasterAndData: (...args) => {
      // if you need custom logic, do it here and at some point use:
      baseMiddleware.dummyPaymasterAndData(...args);
    },
    paymasterAndData: (...args) => {
      // do your custom logic, call baseMiddleware.paymasterAndData(...args), do other custom logic, return result
    }
  }
}
```

moldy530 (commented):
How much of this logic is the same as what's in the `alchemyRequestGasAndPaymasterData` middleware? Since we have to move the code in `aa-sdk/core` into `account-kit/infra` anyways, then we should just share as much of the code as we can

moldy530:
Also, how are we testing this?

PR #1603, opened 2025-05-08
feat: auth linking when in "redirect" oauth mode

dphilipson:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on the removal of several MFA-related documentation files and updates to the `useSigner` hook, `LocalAccountSigner` methods, and the `AlchemyWebSigner` class, enhancing the handling of OAuth and email authentication linking prompts.

### Detailed summary
- Deleted MFA-related documentation files for `AlchemySignerWebClient`.
- Updated `useSigner` hook to accept an optional `override` parameter.
- Changed `LocalAccountSigner` methods to use `privateKeyToAccountSigner`.
- Enhanced `AlchemyWebSigner` to manage authentication linking prompts.
- Added logic for setting authentication linking state in `BaseAlchemySigner`.


account-kit/react/src/AlchemyAccountProvider.tsx
```diff
@@ -101,6 +103,19 @@ export const AlchemyAccountProvider = (
...
+      const cancel = signer.on("emailAuthLinkingRequired", (email) => {
+        setAuthStep({ type: "otp_verify", email });
+      });
+
+      return () => {
+        cancel();
+      };
+    }
+    return undefined;
+  }, [signer, setAuthStep]);
```
moldy530, line 117: we do something similar to this for new user signup. But I think the listener is actually added to the `Modal` itself https://github.com/alchemyplatform/aa-sdk/blob/main/account-kit/react/src/components/auth/modal.tsx#L30

Main reason was to avoid misfires or lack of fires. Also it's not enough to call `setAuthStep`, I think you also need to check if you need to open the modal.

this one might be a bit trickier... in some cases (the demo app), the dev embeds the auth card directly into their dom rather than using the modal. I imagine in that setup, we shouldn't be popping up the modal at all, but keeping the card based flow.

moldy530 (commented):
overall LGTM, the main issue I can see is that it doesn't work for folks using the modal unless they know to pop open the modal on this redirect

account-kit/react/src/AlchemyAccountProvider.tsx
moldy530: something we could do is add a param for `useModal` to `uiConfig`. Should default to true. 

if it's set to false, then `onEmailAuthLinkingRequired `, we can not pop the modal, otherwise we do. For add passkey on signup, we still always pop the modal.

moldy530 (approved):

PR #1683, opened 2025-06-13
feat: allow nonce key overrides by default for non-deployed accounts

Zer0dot:
Currently, we don't support overriding the nonce key for non-deployed accounts besides ma-v2 due to a snippet of code in the default `getNonce()` function. This PR removes that snippet.

---

## PR-Codex overview
This PR focuses on enhancing the handling of nonce keys in user operations within the `Light Account` tests, particularly by introducing logic to override nonce keys based on the version of the light account.

### Detailed summary
- Removed the check for deployed accounts in the `getNonce` function.
- Updated imports to include `fromHex` and `toHex` from `viem`.
- Changed `const` to `let` for `instance` to allow reassignment.
- Added a new test case to verify nonce key overriding based on version.


jakehobbs (approved):
makes sense to me

jakehobbs:
is the docs change on purpose?

PR #1755, opened 2025-07-09
feat: add react native hooks for solana

blakecduncan:
# Pull Request Checklist

---

## PR-Codex overview
This PR introduces new features for Solana transaction handling, including enhanced error handling and confirmation options, as well as the ability to use polling for transaction confirmations. It improves the `useSolanaTransaction` and `broadcast` functionalities.

### Detailed summary
- Added `useSolanaConnection`, `useSolanaSigner`, and `useSolanaTransaction` hooks.
- Implemented error handling in `useSolanaSigner` with a try-catch block.
- Introduced `confirmationOptions` in `SolanaTransactionHookParams`.
- Enhanced `broadcast` function to accept `ConfirmationOptions`.
- Implemented polling mechanism for transaction confirmation in `broadcastWithPolling`.

### DIFF
--- a/account-kit/react-native/src/index.ts
+++ b/account-kit/react-native/src/index.ts
@@ -19,6 +19,8 @@ export {
   useSignMessage,
   useSignTypedData,
   useSmartAccountClient,
+  useSolanaConnection,
+  useSolanaSigner,
   useUser,
   useWaitForUserOperationTransaction,
 } from "@account-kit/react/hooks";
--- a/account-kit/react-native/src/index.ts
+++ b/account-kit/react-native/src/index.ts
@@ -21,6 +21,7 @@ export {
   useSmartAccountClient,
   useSolanaConnection,
   useSolanaSigner,
+  useSolanaTransaction,
   useUser,
   useWaitForUserOperationTransaction,
 } from "@account-kit/react/hooks";
--- a/account-kit/react/src/hooks/useSolanaSigner.ts
+++ b/account-kit/react/src/hooks/useSolanaSigner.ts
@@ -41,7 +41,12 @@ export const useSolanaSigner = (
     const signerStatus = opts.signerSet?.signerStatus || fallbackSignerStatus;
     if (!signer) return null;
     if (!signerStatus.isConnected) return null;
-    return signer.toSolanaSigner();
+    try {
+      return signer.toSolanaSigner();
+    } catch (error) {
+      console.log(error);
+      return null;
+    }
   }, [
     opts.signerSet?.signer,
     opts.signerSet?.signerStatus,
--- a/account-kit/react/src/solanaNetwork.ts
+++ b/account-kit/react/src/solanaNetwork.ts
@@ -11,6 +11,19 @@ import bs58 from "bs58";
 
 export { Connection };
 
+export interface ConfirmationOptions {
+  /**
+   * Whether to use polling instead of WebSocket subscriptions for transaction confirmation.
+   * Useful for environments where WebSocket connections are unreliable (e.g., React Native).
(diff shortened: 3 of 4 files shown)


account-kit/react/src/hooks/useSolanaSigner.ts
```diff
@@ -41,7 +41,12 @@ export const useSolanaSigner = (
     const signerStatus = opts.signerSet?.signerStatus || fallbackSignerStatus;
     if (!signer) return null;
     if (!signerStatus.isConnected) return null;
-    return signer.toSolanaSigner();
+    try {
+      return signer.toSolanaSigner();
```
blakecduncan, line 45: For some reason in expo when the app loads we get to this line (so it passes the above to checks) but inside `toSolanaSigner` the user is not defined and the `NotAuthenticatedError` is thrown

```
if (!this.inner.getUser()) {
  throw new NotAuthenticatedError();
}
```

We don't see this issue in react as far as I know.  I suspect it has something to do with spinning up the app in dev mode in expo.  After editing code and the app does a reload everything works.

This is a quick fix but we should probably figure out if what the root cause of this is

account-kit/react-native/src/index.ts
```diff
@@ -19,6 +19,9 @@ export {
   useSignMessage,
   useSignTypedData,
   useSmartAccountClient,
+  useSolanaConnection,
+  useSolanaSigner,
+  useSolanaTransaction,
```
blakecduncan, line 24: Not sure if there is a reason we never exported these?  In my testing it seems to work

account-kit/react/src/hooks/useSolanaSigner.ts
florrdv: hmmm, signerStatus being 'connected' here should mean the user is present, right? I don't see how this error handling would solve much unless our state is totally screwed up

maybe a better AI would be to make sure we have an SSR/persistent storage method for react native?

florrdv (dismissed):

account-kit/react/src/solanaNetwork.ts
```diff
@@ -11,6 +11,19 @@ import bs58 from "bs58";
...
 export { Connection };
 
+export interface ConfirmationOptions {
+  /**
+   * Whether to use polling instead of WebSocket subscriptions for transaction confirmation.
+   * Useful for environments where WebSocket connections are unreliable (e.g., React Native).
+   *
+   * @default false
+   */
+  usePolling?: boolean;
```
jakehobbs: nit (not blocking, just a suggestion): i think it's best practice to avoid naming anything `use*` in React that is not a hook.

account-kit/react/src/solanaNetwork.ts
```diff
@@ -51,24 +64,101 @@ export async function dropTokens(
...
+): Promise<string> {
+  const txSignature =
+    "version" in signedTransaction
+      ? await connection.sendTransaction(signedTransaction)
+      : await connection.sendRawTransaction(signedTransaction.serialize());
+
+  let confirmed = false;
+  let attempts = 0;
+
+  while (!confirmed && attempts < maxPollingAttempts) {
```
florrdv: nit: I would prefer to do this in a for loop `attempt < maxPollingAttempts` and return out of it when completed / reject when reject when we complete it

florrdv (approved):

account-kit/react/src/solanaNetwork.ts
florrdv: enablePolling could be good

adamegyed (approved):

PR #1804, opened 2025-07-24
feat(signer): v5 interfaces

dphilipson:
* Creates new packages, `@alchemy/signer`, `@alchemy/signer-web` and `@alchemy/signer-rn`. Note that the web signer is now in a separate package rather than being part of `/signer`.
* Introduces a new set of interfaces: instead of multiple signer classes which represent both logged-in and logged-out signers, have separate classes `Signer` for an authenticated signer which is ready to start signing, and `AuthClient` which performs login tasks and returns a `Signer` once login is complete.
* Speaking of auth, there are separate methods for different login types, e.g. `sendAuthEmail`, `submitOtpCode`, rather than a single `authenticate` method that does it all.
* At the moment, nothing is done related to session storage. I'm not actually positive we want that to be automatic, as it adds unrelated responsibilities to the signer classes and isn't useful for some applications (e.g. scripts) `Signer` for an authenticated signer which is ready to start signing, and `AuthClient` which performs login tasks and returns a `Signer` when login is complete.
* At the moment, nothing is done related to session storage. I'm not actually sure we want that to be automatic or even part of the signer packages, as it adds unrelated responsibilities to the signer classes and isn't useful for some use cases (e.g. scripts).

---

## PR-Codex overview
This PR primarily introduces version management and configuration updates across multiple packages in the Alchemy SDK. It implements versioning through autogenerated files, updates TypeScript configurations, and sets up testing configurations for different platforms.

### Detailed summary
- Added autogenerated `VERSION` exports in `version.ts` files for multiple packages.
- Updated `tsconfig.json` and `tsconfig.build.json` for TypeScript settings in all signer packages.
- Implemented `inject-version.ts` to write the version from `package.json`.
- Created `vitest.config.ts` files for testing configurations in all signer packages.
- Modified `package.json` files for all signer packages to include versioning and scripts.
- Introduced new types in `types.ts` for user and authentication methods in `signer`.
- Enhanced `AuthClient` and `Signer` classes with new methods and properties for authentication handling.


packages/signer/src/authClient.ts
```diff
@@ -0,0 +1,88 @@
...
+    private readonly transport: AlchemyTransport,
+    private readonly createTekStamper: CreateTekStamperFn,
+    private readonly createWebAuthnStamper: CreateWebAuthnStamperFn
+  ) {}
+
+  private tekStamperPromise: Promise<TekStamperAndPublicKey> | null = null;
+
+  public async sendAuthEmail(email: string): Promise<void> {
+    const { targetPublicKey } = await this.getTekStamper();
+    const otpId = await notImplemented(email, targetPublicKey);
```
linnall, line 25: is the `otpId` going to be used to map inflight auth flows in the session manager?

linnall: nvrm you basically answer this question in the comment on the next line. I assume the answer is yes

PR #1821, opened 2025-07-29
feat: migrate nonstandard bundler methods

trevoraron:
Build a 'viem bundler' factory that returns a bundler that uses our nonstandard methods that works within viem

Asana Task: https://app.asana.com/1/1129441638109975/project/1210727695159716/task/1210801854306017?focus=true

# Pull Request Checklist

---

## PR-Codex overview
This PR introduces the `bigIntMultiply` function to the `@alchemy/common` package, modifies related documentation, and updates the bundler client to utilize this new function for fee estimation. It also revises the RPC schema to include new methods for Alchemy's bundler.

### Detailed summary
- Added `bigIntMultiply` function in `packages/common/src/utils/bigint.ts`.
- Updated documentation for `bigIntMultiply` in `docs/pages/reference/alchemy/common/functions/bigIntMultiply.mdx`.
- Modified `packages/common/src/index.ts` to export `bigIntMultiply`.
- Updated `createAlchemyBundlerClient` in `packages/smart-accounts/src/clients.ts` to use `bigIntMultiply`.
- Revised RPC schema in `packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts` to include `rundler_maxPriorityFeePerGas`.
- Removed old documentation references in `docs/docs.yml` and added new ones for `bigIntMultiply`.


packages/smart-accounts/src/viemBundlerFactory.ts
```diff
@@ -0,0 +1,77 @@
...
+    entryPointAddress,
+    estimateFeesPerGas: async ({ bundlerClient }: any) => {
+      try {
+        // Try to use rundler_maxPriorityFeePerGas if available
+        const [block, maxPriorityFeePerGasEstimate] = await Promise.all([
+          getBlock(bundlerClient, { blockTag: "latest" }),
+          bundlerClient.request({
+            method: "rundler_maxPriorityFeePerGas",
+            params: [],
+          } as any),
```
adamegyed: Instead of using `as any`, we should define this method in an RPC schema, similar to this: https://github.com/alchemyplatform/aa-sdk/blob/7364a5c98126eddf502ea023d93b124a10d195ff/account-kit/infra/src/client/types.ts#L23-L25

Let's start by defining a custom `AlchemyRPCSchema` that just contains this for now, and we can incrementally add the other methods as needed.

packages/smart-accounts/src/viemBundlerFactory.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @param {object} config - Configuration for the bundler client.
+ * @param {object} config.account - The account to use for the bundler client.
+ * @param {object} config.chain - The chain configuration for the bundler client.
+ * @param {object} config.transport - The transport to use for the bundler client.
+ * @param {string} [config.entryPointAddress] - Optional entry point address. Defaults to DEFAULT_ENTRY_POINT_ADDRESS.
+ * @param {object} [config.userOperation] - Optional userOperation config to override defaults.
+ * @returns {ReturnType<typeof viemCreateBundlerClient>} A bundler client with Alchemy/rundler fee estimation logic.
+ */
+export function AlchemyViemBundlerFactory(config: {
```
adamegyed: instead of defining a replacement for `createBundlerClient`, we should define this as a "custom function" for estimating fees that can be passed into the existing `createBundlerClient` function. Something like this: https://viem.sh/account-abstraction/clients/bundler#useroperationestimatefeespergas

Since the `rundler_maxPriorityFeePerGas` method only provides priority fees, we should keep the existing AA-SDK functionality for estimating base fees, including the options to pass multipliers (from `AlchemyFeeEstimator`).

packages/smart-accounts/src/viemBundlerFactory.ts
```diff
@@ -0,0 +1,77 @@
...
+  account: any;
+  chain: any;
+  transport: any;
+  entryPointAddress?: string;
+  userOperation?: any;
```
adamegyed: Before committing things, let's verify that everything is appropriately type (no `any`)

packages/smart-accounts/src/viemBundlerFactory.ts
```diff
@@ -0,0 +1,77 @@
+import {
+  createBundlerClient as viemCreateBundlerClient,
+  bundlerActions,
+} from "viem/account-abstraction";
+import { getBlock } from "viem/actions";
+import { fromHex, type Hex } from "viem";
+
+export const DEFAULT_ENTRY_POINT_ADDRESS =
+  "0x5ff137d4b0fdcd49dca30c7cf57e578a026d2789";
```
adamegyed: We'll need to support both EP v0.6 and v0.7 (also eventually v0.8). I believe viem's bundler client should already have structure for getting EP versions and default addresses, so we don't need to redefine them.

packages/smart-accounts/src/viemBundlerFactory.ts
```diff
@@ -0,0 +1,77 @@
...
+          maxPriorityFeePerGas: fromHex(
+            maxPriorityFeePerGasEstimate as Hex,
+            "bigint",
+          ),
+          maxFeePerGas:
+            (baseFeePerGas * 150n) / 100n +
+            fromHex(maxPriorityFeePerGasEstimate as Hex, "bigint"),
+        };
+      } catch (error) {
+        // Fallback to standard viem fee estimation if rundler method is not available
```
adamegyed: I don't think we want this behavior as a fallback - it will just put mispriced UOs in the mempool. Let's just allow the error to bubble up.

trevoraron:
@adamegyed removed all the `any`s!

interestingly i had linter issues without the entry point address in this method so i brought it back, but tests passed without it 👀

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (baseFeePerGas == null) throw new Error("baseFeePerGas is null");
+  if (maxPriorityFeePerGasEstimate == null)
+    throw new Error("rundler_maxPriorityFeePerGas returned null or undefined");
+
+  const maxPriorityFeePerGas =
+    typeof maxPriorityFeePerGasEstimate === "bigint"
+      ? maxPriorityFeePerGasEstimate
+      : isHex(maxPriorityFeePerGasEstimate)
+        ? fromHex(maxPriorityFeePerGasEstimate as Hex, "bigint")
+        : throwError(`Invalid hex value: ${maxPriorityFeePerGasEstimate}`);
```
jakehobbs: seems `throwError` is not defined anywhere

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export const alchemyRpcSchema = rpcSchema<AlchemyRpcSchema>();
+
+// Extend viem's Client with a typed rundler RPC method.
+export type PriorityFeeClient = Client & {
+  request(args: {
+    method: "rundler_maxPriorityFeePerGas";
+    params: [];
+  }): Promise<UserOperationRequest["maxPriorityFeePerGas"]>;
+};
```
jakehobbs: viem has a generic that supports the rpc schema, so you shouldn't need to do a union here.
```suggestion
export type PriorityFeeClient<
  transport extends Transport = Transport,
  chain extends Chain | undefined = Chain | undefined,
  account extends Account | undefined = Account | undefined,
> = Client<transport, chain, account, AlchemyRpcSchema>;
```

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
```diff
@@ -0,0 +1,86 @@
...
+ *   },
+ * });
+ * ```
+ */
+export async function alchemyEstimateFeesPerGas({
+  bundlerClient,
+  account: _account,
+  userOperation: _userOperation,
+}: {
+  bundlerClient: Client;
```
jakehobbs: i think the way you are using this w/ `createBundlerClient` is smart enough for the type here to be correct & then you can remove the casting
```suggestion
  bundlerClient: PriorityFeeClient;
```

packages/smart-accounts/tests/coinbase-account.test.ts
```diff
@@ -39,10 +46,15 @@ describe("Viem AA - Coinbase Smart Account", () => {
       owners: [owner],
     });
 
+    // Create a bundler client that uses the Alchemy fee estimator
     const bundlerClient = createBundlerClient({
```
jakehobbs: ✨ should we create some function like `createAlchemyBundlerClient` that automatically includes the rpcSchema, userOperation params, and extended actions?

fine w/ leaving as-is for now though since it's just being used in tests.

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
```diff
@@ -0,0 +1,86 @@
+import { getBlock } from "viem/actions";
+import { fromHex, type Hex, type Client, rpcSchema } from "viem";
+import type {
+  UserOperationRequest,
+  SmartAccount,
+} from "viem/account-abstraction";
+
+// Minimal RPC schema that only declares the Alchemy priority fee method.
+export type AlchemyRpcSchema = [
```
jakehobbs: should we move the rpc schema & types out to a separate file since i assume we'll be adding additional rundler methods here soon?

packages/smart-accounts/tests/coinbase-account.test.ts
trevoraron: @jakehobbs ah that was my original approach (i had like an AlchemyBundlerFactory) but @adamegyed suggested I remove it

I can bring it back and just have both

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
trevoraron: good idea!

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
trevoraron: lol whoops - fixed :)

trevoraron:
ok looks like i broke something during the process of rebasing hahaha, looking into this

trevoraron:
ok should be good to go now! @jakehobbs added a `createAlchemyBundlerClient` function :P

packages/smart-accounts/src/types.ts
```diff
@@ -8,3 +11,12 @@ export type GetAccountParameter<
   IsUndefined<TAccount> extends true
     ? { account: TAccountOverride }
     : { account?: TAccountOverride };
+
+// Minimal RPC schema that only declares the Alchemy priority fee method.
+export type AlchemyRpcSchema = [
+  {
+    Method: "rundler_maxPriorityFeePerGas";
+    Parameters: [];
+    ReturnType: UserOperationRequest["maxPriorityFeePerGas"];
```
adamegyed: I don't think this is right, the RPC method returns hex, but this indexed type is a `bigint` (using the default generic arg of `uint256 = bigint` in `UserOperationRequest`. We should update this to use `RpcUserOperation` instead of `UserOperationRequest`.

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
```diff
@@ -0,0 +1,90 @@
...
+  const maxPriorityFeePerGas =
+    typeof maxPriorityFeePerGasEstimate === "bigint"
+      ? maxPriorityFeePerGasEstimate
+      : isHex(maxPriorityFeePerGasEstimate)
+        ? fromHex(maxPriorityFeePerGasEstimate as Hex, "bigint")
+        : (() => {
+            throw new Error(
+              `Invalid hex value: ${maxPriorityFeePerGasEstimate}`,
+            );
+          })();
```
adamegyed: If the RPC schema were correctly typed, this wouldn't be needed. I think after line 78 the TS compiler would currently deduce the type of `maxPriorityFeePerGasEstimate` as `never`.

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  return {
+    maxPriorityFeePerGas,
+    maxFeePerGas: (baseFeePerGas * 150n) / 100n + maxPriorityFeePerGas,
```
adamegyed: I think viem's default behavior is to do this 1.5x multiple on basefee - is there any way we could reuse that behavior, instead of re-implementing? Does viem only expose `estimateFeesPerGas` in the bundler client config?

packages/smart-accounts/src/clients.ts
```diff
@@ -0,0 +1,45 @@
...
+ * • Extends the RPC schema with {@link alchemyRpcSchema} so the new method is
+ *   typed.
+ *
+ * @param {object} params                    – configuration object.
+ * @param {SmartAccount} params.account      – Smart account instance for the client.
+ * @param {Chain | undefined} params.chain   – Chain the bundler operates on.
+ * @param {Transport} params.transport       – viem transport (eg. `custom(client)`).
+ * @returns {ReturnType<typeof createBundlerClient>} a Bundler Client ready to use with Alchemy’s Rundler.
+ */
+export function createAlchemyBundlerClient<
```
adamegyed, line 24: This is fine to keep for now, but I'm not sure we need a `AlchemyBundlerClient` - it may be sufficient to tell users to use `createBundlerClient` with the `alchemyFeeEstimator`. Could you leave a `TODO(v5)` commet to revisit this export?

packages/smart-accounts/tests/coinbase-account.test.ts
jakehobbs: oh interesting. i don't have the most solid grasp on how all of the pieces are fit together yet, so i'll def defer to him.

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
```diff
@@ -0,0 +1,90 @@
...
+  userOperation: _userOperation,
+}: {
+  bundlerClient: PriorityFeeClient;
+  account?: SmartAccount;
+  userOperation?: UserOperationRequest;
+}): Promise<{
+  maxFeePerGas: bigint;
+  maxPriorityFeePerGas: bigint;
+}> {
+  const rpcClient = bundlerClient as PriorityFeeClient;
```
jakehobbs: you should be able to get rid of this casting & just use the `bundlerClient` param directly now that the type is correct

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (maxPriorityFeePerGasEstimate == null)
+    throw new Error("rundler_maxPriorityFeePerGas returned null or undefined");
+
+  const maxPriorityFeePerGas =
+    typeof maxPriorityFeePerGasEstimate === "bigint"
+      ? maxPriorityFeePerGasEstimate
+      : isHex(maxPriorityFeePerGasEstimate)
+        ? fromHex(maxPriorityFeePerGasEstimate as Hex, "bigint")
+        : (() => {
+            throw new Error(
```
jakehobbs: this should prob be an error that extends `BaseError` from `@alchemy/common`

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
trevoraron: nah its used by a bunch of clients (createPublicClient, createWalletClient, createTestClient). however you can override it for the bundler for user-ops (what we need for the runder calls here)

unfortunately not a good way to reuse (we can call viems full estimateFeePerGas and then swap in our rundler call but that would make an extra call -- this way we skip the call to `eth_gasPrice`

actually viems behavior is 1.2x !

trevoraron:
ok now everything should be good to go!!!

except for fern - docs gen. am I supposed to run something here?

packages/smart-accounts/src/alchemyEstimateFeesPerGas.ts
```diff
@@ -0,0 +1,100 @@
...
+  type Transport,
+  type Chain,
+  type Account,
+} from "viem";
+import type {
+  UserOperationRequest,
+  SmartAccount,
+} from "viem/account-abstraction";
+import { BaseError } from "@alchemy/common";
+import { bigIntMultiply } from "@aa-sdk/core";
```
jakehobbs: we shouldn't be importing from @aa-sdk/core in v5, so we should move this to a new pkg or add a TODO.

adamegyed (approved):
LGTM

PR #1831, opened 2025-07-31
feat(v5): add wallet api scaffold and initial gas manager action

adamegyed:
New package created with `yarn halp add-package -p wallet-apis -d "Alchemy Wallet APIs"` using `type = JSON-RPC`.

Replaced the dummy action from the scaffold with an action for `alchemy_requestGasAndPaymasterAndData`, following the action, decorator, RpcSchema pattern set by viem.

We probably don't want to recommend using this action to end users, as it is just a type-safe wrapper over the underlying JSON-RPC. Instead, we will want to port the `alchemyGasManagerMiddleware` to a viem-equivalent custom paymaster / custom gas estimation function compatible with viem-style smart accounts. This action can be cosumed within those functions.

Also added a simple test of the action to verify that it works, and replaced the deprecated `getBytecode` action from the test harness with `getCode`.

# Pull Request Checklist

---

## PR-Codex overview
This PR introduces the `@alchemy/wallet-apis` package, adding functionality for interacting with Alchemy Wallet APIs, including gas management. It also updates configuration files, adds documentation, and includes tests for new methods.

### Detailed summary
- Added `VERSION` export in `version.ts`.
- Updated `tsconfig.json` for TypeScript configurations.
- Enhanced `packageMap.ts` with new entries for Alchemy Wallet APIs.
- Created `vitest.config.ts` for testing configuration.
- Added reference documentation for `@alchemy/wallet-apis`.
- Implemented `requestGasAndPaymasterAndData` method in `requestGasAndPaymasterAndData.ts`.
- Created `gasManagerActions` decorator in `gasManager.ts`.
- Updated `types.ts` for request and response types.
- Added utility functions in `utils.ts` for formatting requests.
- Introduced tests for `requestGasAndPaymasterAndData` functionality.


.vitest/src/paymaster/transport.ts
```diff
@@ -66,20 +66,24 @@ export const paymasterTransport = (
...
-            {
-              policyId: string;
-              entryPoint: Address;
-              dummySignature: Hex;
-              userOperation: UserOperationRequest;
-              overrides?: UserOperationOverrides;
-            },
-          ];
+          const [{ userOperation, entryPoint, dummySignature, overrides }] =
+            args.params as [
```
adamegyed, line 74: Reviewer note: there's some bad type-casting happening here, because viem and aa-sdk/core's concept of a UserOperationRequest is slightly different, but so far the only issue I've needed to patch is loading the dummy signature into the UO's signature. More may come up as we increase test coverage.

packages/wallet-apis/src/schema.ts
```diff
@@ -0,0 +1,126 @@
...
+    ? {
+        paymasterPostOpGasLimit:
+          | RpcUserOperation<TEntryPointVersion>["paymasterPostOpGasLimit"]
+          | Multiplier;
+      }
+    : {})
+>;
+
+// Once you've defined this schema, import it into the schema.ts file found in the root alchemy package
+export type AlchemyWalletApisRpcSchema = [
```
adamegyed: This schema definition format is following viem's pattern in `eip1193.ts`, but it does lead to some awkward lookups, like `AlchemyWalletApisRpcSchema[0]["ReturnType"]` in this PR's `utils.ts`, which feels fragile by indexing into the RPC schema using the order of methods defined. Is there a cleaner way to do this? Maybe just have additional `Rpc...` exports?

jakehobbs: yeah, i think the cleanest thing is prob to extract individual schemas, i.e.
```ts
export type AlchemyRequestGasAndPaymasterAndData = { ... }

export type AlchemyWalletApisRpcSchema = [AlchemyRequestGasAndPaymasterAndData];
```

.vitest/src/paymaster/transport.ts
jakehobbs: should we leave a comment in the code about this?

adamegyed: Yep, added

packages/wallet-apis/src/schema.ts
adamegyed: Done!

packages/wallet-apis/src/schema.ts
```diff
@@ -0,0 +1,128 @@
+import type { Address, Hex, PartialBy } from "viem";
+import type { Multiplier } from "./actions/types";
```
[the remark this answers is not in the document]
adamegyed: Good bot.

I thought we had a linter rule for this? Updated.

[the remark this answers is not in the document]
jakehobbs: not sure if we have an eslint rule, but i think build would have failed?

packages/wallet-apis/src/actions/requestGasAndPaymasterAndData.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        {
+          target: zeroAddress,
+          data: "0x" as const,
+        },
+      ]),
+      signature: await smartAccount.getDummySignature(),
+    };
+
+    const result = await requestGasAndPaymasterAndData(
+      client as unknown as Client<AlchemyTransport>,
```
trevoraron, line 86: is this as unknown needed?

packages/wallet-apis/src/decorators/gasManager.ts
```diff
@@ -0,0 +1,29 @@
...
+  ) => Promise<RequestGasAndPaymasterAndDataResponse>;
+};
+
+/**
+ * This is a decorator that is used to add gas manager actions to a client.
+ *
+ * @param {object} client - The client to add the actions to.
+ * @returns {object} - The client with the actions added.
+ */
+export const gasManagerActions: <
```
trevoraron, line 21: question: how is this planning on being used? i don't see us using this decorator anyone beyond exporting it

packages/wallet-apis/src/schema.ts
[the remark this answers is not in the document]
adamegyed: it did not - maybe because this is just a type import?

packages/wallet-apis/src/actions/requestGasAndPaymasterAndData.test.ts
adamegyed: Yes, otherwise the client action complains about the transport types not matching between `custom` and `alchemy`:
```
Conversion of type 'Client<(<chain extends Chain | undefined = Chain>(args: { chain?: chain | undefined; pollingInterval?: number | undefined; retryCount?: number | undefined; timeout?: number | undefined; }) => { ...; }), { ...; }, undefined, undefined, { ...; }>' to type 'Client<AlchemyTransport>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'Client<(<chain extends Chain | undefined = Chain>(args: { chain?: chain | undefined; pollingInterval?: number | undefined; retryCount?: number | undefined; timeout?: number | undefined; }) => { ...; }), { ...; }, undefined, undefined, { ...; }>' is not comparable to type 'Client_Base<AlchemyTransport, Chain | undefined, Account | undefined, undefined>'.
    Types of property 'transport' are incompatible.
      Type 'TransportConfig<"custom", EIP1193RequestFn>' is not comparable to type 'TransportConfig<"alchemy", EIP1193RequestFn> & { alchemyRpcUrl: string; fetchOptions: Omit<RequestInit, "body"> | undefined; config: AlchemyTransportConfig; makeHttpRequest: HttpRequestFn<...>; }'.
        Type 'TransportConfig<"custom", EIP1193RequestFn>' is not comparable to type 'TransportConfig<"alchemy", EIP1193RequestFn>'.
          Type '"custom"' is not comparable to type '"alchemy"'.
```

Tbh, this should be reason to reconsider the scaffolding setup that requires actions to be attached to clients with `AlchemyTransport`s, because it's both a leaky abstraction and clumsy to work with.

packages/wallet-apis/src/decorators/gasManager.ts
adamegyed: We can use this in the custom function to modify the client.

We could have use this in the test file, but unfortunately the `custom` transport setup prevents the decorator from working correctly. Something to consider changing while we're working on v5.

PR #1838, opened 2025-08-02
fix: typebox metro config

jakehobbs:
# Pull Request Checklist

---

## PR-Codex overview
This PR updates the `metro.config.js` file for a React Native Expo example to include new module resolutions and modifies the formatting of existing code.

### Detailed summary
- Added a subproject commit reference.
- Updated the resolution path for `@noble/hashes/crypto` to include `node_modules/@noble/hashes/crypto.js`.
- Added a resolution path for `@sinclair/typebox` to `node_modules/@sinclair/typebox/build/cjs/index.js`.
- Changed the formatting of the `crypto` resolution to use double quotes.


examples/react-native-expo-example/metro.config.js
```diff
@@ -12,7 +12,11 @@ const config = getDefaultConfig(projectRoot);
 const ALIASES = {
 	"@noble/hashes/crypto": path.resolve(
 		monorepoRoot,
-		"node_modules/@noble/hashes/crypto.js"
+		"node_modules/@noble/hashes/crypto.js",
+	),
+	"@sinclair/typebox": path.resolve(
```
blakecduncan, line 17: Interesting that we aren't seeing the issue here with `ox/erc6492/WrappedSignature.ts` since this metro config doesn't have my fix from the other example.  Do you know if we updated the `ox` package?  I wonder if there was something specific that fixed the ox package.

Curious to get your thoughts, I think this might be a fix where we won't have to hardcode each individual package.  Happy to move forward with your fix though if you think the Alias approach is better.

https://github.com/alchemyplatform/react-native-expo-example/blob/da3f3ff392ae46655b23da786de212b3fd8752ce/metro.config.js#L36-L58

jakehobbs: oh, i think that ox issue is no longer an issue since we removed "local mode" from the wallet client – so now that is only used on the server.

jakehobbs: do you see this being an issue w/ more pkgs going forward? i think a fix for the typebox pkg is to just not use it directly within the wallet client, which should not require any huge changes.

blakecduncan: Yeah good questions - I was just thinking we've seen if for two packages so far.  But we probably aren't going to be adding a lot more internal packages going forward so it might not be an issue

florrdv (approved):

PR #1842, opened 2025-08-03
chore: create tailwind styles during build process, so there's no need for the end user to install

androbwebb:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing the `account-kit/react` project by updating configurations, adding Tailwind CSS support, and improving package management.

### Detailed summary
- Added Tailwind CSS directives in `index.css`.
- Updated `package.json` to specify Node.js engine version and added a `build:css` script.
- Modified `plugin.ts` to provide a fallback for package resolution.
- Enhanced `package.json` with new paths for CSS and types.
- Updated dependencies in `yarn.lock`, including `postcss-cli`, `tailwindcss`, and others.


account-kit/react/src/styles/index.css
```diff
@@ -0,0 +1,3 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
```
jakehobbs: wasn't this removed in tailwind v4?

https://tailwindcss.com/docs/upgrade-guide#removed-tailwind-directives

jakehobbs: oh wait nvm, i forgot we are still using v3 internally.

blakecduncan (commented):
Curious to hear your thoughts of if we should add this to our tailwind config.  This would make sure that our styles don't conflict with the consuming apps classes.

```
Okay so we just need to do this

// tailwind.config.js
module.exports = {
  prefix: 'alchemy-',
  /* … */
}

```

androbwebb:
> Curious to hear your thoughts of if we should add this to our tailwind config. This would make sure that our styles don't conflict with the consuming apps classes.
> 
> ```
> // tailwind.config.js
> module.exports = {
>   prefix: 'alchemy-',
>   /* … */
> }
> ```

Yeah, I agree completely. Will add.

blakecduncan:
> i assume devs will still needs to install tailwind themselves if they want to customize the colors, since that's handled in `withAccountKitUi` which wraps the tailwind config?
> 
> i think that's fine for now, but i think it's probably something we should revisit for v5. curious if you have any thoughts on that.
> 
> although i'm not sure how popular it is for someone to customize the default UI component colors. i think most apps i've seen either use the default styling OR build fully custom components.

@jakehobbs it's been a while since I've looked into this but I think there may be a way for us to allow customizing colors with css variables.  I think the tailwind theme tokens will be exported as variables in the css file that our build produces and the consuming app can just override those.

Not sure if this is our ideal solution though.  Agree we should revisit for v5

androbwebb:
> i assume devs will still needs to install tailwind themselves if they want to customize the colors, since that's handled in `withAccountKitUi` which wraps the tailwind config?
> 
> i think that's fine for now, but i think it's probably something we should revisit for v5. curious if you have any thoughts on that.
> 
> although i'm not sure how popular it is for someone to customize the default UI component colors. i think most apps i've seen either use the default styling OR build fully custom components.

Yeah that's a fair point. Sounds like we're ok bumping that to V5 though, so I'll skip over it for this PR

androbwebb:
> > i assume devs will still needs to install tailwind themselves if they want to customize the colors, since that's handled in `withAccountKitUi` which wraps the tailwind config?
> > i think that's fine for now, but i think it's probably something we should revisit for v5. curious if you have any thoughts on that.
> > although i'm not sure how popular it is for someone to customize the default UI component colors. i think most apps i've seen either use the default styling OR build fully custom components.
> 
> Yeah that's a fair point. Sounds like we're ok bumping that to V5 though, so I'll skip over it for this PR

Ok, actually... going to push this to v5. It requires manually going and changing every single class in the components to have an `alchemy-` prefix. Idk why I assumed that would be done by the build process.

androbwebb:
> > Curious to hear your thoughts of if we should add this to our tailwind config. This would make sure that our styles don't conflict with the consuming apps classes.
> > ```
> > // tailwind.config.js
> > module.exports = {
> >   prefix: 'alchemy-',
> >   /* … */
> > }
> > ```
> 
> Yeah, I agree completely. Will add.

Ok, actually... going to push this to v5. It requires manually going and changing every single class in the components to have an `alchemy-` prefix. Idk why I assumed that would be done by the build process.

blakecduncan:
> > > Curious to hear your thoughts of if we should add this to our tailwind config. This would make sure that our styles don't conflict with the consuming apps classes.
> > > ```
> > > // tailwind.config.js
> > > module.exports = {
> > >   prefix: 'alchemy-',
> > >   /* … */
> > > }
> > > ```
> > 
> > 
> > Yeah, I agree completely. Will add.
> 
> Ok, actually... going to push this to v5. It requires manually going and changing every single class in the components to have an `alchemy-` prefix. Idk why I assumed that would be done by the build process.

Oh damn 😦 I actually had the same assumption.  But yeah that makes sense

PR #1854, opened 2025-08-08
draft: gas manager middleware migation

trevoraron:
migrate the gas manager to use our viem middleware

Assana Task: https://app.asana.com/1/1129441638109975/project/1210727695159716/task/1210801854306015?focus=true


packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,315 @@
...
+        } = parameters;
+
+        // Prepare user operation
+        const userOp = deepHexlify(
+          await resolveProperties(userOpFields as any),
+        );
+
+        // We need the account to get dummy signature, but viem doesn't provide it
+        // For now, we'll use a default dummy signature
+        const dummySignature =
```
trevoraron, line 190: we don't have the `account` object here so we can't call `account.getDummySignature()` -- @adamegyed any problems with this?

packages/smart-accounts/package.json
```diff
@@ -43,6 +43,7 @@
     "typescript-template": "*"
   },
   "dependencies": {
+    "@account-kit/infra": "*",
```
jakehobbs: what do we need this for?

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,315 @@
...
+import { getBlock } from "viem/actions";
+import type {
+  GetPaymasterDataParameters,
+  GetPaymasterDataReturnType,
+  GetPaymasterStubDataParameters,
+  GetPaymasterStubDataReturnType,
+} from "viem/account-abstraction";
+import { deepHexlify, resolveProperties } from "@aa-sdk/core";
+import { bigIntMultiply } from "@alchemy/common";
+// Type for ERC-20 token context
```
jakehobbs: nit: whitespace
```suggestion
import { bigIntMultiply } from "@alchemy/common";

// Type for ERC-20 token context
```

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,315 @@
...
+  policyToken?: PolicyToken,
+) {
+  const context = createGasManagerContext(policyId, policyToken);
+
+  // Cache to store the result between calls
+  let cachedResult: GasAndPaymasterAndDataResponse | null = null;
+  let cachedUserOpHash: string | null = null;
+
+  // Helper to create a hash of the user operation for cache key
+  const getUserOpHash = (params: any): string => {
```
jakehobbs: can we type this?
```suggestion
  const getUserOpHash = (params: Pick<UserOperationRequest, "sender" | "nonce" | "callData">): string => {
```

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,315 @@
...
+  };
+
+  return {
+    paymaster: (bundlerClient: Client<Transport, Chain>) => ({
+      async getPaymasterStubData(
+        parameters: GetPaymasterStubDataParameters,
+      ): Promise<GetPaymasterStubDataReturnType> {
+        const userOpHash = getUserOpHash(parameters);
+
+        // If we have a cached result for this user op, return it
```
jakehobbs: could this cause issues if the paymaster data expired?

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,315 @@
...
+        const {
+          chainId,
+          entryPointAddress,
+          context: userContext,
+          ...userOpFields
+        } = parameters;
+
+        // Prepare user operation
+        const userOp = deepHexlify(
+          await resolveProperties(userOpFields as any),
```
jakehobbs: do we need to use `resolveProperties` here still? i don't think the viem UO type has any promises in it.

packages/smart-accounts/src/alchemyGasManagerHooks.ts
jakehobbs: can you type `bundlerClient` as `Client<Transport, Chain, SmartAccount>`, then use `bundlerClient.account.getStubSignature()`? or does that break something?

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,315 @@
...
+          userOperation: userOp,
+          dummySignature,
+          overrides: {},
+          ...(context.erc20Context
+            ? { erc20Context: context.erc20Context }
+            : {}),
+        };
+
+        // Make the optimized RPC call
+        const response = await bundlerClient.request<{
```
jakehobbs: should we handle the pm rpc method types similarly to how you did in the `alchemyEstimateFeesPerGas` function? or do you prefer to do it inline like this?

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,315 @@
...
+        if ("paymasterAndData" in stubResult && stubResult.paymasterAndData) {
+          return { paymasterAndData: stubResult.paymasterAndData };
+        }
+
+        return {
+          paymaster: stubResult.paymaster!,
+          paymasterData: stubResult.paymasterData!,
+          paymasterVerificationGasLimit:
+            stubResult.paymasterVerificationGasLimit!,
+          paymasterPostOpGasLimit: stubResult.paymasterPostOpGasLimit!,
```
jakehobbs: do you need these non-null assertions? seems like these fields are already non-nullable.

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,315 @@
...
+      },
+    }),
+
+    userOperation: {
+      // Custom fee estimator that uses the cached gas values or Alchemy's fee estimation
+      async estimateFeesPerGas(params: { bundlerClient: Client }) {
+        // If we have cached gas values, return them
+        if (cachedResult) {
+          return {
+            maxFeePerGas: hexToBigInt(cachedResult.maxFeePerGas),
```
jakehobbs: is this safe to cache? i wouldn't think so unless the cache has a short expiry.

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,315 @@
...
+              throw new Error(
+                `Invalid hex value: ${maxPriorityFeePerGasEstimate}`,
+              );
+            })();
+
+        return {
+          maxPriorityFeePerGas,
+          maxFeePerGas:
+            bigIntMultiply(baseFeePerGas, 1.5) + maxPriorityFeePerGas,
+        };
```
jakehobbs: should just be able to use `alchemyEstimateFeesPerGas` here if not using cache?


```suggestion
        return alchemyEstimateFeesPerGas(params)
```

packages/smart-accounts/package.json
trevoraron: ah i can remove this! i had the package somewhere else and was resolving all sorts of dependency issues

packages/smart-accounts/src/alchemyGasManagerHooks.ts
trevoraron: @jakehobbs the goal is rather than make 2 calls, use our single call. unfortunately viem still requires two functions, hence the caching

packages/smart-accounts/src/alchemyGasManagerHooks.ts
trevoraron: @jakehobbs tbh now that im thinking about this, this cache should invalidate after the second call (so subsequent calls don't use the same cache)

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,286 @@
+import type { Address, Client, Transport, Chain, Hex } from "viem";
+import { hexToBigInt, fromHex, isHex } from "viem";
```
gpuliyar, line 2: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'fromHex' is defined but never used.

gpuliyar, line 2: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'isHex' is defined but never used.

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,286 @@
...
+import { getBlock } from "viem/actions";
```
gpuliyar, line 3: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'getBlock' is defined but never used.

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,286 @@
...
+import type {
+  GetPaymasterDataParameters,
+  GetPaymasterDataReturnType,
+  GetPaymasterStubDataParameters,
+  GetPaymasterStubDataReturnType,
+} from "viem/account-abstraction";
+import { deepHexlify, resolveProperties } from "@aa-sdk/core";
+import { bigIntMultiply } from "@alchemy/common";
```
gpuliyar, line 11: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'bigIntMultiply' is defined but never used.

trevoraron:
@adamegyed  @jakehobbs  this should be ready for re-review! (realized i had some merge conflicts)

big thing -- added a cache object that was originally a map from hash -> result. but then realized this should lead to merge conflicts, and people don't often call TX's in paralell. so now its scoped to support that one flow (first call gets data and paymaster stuff, caches, rest just pull that and return)

account-kit/infra/src/index.ts
```diff
@@ -63,6 +63,7 @@ export type * from "./client/types.js";
 export { getDefaultUserOperationFeeOptions } from "./defaults.js";
 export { getAlchemyPaymasterAddress } from "./gas-manager.js";
 export { alchemyFeeEstimator } from "./middleware/feeEstimator.js";
+
```
adamegyed: omit?

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,318 @@
...
+  return context;
+}
+
+/**
+ * Adapts Alchemy's Gas Manager to viem's paymaster interface using the optimized
+ * `alchemy_requestGasAndPaymasterAndData` flow that combines gas estimation and
+ * paymaster sponsorship in a single RPC call.
+ *
+ * This implementation caches the result from the first call to avoid duplicate RPC calls
+ * when viem calls both getPaymasterStubData and getPaymasterData.
```
adamegyed: nit: viem calls both `estimateFeesPerGas` and `paymaster` functions

packages/smart-accounts/package.json
```diff
@@ -43,6 +43,8 @@
...
```
adamegyed: omit?

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,318 @@
...
+): Promise<any> {
+  const { entryPointAddress, ...userOpFields } = parameters;
+
+  // Prepare user operation
+  const userOp = deepHexlify(userOpFields);
+
+  // Get dummy signature from the account if available, otherwise use default
+  const dummySignature = bundlerClient.account?.getStubSignature
+    ? await bundlerClient.account.getStubSignature(userOp)
+    : DEFAULT_DUMMY_SIGNATURE;
```
adamegyed: Why should we have a default here? Essentially all smart accounts have an encoding different from this.

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,318 @@
...
+
+// Create a hash of the user operation for cache key
+function getUserOpHash(
+  params: GetPaymasterStubDataParameters | GetPaymasterDataParameters,
+): string {
+  return JSON.stringify({
+    sender: params.sender,
+    nonce: params.nonce?.toString(), // Convert bigint to string for stable serialization
+    callData: params.callData,
+  });
```
adamegyed: We should either be using the canonical UO hash calculation from the entrypoint (or a local copy), or name this variable something else

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,318 @@
...
+      paymaster: response.paymaster,
+      paymasterData: response.paymasterData || "0x",
+      paymasterVerificationGasLimit:
+        response.paymasterVerificationGasLimit ?? 100000n,
+      paymasterPostOpGasLimit: response.paymasterPostOpGasLimit ?? 50000n,
+      isFinal: true,
+    };
+  }
+
+  throw new Error("No paymaster data returned from optimized call");
```
adamegyed: We probably want this to have a different name from "optimized call".

Also, I think this is possible to do with either an `assertNever` or regular typeguards, since there are only 2 possible shapes for the response

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,318 @@
...
+    context.erc20Context = {
+      tokenAddress: policyToken.address,
+      maxTokenAmount: policyToken.maxTokenAmount,
+    };
+
+    if (policyToken.permit !== undefined) {
+      console.warn(
+        "⚠️ ERC-20 permits are not supported in viem-native gas manager hooks. " +
+          "Use alchemyGasManagerMiddleware() for permit support.",
+      );
```
adamegyed: could you share why this isn't supported?

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,318 @@
...
+ */
+export function alchemyGasManagerHooks(
+  policyId: string | string[],
+  policyToken?: PolicyToken,
+) {
+  const context = createGasManagerContext(policyId, policyToken);
+  const cache = new UserOpCache();
+
+  return {
+    paymaster: (bundlerClient: Client<Transport, Chain, SmartAccount>) => ({
```
adamegyed: we shouldn't call this `bundlerClient`, since it's just a `Client` type. Maybe just `client` or `paymasterClient`?

packages/smart-accounts/src/alchemyGasManagerHooks.ts
trevoraron: removed the default!

packages/smart-accounts/src/alchemyGasManagerHooks.ts
trevoraron: named something else!

packages/smart-accounts/src/alchemyGasManagerHooks.ts
trevoraron: i think because viem hooks are more lower level

with the hooks by the time they are called the user op is already set, and we can only add paymaster data (can't modify calldata to add the permit)

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,303 @@
...
+): string {
+  return JSON.stringify({
+    sender: params.sender,
+    nonce: params.nonce?.toString(), // Convert bigint to string for stable serialization
+    callData: params.callData,
+  });
+}
+
+// Build the wallet-apis request
+async function buildWalletApisRequest(
```
adamegyed: we can inline this

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,303 @@
...
+  parameters: GetPaymasterStubDataParameters | GetPaymasterDataParameters,
+  client: Client<Transport, Chain, SmartAccount>,
+  context: AlchemyGasManagerContext,
+): Promise<any> {
+  const { entryPointAddress, ...userOpFields } = parameters;
+
+  // Prepare user operation
+  const userOp = deepHexlify(userOpFields);
```
adamegyed: we should avoid `deepHexlify` for two reasons: 1) it comes from `aa-sdk/core`, and 2) it removes the typing on the return. We should explicitly await any promises that must be resolved here.

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,303 @@
...
+
+  // Get dummy signature from the account
+  if (!client.account?.getStubSignature) {
+    throw new Error(
+      "Account must implement getStubSignature for gas manager hooks",
+    );
```
adamegyed: This is probably the wrong error message, because `getStubSignature` is required on the `SmartAccount` type. This condition is only met if the account is undefined, so we should make the error match the cause.

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,303 @@
...
+      paymasterVerificationGasLimit:
+        response.paymasterVerificationGasLimit ?? 100000n,
+      paymasterPostOpGasLimit: response.paymasterPostOpGasLimit ?? 50000n,
+      isFinal: true,
+    };
+  }
+}
+
+// Convert response to data return type
+function toDataReturn(
```
adamegyed: can also inline this where it's used

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,303 @@
...
+): AlchemyGasManagerContext {
+  const context: AlchemyGasManagerContext = { policyId };
+
+  if (policyToken !== undefined) {
+    context.erc20Context = {
+      tokenAddress: policyToken.address,
+      maxTokenAmount: policyToken.maxTokenAmount,
+    };
+
+    if (policyToken.permit !== undefined) {
```
adamegyed: could we move this to the function description and omit the `permit` field entirely, to make this code path impossible to hit?

packages/smart-accounts/src/alchemyGasManagerHooks.ts
```diff
@@ -0,0 +1,303 @@
...
+  maxTokenAmount?: bigint;
+  permit?: {
+    deadline: bigint;
+    nonce: bigint;
+    domain?: any;
+  };
+};
+
+// Response is provided by wallet-apis (already bigint-formatted)
+type GasAndPaymasterAndDataResponse = {
```
adamegyed: This already exists as the type `RequestGasAndPaymasterAndDataResponse` in aa-infra, let's reuse it instead of re-declaring

packages/smart-accounts/tests/alchemyGasManagerHooks.test.ts
```diff
@@ -0,0 +1,366 @@
...
+            verificationGasLimit: "0x5208",
+            maxFeePerGas: "0x3b9aca00",
+            maxPriorityFeePerGas: "0xf4240",
+            paymaster: "0x1234567890abcdef1234567890abcdef12345678",
+            paymasterData: "0xabcdef",
+            paymasterVerificationGasLimit: "0x186a0",
+            paymasterPostOpGasLimit: "0xc350",
+          });
+        }
+        return Promise.resolve(null);
```
adamegyed: nit: I think it's simpler to make the outer function async than wrapping each return statement in `Promise.resolve`

PR #1858, opened 2025-08-11
feat: export wallet

androbwebb:
# Pull Request Checklist

Screenshot 2025-08-29 at 8 48 37 PM

---

## PR-Codex overview
This PR introduces significant changes to the wallet export functionality, transitioning from a boolean return type to a `Promise<unknown>`, enhancing the export process with improved error handling and support for different export formats.

### Detailed summary
- Added `ExportWalletOutput` type as `boolean`.
- Updated `exportWallet` methods to return `Promise<unknown>`.
- Modified `BaseSignerClient` to handle different export types and outputs.
- Enhanced error handling for wallet export operations.
- Refactored `RNSignerClient` to support exporting both private keys and seed phrases.


account-kit/rn-signer/examples/export-wallet-example.ts
```diff
@@ -0,0 +1,183 @@
...
+import { Alert } from "react-native";
+
+/**
+ * Example: Export private key using the local storage approach
+ * 
+ * This implementation uses Turnkey's recommended approach for mobile:
+ * 1. Generate a P256 key pair locally
+ * 2. Request export encrypted to the public key
+ * 3. Decrypt the bundle using HPKE with the private key
+ * 4. Clean up the key from storage
```
gpuliyar, line 11: 🚫 **[eslint]** <[jsdoc/require-returns](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-returns.md#repos-sticky-header)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Missing JSDoc @returns declaration.
```suggestion
/**
 * Example: Export private key using the local storage approach
 * 
 * This implementation uses Turnkey's recommended approach for mobile:
 * 1. Generate a P256 key pair locally
 * 2. Request export encrypted to the public key
 * 3. Decrypt the bundle using HPKE with the private key
 * 4. Clean up the key from storage
 * @returns
```

account-kit/rn-signer/examples/export-wallet-example.ts
```diff
@@ -0,0 +1,183 @@
...
+    Alert.alert(
+      "Export Failed",
+      "Unable to export wallet. Please ensure you are authenticated and try again."
+    );
+    throw error;
+  }
+}
+
+/**
+ * Example: Export seed phrase
```
gpuliyar, line 62: 🚫 **[eslint]** <[jsdoc/require-returns](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-returns.md#repos-sticky-header)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Missing JSDoc @returns declaration.
```suggestion
/**
 * Example: Export seed phrase
 * @returns
```

account-kit/rn-signer/examples/export-wallet-example.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    return result;
+  } catch (error) {
+    console.error("Failed to export seed phrase:", error);
+    throw error;
+  }
+}
+
+/**
+ * Example: Using the base exportWallet method (returns boolean)
```
gpuliyar, line 107: 🚫 **[eslint]** <[jsdoc/require-returns](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-returns.md#repos-sticky-header)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Missing JSDoc @returns declaration.
```suggestion
/**
 * Example: Using the base exportWallet method (returns boolean)
 * @returns
```

account-kit/rn-signer/examples/export-wallet-example.ts
```diff
@@ -0,0 +1,183 @@
...
+    
+    return success;
+  } catch (error) {
+    console.error("Export failed:", error);
+    throw error;
+  }
+}
+
+/**
+ * Example: Direct client usage
```
gpuliyar, line 137: 🚫 **[eslint]** <[jsdoc/require-returns](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-returns.md#repos-sticky-header)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Missing JSDoc @returns declaration.
```suggestion
/**
 * Example: Direct client usage
 * @returns
```

account-kit/rn-signer/src/client.ts
```diff
@@ -2,8 +2,6 @@
 import "react-native-get-random-values";
 import "./utils/buffer-polyfill";
 import "./utils/mmkv-localstorage-polyfill";
-
-/* eslint-disable import/extensions */
```
androbwebb, line 6: This is already set up above.

account-kit/rn-signer/src/client.ts
```diff
@@ -274,8 +286,161 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
...
+        }
+
+        exportBundle = result.exportBundle;
+      }
+
+      // Step 4: Parse the export bundle and decrypt using HPKE
+      // The export bundle is a JSON string containing version, data, etc.
+      const bundleJson = JSON.parse(exportBundle);
+
+      // The data field contains another JSON string that's hex-encoded
```
androbwebb, line 382: This is weirdly complicated, there's a stringified json nested inside a json.

account-kit/rn-signer/src/client.ts
```diff
@@ -55,9 +55,21 @@ export const RNSignerClientParamsSchema = z.object({
...
+export type ExportWalletResult = {
+  privateKey?: string;
+  seedPhrase?: string;
+  address: string;
+  exportAs: "PRIVATE_KEY" | "SEED_PHRASE";
+};
+
 // TODO: need to emit events
-export class RNSignerClient extends BaseSignerClient<undefined> {
+export class RNSignerClient extends BaseSignerClient<ExportWalletParams> {
```
florrdv: ❓ why do we need this?

account-kit/rn-signer/src/client.ts
```diff
@@ -55,9 +55,21 @@ export const RNSignerClientParamsSchema = z.object({
...
   private stamper = NativeTEKStamper;
+  private storage = new MMKV();
```
florrdv: ❓ can we get rid of this? why do we need to persist these keypairs?

account-kit/rn-signer/src/client.ts
```diff
@@ -274,8 +286,161 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
...
+   *
+   * @param {ExportWalletParams} params Export parameters
+   * @param {string} params.exportAs Whether to export as PRIVATE_KEY or SEED_PHRASE (defaults to PRIVATE_KEY)
+   * @returns {Promise<boolean>} Returns true on successful export
+   * @throws {Error} If the user is not authenticated or export fails
+   */
+  override async exportWallet(params?: ExportWalletParams): Promise<boolean> {
+    await this.exportWalletWithResult(params);
+    return true;
+  }
```
florrdv: ❓ should we just have this function return the result?

account-kit/rn-signer/src/client.ts
florrdv: noob question: does turnkey have anything we can use here? they handle it themselves for the react flow, so I would assume they have _some_ level of abstraction there

account-kit/rn-signer/src/client.ts
florrdv: ah is this because it's the way we do it on the web version and we don't have access to the key there? if so, should we even implement this? 

how would I use this function? It seems like it would just create the result in memory somewhere and then erase it right?

account-kit/rn-signer/src/client.ts
androbwebb: `BaseSignerClient` takes a generic type that's `TExportWalletParams`. (Not something I added)

account-kit/rn-signer/src/client.ts
androbwebb: yeah, exactly. Haha yeah, I agree it's weird. Are we ok having different function signatures with react vs react-native? If so, that's fine with me to remove.

account-kit/rn-signer/src/client.ts
androbwebb: Hmm, yeah fair. Left over from a different strategy (when I was trying to iFrame it)

account-kit/rn-signer/src/client.ts
androbwebb: They do, sorta, but it got very complicated in our current architecture. They have a library `@turnkey/react-native-sdk` which offers a higher-level client, BUT it seems to:

1. Only work in a react-y way. Where the client expects to be within a react context `<TurnkeyProvider .../>` which would require use to set that up in our `AlchemyAccountContext`, which felt kinda heavyweight.
2. it only supports seed phrases anyway, not private key export. 

So, it just made more sense to give the fully-featured option. I agree I don't love relying on their lower-level API for this, but I also don't think they'll be changing that API often, since it's what their other libraries are built on.

florrdv: thanks webb! happy to go with this if there's no obvious easier option

account-kit/rn-signer/src/client.ts
florrdv: oh interesting, thanks!!

account-kit/rn-signer/src/client.ts
florrdv: yeah I think I would prefer to eliminate that MMKV dependency here if we can just do it all in-memory

account-kit/rn-signer/src/client.ts
florrdv: should we do the same what we did with the export wallet params and add a generic for this so it can be different per implementation? I'm happy with different function signatures but would prefer the function name to be the same if possible

account-kit/rn-signer/src/client.ts
```diff
@@ -55,9 +55,21 @@ export const RNSignerClientParamsSchema = z.object({
 
 export type RNSignerClientParams = z.input<typeof RNSignerClientParamsSchema>;
 
+export type ExportWalletParams = {
+  exportAs?: "PRIVATE_KEY" | "SEED_PHRASE";
+};
```
florrdv, line 59: ❓ is there more future proofing we have to do here? what if we want to export solana wallets in the future?

androbwebb: So... we could make the typing take a `vm: 'svm' | 'evm'`, to support `Ed25519` keys & the different derivation path for seed phrases. But I don't think there's much value in adding the typing now if we're not gonna support it right now.

florrdv: my concern is more that the private key & seed phrase distinction is an EVM only thing. so adding this option top level makes it harder to support SVM in a backwards compatible way later on

e.g. now when we add solana support, we will still have the exportAs option which is only relevant for EVM chains. something we could do to avoid this is to not have an ExportWalletParams object but just have it be the enum and then later when we add svm support we can switch to an object and nest the parameters as we see fit

account-kit/rn-signer/src/client.ts
```diff
@@ -273,8 +286,139 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
...
+        });
+
+        const walletAccounts = await Promise.all(
+          wallets.map(({ walletId }) =>
+            this.turnkeyClient.getWalletAccounts({
+              organizationId: this.user!.orgId,
+              walletId,
+            }),
+          ),
+        ).then((x) => x.flatMap((x) => x.accounts));
```
florrdv: 🌈 nit: I'm not huge on mixing async/await and then/catch. would prefer to stick to just await. I'm unclear what the order of operations is here. does the then execute before the await? in that case, is there anything for us to await after the then or does the then return a promise? seems hard to handler errors with if we wanted to

account-kit/rn-signer/src/client.ts
```diff
@@ -273,8 +286,139 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
...
+
+        const walletAccount = walletAccounts.find(
+          (x) => x.address === this.user!.address,
+        );
```
florrdv: ❓ will these always both be in the same format? should we toLowerCase or canonicalize these?

account-kit/rn-signer/src/client.ts
```diff
@@ -54,8 +53,22 @@ export const RNSignerClientParamsSchema = z.object({
...
+export type ExportWalletParams = {
+  exportAs?: "PRIVATE_KEY" | "SEED_PHRASE";
+};
+
+export type ExportWalletResult = {
+  privateKey?: string;
+  seedPhrase?: string;
+  address: string;
+  exportAs: "PRIVATE_KEY" | "SEED_PHRASE";
+};
```
florrdv: :bangbang: thoughts on just having this return a hex string? you can never have both a privateKey and seedPhrase at the same time, so this type looks a little odd to me. presumably I also know what address I just exported from (the same as the user) and I know what type I just exported as (input params)

account-kit/rn-signer/src/client.ts
```diff
@@ -54,8 +54,22 @@ export const RNSignerClientParamsSchema = z.object({
 
 export type RNSignerClientParams = z.input<typeof RNSignerClientParamsSchema>;
 
+export enum ExportWalletAs {
+  PRIVATE_KEY = "PRIVATE_KEY",
+  SEED_PHRASE = "SEED_PHRASE",
+}
+
+export type ExportWalletParams = {
```
androbwebb, line 57: @florrdv making this type into a simple `string` (or the `enum` above) requires a larger refactor in the `useWalletExport` function too. I just didn't want to get into it

account-kit/rn-signer/src/client.ts
```diff
@@ -54,8 +54,22 @@ export const RNSignerClientParamsSchema = z.object({
 
 export type RNSignerClientParams = z.input<typeof RNSignerClientParamsSchema>;
 
+export enum ExportWalletAs {
+  PRIVATE_KEY = "PRIVATE_KEY",
+  SEED_PHRASE = "SEED_PHRASE",
+}
```
florrdv: 🌈 can we revert this to not be an enum? no need

account-kit/signer/src/base.ts
```diff
@@ -59,7 +59,9 @@ import {
 import { assertNever } from "./utils/typeAssertions.js";
 import { hashAuthorization } from "viem/utils";
 
-export interface BaseAlchemySignerParams<TClient extends BaseSignerClient> {
+export interface BaseAlchemySignerParams<
```
florrdv: ‼️ do we need this here? I think we have default generic values right?

account-kit/signer/src/base.ts
```diff
@@ -118,8 +120,9 @@ type GetUserParams =
  * Base abstract class for Alchemy Signer, providing authentication and session management for smart accounts.
  * Implements the `SmartAccountAuthenticator` interface and handles various signer events.
  */
-export abstract class BaseAlchemySigner<TClient extends BaseSignerClient>
-  implements SmartAccountAuthenticator<AuthParams, User, TClient>
+export abstract class BaseAlchemySigner<
+  TClient extends BaseSignerClient<any, any>,
```
florrdv: same here, any any

florrdv (approved):

PR #1879, opened 2025-08-14
feat: add accessToken field to user object

dphilipson:
Expose the access token which may be produced as part of the OIDC flow so that it can be used by callers to retrieve additional user information.


account-kit/rn-signer/src/client.ts
```diff
@@ -176,7 +177,11 @@ export class RNSignerClient extends BaseSignerClient<undefined> {
       throw new Error("Failed to inject credential bundle");
     }
 
-    const user = await this.whoami(params.orgId, params.idToken);
+    const user = await this.whoami(
+      params.orgId,
+      params.idToken,
+      params.accessToken,
```
florrdv, line 199: ❓ do we just pass it here to store it? kinda odd that it's not really used functionally but fits our current patterns

examples/ui-demo/src/app/config.tsx
```diff
@@ -113,6 +113,7 @@ export const alchemyConfig = () => {
           : [walletConnect({ projectId: "30e7ffaff99063e68cc9870c105d905b" })],
       storage: cookieStorage,
       enablePopupOauth: true,
+      oauthCallbackUrl: "http://localhost:6288/callback",
```
florrdv: is this intentional?

florrdv (approved):

examples/ui-demo/src/app/config.tsx
dphilipson: Oh no, thanks.

We should probably put this as an environment variable to make it easier to switch on and off, although that might be confusing to readers since this is only useful for Alchemy people.

account-kit/rn-signer/src/client.ts
dphilipson: Yeah, pretty much. `this.whoami` has the side effect of setting `this.user`, which combines information from the result of the /whoami API call and other info we have like the id token and now the access token.

jakehobbs (approved):

PR #1888, opened 2025-08-16
feat(v5): wallet api client 1193 provider

jakehobbs:
# Pull Request Checklist

### DIFF
--- a/packages/wallet-apis/src/client.ts
+++ b/packages/wallet-apis/src/client.ts
@@ -1,16 +1,14 @@
 import {
   createClient,
-  Address,
-  Chain,
-  Client,
-  LocalAccount,
-  ParseAccount,
-  Transport,
+  type Address,
+  type Chain,
+  type LocalAccount,
   createWalletClient,
-  custom,
   toHex,
-  Hex,
-  WalletRpcSchema,
+  type TypedDataDefinition,
+  sliceHex,
+  type WalletSendCallsReturnType,
+  type WalletSendCallsParameters,
 } from "viem";
 import {
   smartWalletActions,
@@ -21,12 +19,14 @@ import type {
   BaseWalletClient,
   ExtractRpcMethod,
   SignerClient,
-  SmartWalletClientRpcSchema,
+  SmartWalletClient1193Methods,
 } from "./types.js";
-import type { WalletServerViemRpcSchema } from "@alchemy/wallet-api-types/rpc";
 import { createInternalState } from "./internal.js";
 import { AccountNotFoundError } from "../../../aa-sdk/core/src/errors/account.js";
 import { ChainNotFoundError } from "../../../aa-sdk/core/src/errors/client.js";
+import type { SmartWalletClientEip1193Provider } from "./types.js";
+import type { PrepareCallsParams } from "./actions/prepareCalls.js";
+import EventEmitter from "events"; // TODO(jh): will this work in browser?
 
 export type CreateSmartWalletClientParams<
   TAccount extends Address | undefined = Address | undefined,
@@ -64,7 +64,7 @@ export const createSmartWalletClient = <
   policyIds,
 }: CreateSmartWalletClientParams<TAccount>): BaseWalletClient<
   SmartWalletActions<TAccount>
-> => {
+> & { getProvider: () => SmartWalletClientEip1193Provider } => {
   const _policyIds = [
     ...(policyId ? [policyId] : []),
     ...(policyIds?.length ? policyIds : []),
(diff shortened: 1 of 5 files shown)


packages/wallet-apis/src/provider.e2e.test.ts
```diff
@@ -0,0 +1,228 @@
...
+        {
+          to: zeroAddress,
+          data: "0x",
+          value: 0n,
+        },
+      ],
+    });
+    await waitForCallsStatus(clientFromProvider, result);
+  });
+
```
jakehobbs, line 205: need to add test for `getCapabilities` once it's live in prod.

packages/wallet-apis/src/provider.ts
```diff
@@ -0,0 +1,207 @@
...
+
+      case "personal_sign": {
+        return handler<"personal_sign">(async ([data, address]) => {
+          if (!client.account) {
+            throw new AccountNotFoundError();
+          }
+          if (
+            address?.toLowerCase() !== client.account.address?.toLowerCase()
+          ) {
+            throw new BaseError(
```
jakehobbs: is it ok to use `BaseError` so often, or should i really be creating individual error classes as needed?

packages/wallet-apis/src/provider.ts
```diff
@@ -0,0 +1,207 @@
...
+export const createEip1193ProviderFromClient = <
+  TAccount extends Address | undefined = Address | undefined,
+>(
+  client: BaseWalletClient<SmartWalletActions<TAccount>>,
+) => {
+  // TODO(v5): implement any other supported events: https://eips.ethereum.org/EIPS/eip-1193#events
+  const eventEmitter = new EventEmitter();
+
+  // TODO(v5): revisit if we actually want to auto-connect the provider like this.
+  (async () => {
```
thebrianchen, line 50: just curious, is auto-connecting the default behavior / part of the spec, or something that we do for convenience?

thebrianchen (approved):

packages/wallet-apis/src/provider.ts
```diff
@@ -0,0 +1,207 @@
...
+    const account = await client.requestAccount();
+    client.account = {
+      type: "json-rpc",
+      address: account.address,
+    };
+    eventEmitter.emit("connect", { chainId: client.chain.id });
```
blakecduncan: Nit, I think EIP-1193 expects a hex to be passed as the chain id on connect

packages/wallet-apis/src/provider.ts
```diff
@@ -0,0 +1,207 @@
...
+            throw new ChainNotFoundError();
+          }
+          return toHex(client.chain.id);
+        })(params);
+      }
+
+      case "eth_accounts": {
+        return handler<"eth_accounts">(async () => {
+          if (!client.account) {
+            throw new AccountNotFoundError();
```
blakecduncan: I'm not sure if this is overkill because we will most likely be the main consumer of the provider.  the 1193 spec has a specific format for errors.

https://eips.ethereum.org/EIPS/eip-1193#errors

If we want to do this, the coinbase sdk has the error codes mapped out here
https://github.com/base/account-sdk/blob/48012fda55661e53b6a02c066eed688a5321b461/packages/account-sdk/src/core/error/constants.ts

packages/wallet-apis/src/provider.ts
```diff
@@ -0,0 +1,207 @@
...
+          if (!client.chain) {
+            throw new ChainNotFoundError();
+          }
+          const { to, data, value } = tx;
+          if (!to) {
+            throw new BaseError("'to' is required.");
+          }
+          const result = await client.sendCalls({
+            calls: [{ to, data, value }],
+            from: client.account.address,
```
blakecduncan: Do you think we should check the from that is passed to the txn like this?  I think I've seen this in other providers.  It might be confusing if someone can just pass a random from value and things still work.

```
if (tx.from && tx.from.toLowerCase() !== client.account.address.toLowerCase()) {
  throw new Error()
}
```

packages/wallet-apis/src/client.ts
```diff
@@ -72,15 +71,20 @@ export const createSmartWalletClient = <
...
 
-  return baseClient
-    .extend(() => ({
-      policyIds: _policyIds,
-      internal: createInternalState(),
-    }))
-    .extend(smartWalletActions(signerClient));
+  return {
+    ...baseClient,
+    getProvider: () => createEip1193ProviderFromClient(baseClient),
```
blakecduncan: I think it's possible for this spread to cause issues.  The spread on baseClient will return a shallow copy of the baseClient where as you are passing the baseClient directly to the `createEip1193ProviderFromClient`.  So essentially we are returning two separate copies of the client, one at the root of the object and one inside the provider.  We do edit the account returned inside the provider so that's one example of how they could get out of sync.

```
    client.account = {
      type: "json-rpc",
      address: account.address,
    };
```

We can do this to ensure the client in the provider is the same object

```
const client = baseClient as BaseWalletClient<SmartWalletActions<T>> & { getProvider: () => SmartWalletClientEip1193Provider };
client.getProvider = () => createEip1193ProviderFromClient(client);
return client;

```

This may be over cautious, AI found this one for me and I thought is was potentially a legit edge case that could cause issues.

blakecduncan (approved):
Pumped for this, nice job!

Just a few comments but nothing blocking

packages/wallet-apis/src/provider.ts
jakehobbs: good catch

packages/wallet-apis/src/provider.ts
jakehobbs: yeah, worth checking 👍

package.json
```diff
@@ -38,6 +38,7 @@
     "postbuild": "yarn lint:write",
     "build:libs": "turbo run build --filter='@account-kit/*' --filter='@aa-sdk/*' --filter='./templates/*' --filter='./packages/*'",
     "build:demo": "turbo run build --filter=ui-demo --filter='./templates/*'",
+    "build:v5": "turbo run build --filter='./packages/*'",
```
adamegyed, line 41: 🙏

packages/wallet-apis/src/client.ts
jakehobbs: better safe than sorry. cleanest way is prob to just use `extend`?

packages/wallet-apis/src/provider.ts
jakehobbs: i see. adapted slightly for this.

packages/wallet-apis/src/provider.ts
jakehobbs: i think coinbase automatically does it when a request comes in, but not immediately when the provider is created.

turnkey sets it any time AFTER a successful rpc request.

this is what the spec says:

Screenshot 2025-08-19 at 11 11 01 AM

Screenshot 2025-08-19 at 11 10 42 AM

mainly doing it out of convenience for now, but def open for discussion & not sure where we'll end up.

packages/wallet-apis/src/provider.ts
```diff
@@ -0,0 +1,257 @@
...
+            }
+            const result = await client.sendCalls({
+              calls: [{ to, data, value }],
+              from: client.account.address,
+              // TODO(v5): do we need to support any overrides here?
+            } as PrepareCallsParams<TAccount>);
+            const uoResult = await client.waitForCallsStatus({
+              id: result.preparedCallIds[0],
+            });
+            const txHash = uoResult.receipts?.[0]?.transactionHash;
```
adamegyed: Interesting... I think we will need to explicitly describe cases where callIds length > 0 (currently none). It seems like a good number of things make the assumption that len(callIds) == 1.

Here, we also assume that it will always be completed in 1 transaction, for getting the receipt.

packages/wallet-apis/src/provider.ts
```diff
@@ -0,0 +1,257 @@
...
+              throw new InvalidRequestError(
+                "Cannot use 'from' address other than connected account.",
+              );
+            }
+            const result = await client.sendCalls({
+              calls: [{ to, data, value }],
+              from: client.account.address,
+              // TODO(v5): do we need to support any overrides here?
+            } as PrepareCallsParams<TAccount>);
+            const uoResult = await client.waitForCallsStatus({
```
adamegyed: nit: uoResult -> callsStatusResult

packages/wallet-apis/src/client.ts
```diff
@@ -51,14 +51,13 @@ export const createSmartWalletClient = <
...
-  ParseAccount<TAccount>,
-  WalletServerViemRpcSchema,
+}: CreateSmartWalletClientParams<TAccount>): BaseWalletClient<
   SmartWalletActions<TAccount>
-> => {
-  const baseClient = createClient({ account, transport, chain });
+> & { getProvider: () => SmartWalletClientEip1193Provider } => {
+  const _policyIds = [
+    ...(policyId ? [policyId] : []),
+    ...(policyIds?.length ? policyIds : []),
```
dphilipson: (Nit) Seems like this line could be simplified to
```ts
...(policyIds ?? [])
```

packages/wallet-apis/src/provider.ts
```diff
@@ -0,0 +1,257 @@
...
+    };
+    eventEmitter.emit("connect", { chainId: toHex(client.chain.id) });
+  })();
+
+  const request = (async ({ method, params }) => {
+    try {
+      switch (method) {
+        case "eth_chainId": {
+          return handler<"eth_chainId">(async () => {
+            if (!client.chain.id) {
```
dphilipson: (Nit) Just in case some self-important chain decides to take chain id 0 in the future, I'd check that `client.chain.id != null` here.

packages/wallet-apis/src/provider.ts
```diff
@@ -0,0 +1,257 @@
...
+            if (
+              address?.toLowerCase() !== client.account.address.toLowerCase()
+            ) {
+              throw new InvalidRequestError(
+                "Cannot sign for an address other than the current account.",
+              );
+            }
+            return await client.signTypedData({
+              account: client.account.address,
+              ...(JSON.parse(tdJson) as TypedDataDefinition),
```
dphilipson, line 115: Is `tdJson` a json string? Is this right? I thought the second param of `eth_signTypedData_v4` should be an object, for example in [MetaMask's docs](https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_signtypeddata_v4/).

packages/wallet-apis/src/provider.ts
```diff
@@ -0,0 +1,257 @@
...
+            }
+            const [{ calls, capabilities, chainId, from }] = _params;
+            if (!client.account) {
+              throw new AccountNotFoundError();
+            }
+            if (!client.chain) {
+              throw new ChainNotFoundError();
+            }
+            if (chainId !== toHex(client.chain.id)) {
+              throw new InvalidRequestError("Invalid chain ID.");
```
dphilipson: (Nit) Could be nice to include the expected chain id in this error message.

packages/wallet-apis/src/client.ts
dphilipson: Oh, I see this line already existed and got moved, never mind!

packages/wallet-apis/src/provider.ts
adamegyed: Also, waiting for the UO to mine within the handler for `eth_sendTransaction` is a bit odd, since typically, the workflow would be to have the transaction hash known immediately and returned, but take a while to mine from the mempool. I think this is probably the best backwards-compatible solution, but it will change the loading state behavior for old apps.

adamegyed (approved):
Looking good!

packages/wallet-apis/src/provider.ts
dphilipson: This point came up in the very first meetings about SDK when we were planning v1. I said the same thing you did, that this behavior would be breaking for apps that expect tx hashes to show up right away, but we decided that pragmatically there isn't really anything better to do here. If you have other ideas I'd be pretty interested.

packages/wallet-apis/src/client.ts
jakehobbs: worth changing, good spot

packages/wallet-apis/src/provider.ts
adamegyed: Unfortunately I don't have an alternative. In most cases, the old app would be querying the transaction status from a different RPC than the wallet, which means we can't return a fake transaction hash that is really served by a `wallet_getCallsStatus` handler or similar.

packages/wallet-apis/src/provider.ts
jakehobbs: according to viem, it should be a string by the time it gets to the rpc provider level: https://github.com/wevm/viem/blob/6c7332485669245ee18fde1a6abfce33d9dcafee/src/types/eip1193.ts#L1756

it gets serialized in the viem action: https://github.com/wevm/viem/blob/6c7332485669245ee18fde1a6abfce33d9dcafee/src/actions/wallet/signTypedData.ts#L187

you can it here in base's example too: https://github.com/base/docs/blob/7549bf084e203b2abb79bb292cc8afebbfb5e724/docs/base-account/guides/sign-and-verify-typed-data.mdx?plain=1#L101-L104

packages/wallet-apis/src/provider.ts
jakehobbs: another interesting thing that i just noticed is that if you try to use `sendCalls` only a wallet that doesn't natively support it, viem can fallback to sending it as individual txns and then return a call id that is kinda similar to ours... https://github.com/wevm/viem/blob/6c7332485669245ee18fde1a6abfce33d9dcafee/src/actions/wallet/sendCalls.ts#L223-L228 (they just concat all of the hashes + the chain id + a magic identifier).

jakehobbs: also worth mentioning is that the `wallet_sendCalls` return type viem specifies is just a single `id` that represents the batch of calls: https://github.com/wevm/viem/blob/6c7332485669245ee18fde1a6abfce33d9dcafee/src/types/eip1193.ts#L253-L258

which is also in alignment with the spec: https://www.eip5792.xyz/reference/sendCalls#returns

packages/wallet-apis/src/provider.ts
jakehobbs: yeah, i just did this since it's what we did in v4. i don't really have any other ideas that adhere to the expected rpc response type, since just returning a UO hash seems like it would be pretty confusing.

PR #1890, opened 2025-08-18
feat: end to end signer oauth

linnall:
# Description
- [x] popup oauth flow
- [ ] redirect oauth flow

These flows are demoed in the `signer-web-example` project

# Pull Request Checklist

---

## PR-Codex overview
This PR primarily focuses on enhancing the OAuth authentication flow within the `signer` and `signer-web` packages. It introduces new error handling, updates type definitions, and refines the logic for handling OAuth logins.

### Detailed summary
- Deleted numerous `.mdx` documentation files related to various functions and hooks.
- Added `OauthCancelledError` and `OauthFailedError` classes for improved error handling.
- Enhanced the OAuth login process in `AuthClient` with better configuration and state management.
- Updated `createWebAuthClient` to handle OAuth flows more effectively.
- Introduced new types for OAuth parameters and providers in `types.ts`.
- Refactored the `loginWithOauth` method to streamline the authentication process.
- Adjusted `tsconfig.build.json` to include `rootDir` and `outDir`.
- Modified example usage in `page.tsx` to include a button for Google OAuth login.


examples/signer-web-example/app/page.tsx
```diff
@@ -18,14 +18,20 @@ export default function Home(): ReactElement {
...
   const handleOauthLogin = useCallback(async () => {
     console.log("Logging in with OAuth...");
-    const signer = await authClient.loginWithOauth({
+    await authClient.loginWithOauth({
       type: "oauth",
       authProviderId: "google",
       mode: "redirect",
     });
-    setSigner(signer);
+    console.log("do we ever get here?");
```
dphilipson, line 34: You're right that if `mode: "redirect"` is used, then this won't be reached.

packages/signer/src/authClient.ts
```diff
@@ -232,6 +331,50 @@ export class AuthClient {
...
+      if (callbackParams.status === "ERROR") {
+        throw new Error(callbackParams.error);
+      }
+      return this.completeAuthWithBundle({
+        bundle: callbackParams.bundle!,
+        orgId: callbackParams.orgId!,
+        idToken: callbackParams.idToken,
+      });
+    }
+    console.log("No OAuth callback parameters found");
```
dphilipson, line 375: Make sure to remove this console.log before merging.

packages/signer/src/authClient.ts
```diff
@@ -44,13 +69,141 @@ type PendingOtp = {
...
 
+/**
+ * Reads and removes the specified query params from the URL.
+ *
+ * @param {T} keys object whose values are the query parameter keys to read and
+ * remove
+ * @returns {{ [K in keyof T]: string | undefined }} object with the same keys
+ * as the input whose values are the values of the query params.
+ */
+function getAndRemoveQueryParams<T extends Record<string, string>>(
```
dphilipson, line 80: This function, and the functions that call it, will fail if not in a browser and so probably belong in the `signer-web` package.

Edit: actually I'm not sure off the top of my head how this flow works in react native. Is it creating a webview to run the redirect flow and the extracting the query params from the url at the end? If so, maybe we do need it in the common package.

packages/signer/src/authClient.ts
```diff
@@ -91,10 +276,48 @@ export class AuthClient {
...
+   *   authProviderId: "google",
+   *   mode: "popup"
+   * });
+   * ```
+   */
   public async loginWithOauth(params: LoginWithOauthParams): Promise<Signer> {
-    const authUrl = await notImplemented(params);
-    const response = await this.handleOauthFlow(authUrl);
+    const { targetPublicKey } = await this.getTekStamper();
+    const oauthConfig = await this.dev_request("prepare-oauth", {
```
dphilipson, line 303: This may not work: we want to call `/prepare-oauth` on startup so we may need a separate method for that.

The reason for this is that browsers restrict when popup windows can be opened to stop spammy websites from opening a bunch of popups on load. Browser policies differ, but some examples are only allowing popups to open for a short time after a user operation or from contexts that the browser considers "downstream" from user operations. [Here's a thread](https://alchemyinsights.slack.com/archives/C082ZDA1822/p1755797329715369?thread_ts=1755796746.708609&cid=C082ZDA1822) talking about this.

If this is working in your tests, we can merge with this for now and revisit, but we'll want to check on this before releasing v5.

packages/signer/src/authClient.ts
```diff
@@ -91,10 +276,48 @@ export class AuthClient {
...
+    });
+    const authUrl = getOauthProviderUrl({
+      oauthParams:
+        params.mode === "redirect" ? { ...params, redirectUrl: "/" } : params, // TO DO: clean up redirect vs popup path
+      turnkeyPublicKey: targetPublicKey,
+      oauthCallbackUrl: "https://signer.alchemy.com/callback",
+      oauthConfig,
+    });
+    const response = await this.handleOauthFlow(authUrl, params.mode);
+    console.log({ response });
```
dphilipson, line 314: Make sure to remove this console.log before merging, likewise just below this.

packages/signer/src/utils.ts
```diff
@@ -0,0 +1,243 @@
...
+ * @returns {string} The absolute URL
+ */
+export function resolveRelativeUrl(url: string): string {
+  // Funny trick.
+  const a = document.createElement("a");
+  a.href = url;
+  return a.href;
+}
+
+export class OAuthProvidersError extends Error {
```
dphilipson, line 95: Let's rename this to be more specific to what it means, like `OauthProvidersNotFoundError` or `NoOauthProvidersFoundError`.

packages/signer/src/utils.ts
```diff
@@ -0,0 +1,243 @@
...
+
+/**
+ * Resolves a relative URL to an absolute URL (browser environment)
+ *
+ * @param {string} url - The URL to resolve
+ * @returns {string} The absolute URL
+ */
+export function resolveRelativeUrl(url: string): string {
+  // Funny trick.
+  const a = document.createElement("a");
```
dphilipson, line 90: Since this breaks outside of browsers, ideally it would be in `signer-web`, although I don't know if that makes other parts of the code structure awkward. But we have to be sure this doesn't get called on react-native.

dphilipson (commented):
This looks really good. I left some comments but this is really massive progress so amazing job here.

packages/signer/src/authClient.ts
linnall: I will move these functions into `auth/utils.ts` for now. We can revisit this when we implement react native support

linnall:
open comments being addressed in #2098

PR #828, opened 2024-07-11
refactor: clean up some of the light account types

moldy530:
# Pull Request Checklist

- [ ] Did you add new tests and confirm existing tests pass? (`yarn test`)
- [ ] Did you update relevant docs? (docs are found in the `site` folder, and guidelines for updating/adding docs can be found in the [contribution guide](https://github.com/alchemyplatform/aa-sdk/blob/main/CONTRIBUTING.md))
- [ ] Do your commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [ ] Does your PR title also follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [ ] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples)? (e.g. `feat!: breaking change`)
- [ ] Did you run lint (`yarn lint:check`) and fix any issues? (`yarn lint:write`)
- [ ] Did you follow the [contribution guidelines](https://github.com/alchemyplatform/aa-sdk/blob/main/CONTRIBUTING.md)?

---

## PR-Codex overview
This PR updates functions and types related to light accounts in the smart contracts. 

### Detailed summary
- Renamed `getLightAccountVersionDef` to `getLightAccountVersionForAccount`
- Added a `type` property in `multiOwnerAlchemyClient.ts`
- Updated type definitions for `LightAccountVersion`
- Updated function parameters and return types in `createAccount.ts`
- Updated function imports and descriptions in multiple files

> The following files were skipped due to too many changes: `account-kit/smart-contracts/src/light-account/accounts/multiOwner.ts`, `account-kit/smart-contracts/src/light-account/accounts/base.ts`, `account-kit/smart-contracts/src/light-account/types.ts`, `account-kit/smart-contracts/src/light-account/utils.ts`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #1907, opened 2025-08-20
feat: static smart account interface & impl for light account

adamegyed:
## Motivation

We want to be able to use account-implementation-specific logic without constructing a concrete instance of a smart account, where possible.

Currently, we have to create mocks or stubs of ether a signer or an account address, to access properties of an account implementation that don’t depend on it.

## Solution

Define a `StaticSmartAccountImplementation` type to provide these fields.

Implement this static implementation for all LightAccount types.
- Currently these still depend on the `lightAccountVersionRegistry`, but we could migrate this in the future if we want to further simplify.

### Misc cleanup

- Move `lowerAddress` to `@alchemy/common`
- Remove the unused export function `getLightAccountVersionForAccount`

# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on adding the `lowerAddress` function to the `@alchemy/common` package, enhancing the utility for address manipulation. It also updates related files to utilize this new function and modifies documentation accordingly.

### Detailed summary
- Added `lowerAddress` function in `packages/common/src/utils/lowerAddress.ts`.
- Updated exports in `packages/common/src/index.ts` to include `lowerAddress`.
- Removed `getLightAccountVersionForAccount` function from `packages/smart-accounts/src/light-account/utils.ts`.
- Updated imports in multiple files to include `lowerAddress`.
- Created documentation for `lowerAddress` in `docs/pages/reference/alchemy/common/functions/lowerAddress.mdx`.
- Adjusted navigation in `docs/docs.yml` to include the new `lowerAddress` page.


packages/smart-accounts/src/light-account/utils.ts
```diff
@@ -198,71 +194,3 @@ export async function getLightAccountImplAddress<
...
-
-/**
- * Get the light account version definition for the given light account and chain
- *
- * @template {LightAccountBase} TAccount
- * @param {LightAccountBase} account the light account to get the version for
- * @param {Chain} chain - the chain to get the version for
- * @returns {Promise<LightAccountVersionConfig>} the light account version definition for the given light account and chain
- */
-export async function getLightAccountVersionForAccount<
```
adamegyed, line 210: For context: this was created in an ancient PR here: https://github.com/alchemyplatform/aa-sdk/pull/828/files#diff-0bbd548d4d2b59fe196096a4b14f0a2441592ed90f2dd0000fec9f3f84b0d03eL235

And based on my research, it never seems to have been used internally, just exported.

This function gets you the version of an instance of a LightAccount at runtime, which doesn't actually seem useful, since you would typically have that information already based on how it was constructed.

Removing here to minimized exported API surface.

packages/smart-accounts/src/light-account/lightAccountStaticImpl.ts
```diff
@@ -0,0 +1,193 @@
...
+// Shared type and fields for all light account v2 implementations
+const lightAccountV2Base = {
+  entryPoint: {
+    abi: entryPoint07Abi,
+    address: entryPoint07Address,
+    version: "0.7",
+  },
+} satisfies Partial<StaticSmartAccountImplementation<false, "0.7">>;
+
+export const lightAccountStaticImplV2_0_0: StaticSmartAccountImplementation<
```
thebrianchen, line 120: as a newcomer, all the different light account versions seem pretty intimidating. can we add comments or a rough guide on when to use which version?

thebrianchen: similarly, do we need a migration guide for devs migrating from v4 that have been creating dummy signers?

thebrianchen (approved):
lgtm with some questions

packages/smart-accounts/src/light-account/lightAccountStaticImpl.ts
adamegyed: I agree we should have that guide, but this code section doesn't feel like the right place for it. In normal circumstances, users won't even be in this package unless they already know what account type they're using.

The funny thing is that the existing dummy signer pattern will continue to work (after applying the overall v5 migration from `SmartAccountSigner` -> `WalletClient`), so I don't know if we need a specific section for it. I think just documenting these static impls as public exported functions should be sufficient for users digging around at this level.

PR #1923, opened 2025-08-22
fix: improve dark mode for docs quickstart

0xmts:
## Summary
- ensure React Quickstart CTA uses white text on dark mode
- lighten dark mode background color for CTA button

## Testing
- `npx prettier docs/pages/index.mdx --check`
- `yarn install` (fails: The engine "node" is incompatible with this module)
- `npx vitest run docs/pages/index.mdx` (fails: 403 Forbidden)

------
https://chatgpt.com/codex/tasks/task_b_68a8a53eb88083298576b7f54830f180

---

## PR-Codex overview
This PR introduces a new button style for the `react-quickstart-btn`, enhancing its appearance with CSS variables for light and dark themes, and applying these styles to a button in the documentation.

### Detailed summary
- Added `className="react-quickstart-btn"` to a button.
- Introduced CSS variables for button background and text color.
- Defined styles for light and dark themes.
- Created a `.react-quickstart-btn` class to apply the new styles.

### DIFF
--- a/docs/pages/index.mdx
+++ b/docs/pages/index.mdx
@@ -17,9 +17,8 @@ slug: wallets
   <Button
     icon="bolt"
     href="/wallets/react/quickstart"
+    className="react-quickstart-btn"
     style={{
-      backgroundColor: "var(--accent-2, #ECF3FF)",
-      color: "black",
       padding: "2rem 2rem",
       fontSize: "1rem",
     }}
@@ -39,6 +38,19 @@ slug: wallets
     Try our demo
   </Button>
 </div>
+
+<style>{`
+  .react-quickstart-btn {
+    background-color: var(--accent-2, #ECF3FF);
+    color: black;
+  }
+  @media (prefers-color-scheme: dark) {
+    .react-quickstart-btn {
+      background-color: #4B5563;
+      color: white;
+    }
+  }
+`}</style>
 
 ## Everything you need for onchain applications
 
--- a/docs/pages/index.mdx
+++ b/docs/pages/index.mdx
@@ -40,15 +40,19 @@ slug: wallets
 </div>
 
 <style>{`
-  .react-quickstart-btn {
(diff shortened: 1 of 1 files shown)


docs/pages/index.mdx
```diff
@@ -40,6 +39,23 @@ slug: wallets
   </Button>
 </div>
 
+<style>{`
+  :root {
+    --react-quickstart-btn-bg: var(--accent-2, #ECF3FF);
+    --react-quickstart-btn-color: black;
+  }
```
dslovinsky, line 46: `--react-quickstart-btn-bg: var(--accent-2, #ECF3FF);`
Ya this doesn't make sense - it's trying to set a new variable using... an existing variable. So why wouldn't we use the original variable haha

docs/pages/index.mdx
```diff
@@ -40,6 +39,23 @@ slug: wallets
...
+
+  html[data-theme='dark'] {
+    --react-quickstart-btn-bg: #777777;
+    --react-quickstart-btn-color: white;
+  }
+
+  .react-quickstart-btn {
+    background-color: var(--react-quickstart-btn-bg);
+    color: var(--react-quickstart-btn-color);
+  }
```
Richard-Dang, line 56: Fern supports dark mode out of the box right? I think this change would be better defined at the design system level as opposed to a one-off style on the button

docs/pages/index.mdx
dslovinsky: `--react-quickstart-btn-color: black;`
This line is the source of the problem. It's basically the same as it was before. It's hard-coding the text color as `black` all the time which is not what we want. Fern's CSS natively handles text colors for us so we should not overwrite it.

dslovinsky (changes requested):

dslovinsky:
Superseded by https://github.com/alchemyplatform/aa-sdk/pull/1924

PR #1949, opened 2025-08-27
feat: add wallet_getCapabilities to SmartWalletClient

adamegyed:
# Pull Request Checklist


jakehobbs (approved):
should we test it in the provider tests too? i left a comment here for that: https://github.com/alchemyplatform/aa-sdk/blob/af347e0c09f6648ebc3c5a0762a1b534c42440c0/packages/wallet-apis/src/provider.e2e.test.ts#L206

but lgtm apart from that

packages/wallet-apis/src/provider.ts
```diff
@@ -193,6 +194,12 @@ export const createEip1193ProviderFromClient = <
           })(params);
         }
 
+        case "wallet_getCapabilities": {
+          return handler<"wallet_getCapabilities">(async () => {
+            return await getCapabilities(client);
+          })(params);
+        }
```
jakehobbs, line 201: do we need this here? for `wallet_getCallsStatus`, we just let the fallback below handle it, since no transformations are needed.

if you prefer to be explicit like this though, it might be good to add`wallet_getCallsStatus` here too?

jakehobbs (approved):

packages/wallet-apis/src/provider.ts
adamegyed: Ah I see, I added this here mostly because the catchall clause still had a todo comment that made it unclear if we would keep it.

I'm leaning towards keeping it and adding `wallet_getCallsStatus`, mostly just as a reader benefit to make it obvious that we support it. Wdyt?

jakehobbs: yeah, i think that makes sense.

adamegyed: Hmm, having some typing issues with `wallet_getCallsStatus`, it seems like the corresponding viem action annotates the return type with some extra fields and also converts types a bit. I will leave it out for now.

jakehobbs: sounds good

PR #1980, opened 2025-09-02
feat: add swap api decorator

adamegyed:
Adds `wallet_requestQuote_v0` support to `@alchemy/wallet-apis` via a decorator exported at `@alchemy/wallet-apis/experimental`.

- Refactors `requestAccount` to allow overriding the `signerAddress` param.
- Refactors common test setup components to `testSetup.ts`. This file cannot be named `testSetup.test.ts`, otherwise vitest complains about no tests defined in the file.

# Pull Request Checklist

---

## PR-Codex overview
This PR introduces new types and functions for handling optional signer addresses and swap actions in the wallet APIs. It also refactors existing code for better readability and adds tests for the new functionalities.

### Detailed summary
- Added `OptionalSignerAddress` type in `types.ts`.
- Introduced `SwapActions` type and `swapActions` function in `swapActionsDecorator.ts`.
- Refactored `signMessage` and `signTypedData` functions to simplify account retrieval.
- Added `requestQuoteV0` function in `requestQuoteV0.ts`.
- Created tests for swap actions in `swapActionsDecorator.e2e.test.ts`.
- Updated transport configuration in `testSetup.ts` and removed redundant code in `client.e2e.test.ts`.

### DIFF
--- a/packages/wallet-apis/src/actions/signMessage.ts
+++ b/packages/wallet-apis/src/actions/signMessage.ts
@@ -10,7 +10,6 @@ import { prepareSign } from "./prepareSign.js";
 import { signSignatureRequest } from "./signSignatureRequest.js";
 import { formatSign } from "./formatSign.js";
 import { signableMessageToJsonSafe } from "../utils/format.js";
-import { AccountNotFoundError } from "@alchemy/common";
 
 export type SignMessageParams = Prettify<{
   message: SignableMessage;
@@ -44,12 +43,13 @@ export async function signMessage(
   signerClient: SignerClient,
   params: SignMessageParams,
 ): Promise<SignMessageResult> {
-  const account = await requestAccount(client, signerClient, {
-    accountAddress: params.account ?? client.account?.address,
-  });
-  if (!account) {
-    throw new AccountNotFoundError();
-  }
+  const accountAddress = params.account ?? client.account?.address;
+
+  const account = await requestAccount(
+    client,
+    signerClient,
+    accountAddress != null ? { accountAddress } : undefined,
+  );
 
   const prepared = await prepareSign(client, {
     from: account.address,
--- a/packages/wallet-apis/src/experimental/actions/requestQuoteV0.ts
+++ b/packages/wallet-apis/src/experimental/actions/requestQuoteV0.ts
@@ -1,0 +1,52 @@
+import type { Static } from "@sinclair/typebox";
+import {
+  toHex,
+  type Address,
+  type IsUndefined,
+  type Prettify,
+  type UnionOmit,
+} from "viem";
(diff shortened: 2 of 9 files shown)


packages/wallet-apis/src/actions/signMessage.ts
```diff
@@ -44,12 +43,13 @@ export async function signMessage(
...
-  });
-  if (!account) {
-    throw new AccountNotFoundError();
-  }
+  const accountAddress = params.account ?? client.account?.address;
+
+  const account = await requestAccount(
+    client,
+    signerClient,
+    accountAddress != null ? { accountAddress } : undefined,
```
Zer0dot, line 51: afaict accountAddress will either be an address or undefined right? When can it be null?

packages/wallet-apis/src/experimental/actions/requestQuoteV0.ts
```diff
@@ -0,0 +1,52 @@
...
+> = Prettify<
+  OptionalChainId<
+    UnionOmit<
+      Static<
+        (typeof wallet_requestQuote_v0)["properties"]["Request"]["properties"]["params"]
+      >[0],
+      "from"
+    >
+  > &
+    (IsUndefined<TAccount> extends true ? { from: Address } : { from?: never })
```
Zer0dot, line 24: If I understand correctly, this means that if an account is hoisted, you cannot provide a "from" right? 

I wonder if there's a use case where users might actually want to request quotes for other from accounts using the same client. Probably better not to allow that though. Great as is! Just floating a thought.

packages/wallet-apis/src/types.ts
```diff
@@ -34,6 +34,10 @@ export type OptionalChainId<T> = T extends { chainId: Hex }
   ? Omit<T, "chainId"> & { chainId?: Hex | undefined }
   : T;
 
+export type OptionalSignerAddress<T> = T extends { signerAddress: Address }
```
Zer0dot, line 37: ultra small nit but is there a reason we need to check that T extends { signerAddress: Address } ?

I ask cause I feel like using `OptionalSignerAddress` on a type that doesn't have a signer address is a mistake and the compiler should flag it

adamegyed: We need to do this using a `... extends` clause to preserve union types. Conditional types are distributed over unions, and this is necessary because just doing `Omit<..., "signerAddress">` on a discriminated union will break the union type.

packages/wallet-apis/src/actions/signMessage.ts
adamegyed: `!= null` matches on both `null` and `undefined` (note how it is not `!==`).

Afaik, it is considered idiomatic to use `!= null` instead of `!== undefined` to group both of them together.

PR #1981, opened 2025-09-02
chore: move wallet client hooks out of experimental

jakehobbs:
# Pull Request Checklist


account-kit/core/package.json
```diff
@@ -32,11 +32,6 @@
       "import": "./dist/esm/environments/web/index.js",
       "default": "./dist/esm/environments/web/index.js"
     },
-    "./experimental": {
-      "types": "./dist/types/experimental/index.d.ts",
-      "import": "./dist/esm/experimental/index.js",
-      "default": "./dist/esm/experimental/index.js"
-    },
```
adamegyed, line 39: Should we maintain a duplicate export at the same path to allow seamless upgrades, or fully remove it?

account-kit/react/src/hooks/useCallsStatus.ts
```diff
@@ -1,3 +1,5 @@
+"use client";
```
adamegyed, line 1: We can do this in a hook? 👀

account-kit/react/src/hooks/useSmartWalletClient.ts
```diff
@@ -0,0 +1,73 @@
...
+    config: {
+      _internal: { wagmiConfig },
+    },
+    config,
+  } = useAlchemyAccountContext();
+
+  const result = useSyncExternalStore(
+    watchSmartWalletClient(config),
+    () => getSmartWalletClient(config, params),
+    () => getSmartWalletClient(config, params),
```
adamegyed, line 55: This is pushing the limits of my React knowledge, but I thought the 3rd param `getServerSnapshot` should only be used in the context of SSR, but this hook has a `"use client"` directive. Should we be omitting that param?

account-kit/core/package.json
jakehobbs: good question. i think we normally remove. also the upgrade won't be totally seamless b/c the the hooks are taking a smart account client now (instead of a smart wallet client).

account-kit/react/src/hooks/useCallsStatus.ts
jakehobbs: almost all of the other hooks have this. i think not including it previously was an accident.

wagmi includes it in their hooks too: https://github.com/wevm/wagmi/blob/main/packages/react/src/hooks/useCallsStatus.ts

i think it's not really necessary, but some bundlers may complain if it's not there. tbh i haven't used RSCs a ton, so my knowledge here isn't that deep.

account-kit/react/src/hooks/useSmartWalletClient.ts
jakehobbs: components w/ `"use client"` still get SSR'd for the initial render.

components that are NOT marked `"use client"` ONLY render on the server and are not in the client-side bundle.

https://dev.to/shanu001x/why-do-client-components-render-as-ssr-in-nextjs-marking-components-as-use-client-still-render-its-html-as-ssr-why-1e70

adamegyed: interesting, TIL!

PR #2009, opened 2025-09-04
feat: support getAuthClient() on alchemyAuth() connector

thebrianchen:
Adding support for getAuthClient()  on the alchemy auth connector. Filling in the most basic functionality for now. 
Also added an option for the connector
Screenshot 2025-09-04 at 5 52 37 PM

# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing the Alchemy authentication process by integrating the `alchemyAuth` connector, improving documentation, and adding email OTP functionality for user authentication.

### Detailed summary
- Added `resolveAlchemyAuthConnector` function to locate the Alchemy Auth connector.
- Introduced `alchemyAuth` function for creating a connector with API key support.
- Implemented email OTP authentication via `sendEmailOtp` and `submitOtpCode` functions.
- Updated documentation for `resolveAlchemyAuthConnector` and related functions.
- Enhanced error handling and type definitions in various files.
- Added UI components for email authentication in the example application.

### DIFF
--- a/packages/smart-accounts/src/light-account/accounts/multi-owner-account.test.ts
+++ b/packages/smart-accounts/src/light-account/accounts/multi-owner-account.test.ts
@@ -23,7 +23,8 @@ import type { LightAccountVersion } from "../registry.js";
 import { toMultiOwnerLightAccount } from "./multi-owner-account.js";
 import { bigIntMultiply } from "../../utils.js";
 
-describe("MultiOwner Light Account Tests", () => {
+// Run sequentially to avoid interference across tests sharing anvil/rundler state
+describe.sequential("MultiOwner Light Account Tests", () => {
   const instance = local070Instance;
   let client: ReturnType<typeof instance.getClient>;
   let salt: bigint = BigInt(poolId());
--- a/packages/smart-accounts/tests/coinbase-account.test.ts
+++ b/packages/smart-accounts/tests/coinbase-account.test.ts
@@ -79,6 +79,7 @@ describe("Viem AA - Coinbase Smart Account", () => {
     // Wait for the user operation to be included
     const receipt = await bundlerClient.waitForUserOperationReceipt({
       hash: userOpHash,
+      timeout: 120_000,
     });
 
     expect(receipt).toBeDefined();
@@ -92,7 +93,7 @@ describe("Viem AA - Coinbase Smart Account", () => {
 
     // Account should be deployed after the first user operation
     expect(await account.isDeployed()).toBe(true);
-  }, 60_000);
+  }, 120_000);
 
   it("should verify signature using EIP-1271 after deployment", async () => {
     const owner = privateKeyToAccount(generatePrivateKey());
--- a/packages/wallet-apis/src/client.e2e.test.ts
+++ b/packages/wallet-apis/src/client.e2e.test.ts
@@ -226,9 +226,7 @@ describe("Client E2E Tests", () => {
 
       expect(result.status).toBe("success");
     },
-    {
-      timeout: 45_000,
-    },
+    60_000,
   );
 
   it.each(sendVariants)(
@@ -267,9 +265,7 @@ describe("Client E2E Tests", () => {
 
       expect(result2.status).toBe("success");
     },
-    {
-      timeout: 90_000,
-    },
+    90_000,
   );
 
   it.each(sendVariants)(
@@ -308,7 +304,7 @@ describe("Client E2E Tests", () => {
 
       expect(result.status).toBe("success");
     },
-    { timeout: 45_000 },
+    60_000,
   );
 
   it.each(sendVariants)(
@@ -351,9 +347,7 @@ describe("Client E2E Tests", () => {
 
       expect(result.status).toBe("success");
     },
-    {
-      timeout: 45_000,
-    },
+    60_000,
   );
 
   const givenTypedData = {
--- a/packages/wallet-apis/src/provider.e2e.test.ts
+++ b/packages/wallet-apis/src/provider.e2e.test.ts
@@ -25,7 +25,7 @@ import {
   waitForCallsStatus,
 } from "viem/actions";
 
-describe("Provider E2E Tests", async () => {
+describe.sequential("Provider E2E Tests", async () => {
   let clientFromProvider: Client<Transport, Chain, Account>;
   let transport: AlchemyTransport;
   let account: Address;
--- a/packages/wallet-apis/vitest.config.ts
+++ b/packages/wallet-apis/vitest.config.ts
@@ -7,6 +7,8 @@ export default mergeConfig(
   defineProject({
     test: {
       name: "alchemy/wallet-apis",
+      singleThread: true,
+      retry: 2,
     },
   }),
 );


packages/connectors-web/src/utils/resolveConnector.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Helper function to locate and return the Alchemy Auth connector from the config
+ *
+ * @param {Config} config - The Wagmi config containing connectors
+ * @param {string} [id] - The connector ID to search for (defaults to "alchemyAuth")
+ * @returns {AlchemyAuthConnector} The Alchemy Auth connector
+ * @throws {Error} If the connector is not found
+ */
+export function resolveAlchemyAuthConnector(
```
thebrianchen, line 21: not sure if we need this helper method, but it was confusing to figure out how to get the alchemy Auth connector when implementing the OTP actions

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -2,95 +2,138 @@ import {
...
+      }
+
+      const accounts = await this.getAccounts();
+      if (accounts.length === 0) {
+        throw new Error("No accounts available from signer");
+      }
+
+      return {
+        accounts,
+        chainId: chainId || 1, // Default to mainnet if not specified
```
dphilipson: (Nit): I'd suggest `chainId ?? 1` or up above
```ts
async connect({ chainId = 1 } = {}) {
```
so it works properly if some chain decides to take chain id 0.

examples/connectors-web-example/src/main.ts
```diff
@@ -29,6 +30,20 @@ document.querySelector<HTMLDivElement>('#app')!.innerHTML = `
...
+    <div id="email-auth">
+      <h2>Email Authentication</h2>
+      
+      <div>
+        <input type="email" id="email-input" placeholder="Enter your email" />
+        <button id="send-otp" type="button">Send OTP</button>
+      </div>
+      <div>
+        <input type="text" id="otp-input" placeholder="Enter OTP code" maxlength="6" />
+        <button id="submit-otp" type="button">Submit OTP</button>
```
jakehobbs, line 44: do you think we should remove the "Alchemy Auth" button from the top of this example page since the expected flow is to connect by using this button now?

packages/connectors-web/src/utils/resolveConnector.ts
jakehobbs: i think this is reasonable and aligns w/ wagmi pretty well.

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -2,95 +2,138 @@ import {
...
   // type Connector,
 } from "@wagmi/core";
+import { createWebAuthClient } from "@alchemy/signer-web";
+import type { AuthClient } from "@alchemy/signer";
+import type { Signer } from "@alchemy/signer";
 
 export interface AlchemyAuthOptions {
-  // web-specific: popup OAuth, iframe, cookie/session opts, etc.
+  /** API key for authentication with Alchemy services */
+  apiKey: string;
```
jakehobbs: do we want to support other connection config options here now, such as a proxy url? or will that come later?

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -2,95 +2,138 @@ import {
...
+ * @param {string} [parameters.iframeElementId] - Optional ID for the iframe element used by Turnkey stamper
+ * @param {string} [parameters.iframeContainerId] - Optional ID for the container element that holds the iframe
  * @returns {ReturnType<typeof createConnector>} A Wagmi connector compatible with `createConfig`.
  */
-export function alchemyAuth(parameters: AlchemyAuthOptions = {}) {
-  // TODO: remove this
-  void parameters;
+export function alchemyAuth(parameters: AlchemyAuthOptions) {
+  if (!parameters.apiKey) {
+    throw new Error("AlchemyAuthOptions.apiKey is required");
```
jakehobbs: do we want to actually throw an error here? i believe this will totally block wagmi from loading the other connectors. might using `_config.emitter.emit` to emit an `"error"` event be safer so that the other connectors continue to function even if ours isn't configured correctly? or are we okay w/ blocking on this?

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -2,95 +2,138 @@ import {
...
     },
 
     async connect({ chainId } = {}) {
-      void chainId;
-      throw new Error("Not implemented: connect");
+      // Connection is handled through the auth flow (sendEmailOtp -> submitOtpCode)
+      // This method is called by wagmi after authentication completes
+      if (!signerInstance) {
+        throw new Error(
+          "No signer available. Please authenticate first using sendEmailOtp and submitOtpCode.",
```
jakehobbs, line 72: is this error too specific since later this will support different auth methods? if fine for now, maybe we can add a `TODO(v5)` here to be sure we change it later?

packages/connectors-web/src/alchemyAuth.ts
jakehobbs: is this common practice in wagmi connectors? i wonder if we should at least try to use a chain from the config first before defaulting to mainnet. what if the app doesn't even have mainnet configured?
```ts
chainId: chainId || _config.chains[0].id || 1, // Default to mainnet if not specified
```
i am wondering it it might be even worth throwing if there is no `chainId` and the `config` doesn't have any chains, and never defaulting to mainnet?

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -2,95 +2,138 @@ import {
...
+
+      return {
+        accounts,
+        chainId: chainId || 1, // Default to mainnet if not specified
+      };
     },
 
     async disconnect() {
-      // Clean up instances, may need to reinitialize a new authClientInstance
+      // Clean up instances
```
jakehobbs, line 98: i don't think everything is getting totally cleaned up here, b/c i get a `Error: Iframe element with ID turnkey-iframe already exists` error if i try to reconnect after disconnecting. maybe this is due to an issue in a lower level auth pkg though and not the connector.

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -2,95 +2,138 @@ import {
...
     async getAccounts() {
-      // TODO: Get the account from the signer
-      return [];
+      if (!signerInstance) {
+        return [];
+      }
+
+      try {
+        // Get the address from the signer
+        const address = signerInstance.getAddress() as `0x${string}`;
```
jakehobbs: nit: could we update `getAddress` in the signer pkg to return an `Address` here so we don't have to cast it here?

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -2,95 +2,138 @@ import {
...
+        return [];
+      }
     },
 
     async getChainId() {
-      // TODO: the connector will need to presist the chain id and instantiate the client/provider with it
-      throw new Error("Not implemented: getChainId");
+      // For now, return mainnet. In the future, this should be configurable
+      // and potentially stored based on user selection or last used chain
+      return 1; // mainnet
```
jakehobbs: might be nice to at least use a chain from the config, but i know this is a WIP so def won't block on that.

packages/connectors-web/src/alchemyAuth.ts
dphilipson: Oh, that's a really good point. For ergonomics, we should probably not throw an error. Compare to e.g. the `walletconnect` connector, where if you don't fill in a walletconnect id then it still lets you use all the other connectors just fine (and this is in fact what happens by default in the wagmi quickstart apps).

packages/connectors-web/src/alchemyAuth.ts
dphilipson: We'll want to call `signerInstance?.disconnect()` here. It's possible that will make the error you're talking about go away. If it doesn't, then it's a bug in the signer package that we should fix.

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -2,95 +2,138 @@ import {
...
     },
 
     async getAccounts() {
-      // TODO: Get the account from the signer
-      return [];
+      if (!signerInstance) {
+        return [];
+      }
+
+      try {
```
dphilipson: Don't think we need a try-catch here. This operation should never fail, and if for some reason it does then it's a legitimate error to bubble up.

packages/signer/src/signer.ts
```diff
@@ -67,6 +67,14 @@ export class Signer {
     return new Signer(apiKey, turnkey, user);
   }
 
+  public getAddress(): string {
```
dphilipson: We might consider making these getter properties, e.g.
```ts
public get address(): string {
```
likewise for `user`.

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -2,95 +2,138 @@ import {
...
+export function alchemyAuth(parameters: AlchemyAuthOptions) {
+  if (!parameters.apiKey) {
+    throw new Error("AlchemyAuthOptions.apiKey is required");
+  }
 
   type Provider = any; // TODO: Define proper provider type
   type Properties = {
-    getAuthClient(): Promise<any>; // TODO: Define proper auth client type
-    getSigner(): Promise<any>; // TODO: Define proper signer type
+    getAuthClient(): Promise<AuthClient>;
```
dphilipson: As it turns out, this doesn't need to be async. We can have it return just `AuthClient`, which also gives us the option of having it be a getter property if we want.

packages/connectors-web/src/utils/resolveConnector.ts
```diff
@@ -0,0 +1,30 @@
+import type { Config, Connector } from "wagmi";
+import type { AuthClient, Signer } from "@alchemy/signer";
+
+/**
+ * Extended connector interface that includes Alchemy Auth-specific methods
+ */
+export interface AlchemyAuthConnector extends Connector {
+  getAuthClient(): Promise<AuthClient>;
```
dphilipson: As mentioned above, does not need to be async.

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: ah thanks for clarifying. thought it was some spec that we had to adhere to.

packages/connectors-web/src/utils/resolveConnector.ts
thebrianchen: changed.

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: was leaving this out for now since dphil was working on a fix to use the new transport.

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: changed to use the emitter and moved the error to be thrown in `getAuthClient()` when you actually need the api key.

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: good catch, added todo

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: Good point! I think throwing is actually the better approach than defaulting to mainnet. Here's
  my reasoning:
  - If no chainId is provided and no chains are configured, that's likely a setup issue that should be caught early rather than defaulting to mainnet
  - Avoid breaking when mainnet isn't supported - If an app only configures testnet chains, defaulting to mainnet could cause confusing downstream errors when the transport/provider can't handle mainnet requests

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: Added it in, thanks for catching. double checking that we want to `await` this right

packages/signer/src/signer.ts
thebrianchen: gonna leave it as `getAddress()` since that seems to be the pattern with other methods like `getTekStamper()`, `getAccounts()`, etc. Granted those are more complex, but open to other ideas here

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: fixed

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: good point, done

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: after implementing `connect({chainId})`, i realized that we need to track the current chain id since we can't always rely on `config`. added a `currentChainId` to the tracking.

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: see my response to jake below

examples/connectors-web-example/src/main.ts
thebrianchen: good catch, removed

packages/connectors-web/src/alchemyAuth.ts
jakehobbs: yep that makes sense to me

jakehobbs (approved):
lgtm

PR #2039, opened 2025-09-12
feat: adds auth session state persistence helper methods

linnall:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on updating the authentication system in the `alchemy` package, enhancing methods for user authentication, session management, and adding support for new features related to passkeys and OAuth providers.

### Detailed summary
- Updated `name` in `vitest.config.ts` from `alchemy/signer` to `alchemy/auth`.
- Added support for `loadAuthSessionState` and `loginWithPasskey` methods.
- Enhanced documentation with `twoslash` code blocks for better examples.
- Added new types and parameters for user authentication details in `types.ts`.
- Updated tests to cover new authentication flows and session states.
- Refactored `AuthSession` to handle different authentication types and session restoration.


packages/signer/src/signer.ts
```diff
@@ -149,6 +150,12 @@ export class Signer {
     (this.turnkey.stamper as TurnkeyStamper).clear?.();
   }
 
+  public async getAuthSessionState(): Promise<AuthSessionState> {
+    this.throwIfDisconnected();
+    // TODO: retrieve and deserialize AuthSessionState from wagmi connector
```
dphilipson: I know this is just a temp comment, but to clarify: the signer shouldn't be retrieving state from the wagmi connector. The connector should depend on signer, not the other way around, and it should be possible to use a signer outside of wagmi. Also the direction described in the comment is wrong: this method is for serializing the state which will be stored in the connector, not deserializing state which is already in the connector.

packages/signer/src/signer.ts
```diff
@@ -149,6 +150,12 @@ export class Signer {
     (this.turnkey.stamper as TurnkeyStamper).clear?.();
   }
 
+  public async getAuthSessionState(): Promise<AuthSessionState> {
```
dphilipson: This probably won't end up needing to be async, but we can worry about that once we have the implementation.

packages/signer/src/types.ts
```diff
@@ -156,3 +156,14 @@ export type AuthProviderCustomization = {
   claims?: string;
   otherParameters?: Record<string, string>;
 };
+
+// Signer state types
+export type AuthSessionState =
+  | {
+      type: "email" | "oauth" | "otp";
```
dphilipson, line 188: We may not end up needing this field, and possibly not `chainId` either. We can figure that out once we have the implementation: the only use of this type is for feeding it to `loadFromAuthSessionState`, so anything we end up not using there we can take out.

packages/signer/src/signer.ts
linnall: noted, thanks for clarifying

packages/signer/src/types.ts
linnall: don't we need the `type` to determine how we generate the signer? or is it possible to generate the signer in one way regardless of authentication method? (haven't looked into passkey authentication yet so I'm unsure about this)

packages/signer/src/signer.ts
```diff
@@ -149,6 +150,12 @@ export class Signer {
     (this.turnkey.stamper as TurnkeyStamper).clear?.();
   }
 
+  public getAuthSessionState(): Promise<AuthSessionState> {
```
thebrianchen: can we call this `getPersistableState()`, since it serializes the signer + state into a `string` or `bytes[]` array.

linnall:
why do my jsdoc comments keep disappearing? 🤔

yarn.lock
```diff
@@ -92,7 +92,7 @@
   resolved "https://registry.yarnpkg.com/@alloc/quick-lru/-/quick-lru-5.2.0.tgz#7bf68b20c0a350f936915fcae06f58e32007ce30"
   integrity sha512-UrcABB+4bUrFABwbluTIBErXwvbsU/V7TZWfmbgJfbkwiBuziS9gxdODUyuiecfdGQ85jglMW6juS3+z5TsKLw==
 
-"@ampproject/remapping@^2.2.0", "@ampproject/remapping@^2.3.0":
+"@ampproject/remapping@^2.2.0":
```
jakehobbs: why did the lock file change when no package.json changed?

packages/auth/src/authClient.ts
```diff
@@ -380,34 +383,77 @@ export class AuthClient {
...
   private async completeAuthWithBundle({
     bundle,
     orgId,
     idToken,
+    authType,
   }: {
     bundle: string;
     orgId: string;
     idToken?: string;
+    authType: Omit<AuthType, "passkey">;
```
jakehobbs: i think what you want here is `Exclude`
```suggestion
    authType: Exclude<AuthType, "passkey">;
```

packages/auth/src/authClient.ts
```diff
@@ -419,6 +465,8 @@ export class AuthClient {
       stamper,
       orgId,
       idToken,
+      bundle,
+      authType: authType as AuthType,
```
jakehobbs: ...then this will work:
```suggestion
      authType,
```

packages/auth/src/authClient.ts
```diff
@@ -380,34 +383,77 @@ export class AuthClient {
...
+   * const authSession = await authClient.loginWithPasskey("existing-credential-id");
    * ```
    */
-  public async loginWithPasskey(): Promise<AuthSession> {
+  public async loginWithPasskey(credentialId?: string): Promise<AuthSession> {
     // TODO: figure out what the current passkey code is doing.
+    // For new passkey authentication, credentialId would be undefined initially
     const stamper = await this.createWebAuthnStamper({
-      credentialId: undefined,
+      credentialId: credentialId ?? undefined,
```
jakehobbs: nit: the type here is already `string | undefined`, so you can simplify this:
```suggestion
      credentialId,
```

packages/auth/src/authClient.ts
```diff
@@ -380,34 +383,77 @@ export class AuthClient {
...
+        const { orgId, idToken } = user;
+        return await this.completeAuthWithBundle({
+          bundle,
+          orgId,
+          idToken,
+          authType: type,
+        });
+      }
+    }
+    return undefined;
```
jakehobbs: nit: would strongly prefer to check the expiration first & return early if it's expired, to get rid of some nesting. i.e.

```suggestion
    if (expirationDateMs <= Date.now()) {
      return undefined
    }
    if (type === "passkey") {
      ...
    }
    const { bundle } = state;
    const { orgId, idToken } = user;
    ...
```

packages/auth/src/authSession.ts
```diff
@@ -105,59 +220,196 @@ export class AuthSession {
...
+   *
+   * @param {string} authenticatorId - The ID of the authenticator/passkey to remove
+   * @returns {Promise<void>} A promise that resolves when the passkey is removed
+   */
   public async removePasskey(authenticatorId: string): Promise<void> {
     this.throwIfDisconnected();
     return notImplemented(authenticatorId);
   }
 
+  /**
```
jakehobbs, line 347: thanks for adding all of this jsdoc! 🚀

packages/auth/src/authSession.ts
```diff
@@ -105,59 +220,196 @@ export class AuthSession {
...
+   * ```
+   */
+  public getAuthSessionState(): string {
+    this.throwIfDisconnected();
+
+    // Calculate expiration time (24 hours from now as default)
+    const expirationDateMs = Date.now() + 24 * 60 * 60 * 1000;
+
+    // Use stored authType or default to "otp" for backward compatibility
+    const type = this.authType || "otp";
```
jakehobbs, line 391: style nit (i prefer to use `??` unless you are specifically needing to handle falsey values other than `null`/`undefined`:
```suggestion
    const type = this.authType ?? "otp";
```

packages/auth/src/authSession.ts
```diff
@@ -105,59 +220,196 @@ export class AuthSession {
...
+   * if (savedState) {
+   *   const restoredSession = await authClient.loadAuthSessionState(JSON.parse(savedState));
+   * }
+   * ```
+   */
+  public getAuthSessionState(): string {
+    this.throwIfDisconnected();
+
+    // Calculate expiration time (24 hours from now as default)
+    const expirationDateMs = Date.now() + 24 * 60 * 60 * 1000;
```
jakehobbs, line 388: the actual TK "session" has an expiry too, right? do we care about that here or not really?

packages/auth/src/authClient.ts
```diff
@@ -380,34 +383,77 @@ export class AuthClient {
...
+        if (!credentialId) {
+          throw new Error(
+            "Credential ID is required for passkey authentication",
+          );
+        }
+        return await this.loginWithPasskey(credentialId);
+      } else {
+        const { bundle } = state;
+        const { orgId, idToken } = user;
+        return await this.completeAuthWithBundle({
```
jakehobbs: excuse my ignorance here, but do you know how long this bundle is valid for? it seems a bit surprising to me that we can keep reusing it. is this similar to what we did in v4? (i don't have the most knowledge about this.)

packages/auth/src/authClient.ts
linnall: genius (brb looking into `Omit` vs `Exclude`)

packages/auth/src/authSession.ts
```diff
@@ -107,59 +222,196 @@ export class AuthSession {
...
+   * @example
+   * ```ts
+   * const sessionState = authSession.getAuthSessionState();
+   * localStorage.setItem('authSession', sessionState);
+   *
+   * // Later restore:
+   * const savedState = localStorage.getItem('authSession');
+   * if (savedState) {
+   *   const restoredSession = await authClient.loadAuthSessionState(JSON.parse(savedState));
+   * }
```
jakehobbs, line 381: it seems a bit inconsistent imo for `getAuthSessionState` to take care of stringifying the json for you, but `loadAuthSessionState` expects you to parse it first. curious your thoughts on that.

packages/auth/src/authClient.ts
jakehobbs: `Omit` is to remove a specific property from an object. (the opposite of `Pick`)
`Exclude` is to exclude a member from a union. (the opposite of `Extract`)

linnall: I see, TIL `Omit` for objects types, `Exclude` for union types

packages/auth/src/authSession.ts
linnall: I agree, we should parse within `loadAuthSessionState` to stay consistent. This was an oversight on my part. Good catch!

packages/auth/src/authClient.ts
linnall: the bundle itself doesn't have an expiration date directly attached to it, so we are responsible for setting its expiration date which we store in the `AuthSessionState`. this is also what we do in v4 (see [here](https://github.com/alchemyplatform/aa-sdk/blob/415bfdf7cbc633ce1b136ee16b512b30c3af2553/account-kit/signer/src/session/types.ts#L7))

yarn.lock
linnall: I suspect it has something to do with my local linting config, will remove this change

packages/auth/src/authClient.ts
linnall: cc: @dphilipson please correct me if I'm wrong

packages/signer/src/types.ts
linnall: resolving this because we need to use `type` to determine how we reconstruct an instance of `AuthSessionState` e.g. passkey does not use a bundle, but oauth, otp, and email do

packages/auth/src/authClient.ts
jakehobbs: ok ty!

packages/auth/src/authClient.ts
```diff
@@ -404,42 +403,62 @@ export class AuthClient {
...
+   * The method will validate the session expiration and handle different authentication types appropriately.
+   *
+   * @param {string} state - The serialized authentication session state as a JSON string
+   * @returns {Promise<AuthSession | undefined>} A promise that resolves to an AuthSession instance if the session is valid and not expired, undefined if expired
+   *
+   * @throws {Error} Throws an error if the session state is invalid or if credential injection fails
+   * @throws {Error} For passkey authentication, throws if credentialId is missing from the session state
+   *
+   * @example
+   * ```ts
```
jakehobbs:
```suggestion
   * ```ts twoslash
```

jakehobbs (approved):
lgtm! not sure if you want @dphilipson to give a final review too or not.

(disabled auto-merge just in caes, but feel free to turn it back on.)

packages/auth/src/authClient.ts
linnall: updated all the jsdoc comments in `authSession` and `authClient` to use `twoslash` and regenerated the docs

packages/auth/src/authClient.ts
```diff
@@ -380,34 +382,97 @@ export class AuthClient {
...
+   *   const authSession = await authClient.loadAuthSessionState(sessionJson);
+   *   if (authSession) {
+   *     console.log('Session restored successfully');
+   *   } else {
+   *     console.log('Session expired');
+   *   }
+   * }
+   * ```
+   */
+  public async loadAuthSessionState(
```
thebrianchen, line 438: can we rename to `restoreAuthSession()? It's a clearer intent that you're deserializing/restoring a session. the state is something that we pass in in order to restore, but not something that we're loading

packages/auth/src/authSession.ts
```diff
@@ -107,59 +222,196 @@ export class AuthSession {
...
+   * ```
+   */
+  public getAuthSessionState(): string {
```
thebrianchen, line 384: not a huge fan of this naming, this feels like a getter that would return the `AuthSessionState` object directly, and doesn't indicate that it's serializing to a JSON string. i'd prefer something shorter that's descriptive like `serialize()`  or `getSerializedState()`.

packages/auth/src/authClient.ts
```diff
@@ -380,34 +382,97 @@ export class AuthClient {
...
+    state: string,
+  ): Promise<AuthSession | undefined> {
+    const parsedState = JSON.parse(state);
```
dphilipson, line 441: (Minor) I'd suggest declaring the type
```ts
const parsedState: AuthSessionState = JSON.parse(state);
```
for readability and so the type system helps make sure the fields you're reading out here are the same ones you're putting in elsewhere.

packages/auth/src/authSession.ts
```diff
@@ -107,59 +222,196 @@ export class AuthSession {
...
+    this.throwIfDisconnected();
+
+    // Calculate expiration time (24 hours from now as default)
+    const expirationDateMs = Date.now() + 24 * 60 * 60 * 1000;
```
dphilipson, line 388: We're going to want to update this later, right? The expiration time won't always be 24 hours from the time the state is serialized. We can do that when we add expiration handling in general, but I might add a TODO.

packages/auth/src/authSession.ts
```diff
@@ -107,59 +222,196 @@ export class AuthSession {
...
+    // Use stored authType or default to "otp" for backward compatibility
+    const type = this.authType || "otp";
+
+    if (type === "passkey") {
+      return JSON.stringify({
+        type: "passkey",
+        user: this.user,
+        expirationDateMs,
+        credentialId: this.credentialId,
+      });
```
dphilipson, line 399: Related to the above, I'd consider defining a type for this object and referencing it again in the deserialization code so the type system can check that the fields being put in are the same as the fields being read out.

packages/auth/src/authSession.ts
```diff
@@ -66,18 +139,60 @@ export class AuthSession {
     });
     // TODO: combine whoami response with idToken to get the full user object.
     // For now, just return the whoami response.
```
dphilipson, line 141: (Very minor) This comment is a bit outdated since we're not just returning the whoami response. But there's still a todo here where we'll eventually read stuff like email out of the id token to display as the user name.

packages/auth/src/authSession.ts
linnall: good catch! I'll add a todo note

packages/auth/src/authSession.ts
linnall: can you clarify how this differs from [this comment](https://github.com/alchemyplatform/aa-sdk/pull/2039/files/b97e3831619946ca47aae086a6162ee333899be9#r2370097385) that you made above?

packages/auth/src/authSession.ts
linnall: updated in #2090

packages/signer/src/signer.ts
linnall: updated to `getSerializedState` in #2090

linnall:
please find the comments after this pr was merged addressed in #2090

packages/auth/src/authSession.ts
linnall: I think I have addressed this comment [here](https://github.com/alchemyplatform/aa-sdk/blob/1de57467df20888b3f83946cad828820b38b5033/packages/auth/src/authSession.ts#L395)

PR #2074, opened 2025-09-17
feat(v5): 1193 provider for alchemyAuth

jakehobbs:
# Pull Request Checklist


packages/auth/src/provider.ts
```diff
@@ -0,0 +1,142 @@
...
+import {
+  BaseError,
+  InvalidRequestError,
+  MethodUnsupportedError,
+  createEip1193HandlerFactory,
+} from "@alchemy/common";
+import type { ExtractRpcMethod } from "./types";
+import type { AuthSession } from "./authSession.js";
+
+export type AlchemyAuth1193Methods = [
```
thebrianchen, line 19: should we be exporting this in `index.ts`? anything else to export?

packages/auth/src/types.ts
```diff
@@ -156,3 +156,12 @@ export type AuthProviderCustomization = {
   claims?: string;
   otherParameters?: Record<string, string>;
 };
+
+export type ExtractRpcMethod<
```
thebrianchen, line 160: consolidate with the export in `wallet-apis`?

packages/common/src/utils/createEip1193HandlerFactory.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Creates a typed handler factory for EIP1193 request methods.
+ * This helps with Viem's typing within custom EIP1193 request functions by
+ * automatically casting input params and ensuring the result matches what is required.
+ *
+ * @returns {(params: unknown) => unknown} A function that creates a handler factory for specific methods
+ */
+export const createEip1193HandlerFactory =
```
thebrianchen, line 8: this is a thing of beauty

packages/connectors-web/src/utils/resolveConnector.ts
```diff
@@ -1,5 +1,6 @@
 import type { Config, Connector } from "wagmi";
 import type { AuthClient, AuthSession } from "@alchemy/auth";
+import { getConnectors } from "@wagmi/core";
```
thebrianchen, line 3: did not know this was a thing, thanks!

packages/wagmi-core/src/actions/sendEmailOtp.ts
```diff
@@ -22,5 +22,6 @@ export async function sendEmailOtp(
   const connector = resolveAlchemyAuthConnector(config);
 
   const authClient = connector.getAuthClient();
+  // TODO(v5): this fails if the account doesn't exist yet. do we want to handle creating accounts here?
```
thebrianchen, line 25: i thought that's a bug with the signer/auth implementation, and that a new account should automatically be created since we don't have a separate flow for new accounts. @dphilipson is that correct?

packages/wallet-apis/src/provider.ts
```diff
@@ -61,7 +64,7 @@ export const createEip1193ProviderFromClient = <
     try {
       switch (method) {
         case "eth_chainId": {
-          return handler<"eth_chainId">(async () => {
+          return await handler<"eth_chainId">(async () => {
```
thebrianchen, line 67: went down a rabbit hole with this, didn't know that `return await` is the new meta in try/catch now. ty for fixing

jakehobbs: yeahh..... this has bitten me a few times tbh. i only learned this a few months ago. didn't notice until i added tests & the errors were not the expected type.

packages/wagmi-core/src/actions/sendEmailOtp.ts
jakehobbs: i flagged to @dphilipson on slack a couple of days ago too.

packages/connectors-web/src/utils/resolveConnector.ts
jakehobbs: np, i just learned about it this week. looks like all it does is memoize the connectors, which could lead to a perf improvement for consumers.

packages/common/src/utils/createEip1193HandlerFactory.ts
jakehobbs: haha thanks, i'm surprised viem doesn't already do something like this itself tbh.

packages/auth/src/provider.ts
jakehobbs: i'll def export the types. good catch.

not sure if we want to export the actual implementation or force everyone to access it via an instance of the `authSession`. i think i prefer to not export for now unless we need it later, to avoid exposing extra internal implementations.

packages/auth/src/types.ts
jakehobbs: ah yes, thanks for catching that. didn't notice that's not what i did.

packages/common/src/utils/types.ts
```diff
@@ -4,80 +4,11 @@ export type Never<T> = T extends object
     }
   : never;
 
-export type RequiredBy<TType, TKeys extends keyof TType> = Required<
```
jakehobbs, line 7: removing these since they are unused. i think there were just blindly copied over from v4 during scaffolding, but i'd rather not have them if we aren't using them.

PR #2079, opened 2025-09-18
feat: HTTP client for Signer API

dphilipson:
Defines an `AlchemyHttpClient` type as an alternative to `AlchemyTransport` for hitting non-JSON-RPC APIs. Then define a schema for the Signer API endpoints, and use the new client with this schema in the auth packages, replacing the dummy `dev_request` method that was previously used.


packages/common/src/http/httpClient.ts
```diff
@@ -0,0 +1,59 @@
...
+  /** Custom RPC URL (optional - defaults to Alchemy's chain-agnostic URL, but can be used to override it) */
+  url?: string;
+  /** Custom headers to be sent with requests */
+  headers?: HeadersInit;
+};
+
+/**
+ * A client for making requests to Alchemy's non-JSON-RPC endpoints.
+ */
+export class AlchemyHttpClient<Schema extends HttpRequestSchema> {
```
thebrianchen: I wonder if this class is more aptly called `AlchemyRestClient`, since the json-rpc requests are technically sent over http as well.

packages/common/src/http/types.ts
```diff
@@ -0,0 +1,42 @@
+import type { Prettify } from "viem";
+
+export type HttpRequestSchema = readonly {
```
thebrianchen: can we consolidate this with the `HttpRequestSchema` in `transport/types.ts` ([github link](https://github.com/alchemyplatform/aa-sdk/blob/dp/http-client/packages/common/src/transport/types.ts#L34))?

iiuc, the one there returns a wrapped result object, but the one here returns `T` or throws an error.

packages/common/src/http/httpClient.ts
```diff
@@ -0,0 +1,59 @@
...
+  private readonly url: string;
+  private readonly headers: Headers;
+
+  constructor({ apiKey, jwt, url, headers }: AlchemyHttpClientParams) {
```
thebrianchen: With this change, do we still need `makeHttpRequest` on `alchemyTransport`? Are there any internal use cases for leaving that in, or do we just remove it and standardize `alchemyTransport` to only use `rpcSchema`?

packages/auth/src/authClient.ts
```diff
@@ -1,23 +1,25 @@
...
-  // transport: AlchemyTransport;
-  // TODO: this is temporary for testing before the transport is ready.
-  /** API key for authentication with Alchemy services */
-  apiKey: string;
-  /** Function to create a TEK (Turnkey Ephemeral Key) stamper */
+  /** API key for authentication with Alchemy services. One of apiKey, jwt, or url must be provided. */
+  apiKey?: string;
+  /** JWT token for authentication with Alchemy services. One of apiKey, jwt, or url must be provided. */
+  jwt?: string;
+  /** Custom RPC URL (optional - defaults to Alchemy's chain-agnostic URL, but can be used to override it) */
```
jakehobbs: nit: should this still say "RPC" URL?

packages/auth/src/authClient.ts
```diff
@@ -299,11 +307,16 @@ export class AuthClient {
...
   public async loginWithOauth(
-    params: LoginWithOauthParams,
+    params: LoginWithOauthParams
   ): Promise<AuthSession> {
     const { targetPublicKey } = await this.getTekStamper();
-    const oauthConfig = await this.dev_request("prepare-oauth", {
-      nonce: getOauthNonce(targetPublicKey),
+    // const oauthConfig = await this.sign("prepare-oauth", {
+    //   nonce: getOauthNonce(targetPublicKey),
+    // });
```
jakehobbs: intentionally leaving this?

packages/common/src/http/httpClient.ts
```diff
@@ -0,0 +1,59 @@
+import { FetchError } from "../errors/FetchError.js";
+import { ServerError } from "../errors/ServerError.js";
+import { withAlchemyHeaders } from "../utils/headers.js";
+import type { HttpRequestFn, HttpRequestSchema } from "./types.js";
+
+export const ALCHEMY_RPC_URL = "https://api.g.alchemy.com";
```
jakehobbs: nit: should this name be different considering it handles some calls that aren't json-rpc?

packages/auth/src/authClient.ts
jakehobbs: are these comments correct? looking at the implementation, it looks like we allow a `url` in addition to either an `apiKey` or a `jwt`?

packages/common/src/utils/headers.ts
```diff
@@ -0,0 +1,48 @@
...
+export function withAlchemyHeaders({
+  headers,
+  apiKey,
+  jwt,
+}: WithAlchemyHeadersParams): Record<string, string> {
+  return {
+    ...convertHeadersToObject(headers),
+    "Alchemy-AA-Sdk-Version": VERSION,
+    ...(jwt ? { Authorization: `Bearer ${jwt}` } : {}),
+    ...(apiKey ? { Authorization: `Bearer ${apiKey}` } : {}),
```
jakehobbs: not sure why, but we have a test that says the `jwt` should take precedence:
```
tests/transport/alchemyTransport.test.ts > Alchemy Transport Tests > Error Handling > should handle config with both apiKey and jwt (JWT takes precedence)

```

packages/common/src/http/httpClient.ts
dphilipson: Yeah, good idea, I'll change that. Although I'll say that one of the points of JSON-RPC is that it need not be sent over http, as you probably know from dealing with websocket headaches lol.

packages/common/src/http/types.ts
dphilipson: Oh, I meant to delete that one, thanks! It's only used to extend the alchemy transport with http methods, which we're taking out.

I didn't see a point in returning the error union type so I had it just return a result or throw. If anyone disagrees I can change it back, but that seems like unfamiliar behavior for no particular reason.

packages/common/src/http/httpClient.ts
dphilipson: You're right, I'll delete `makeHttpRequest` from the transport.

packages/auth/src/authClient.ts
dphilipson: Will change. The distinction is good to be able to distinguish the APIs, but I think the terminology is silly: posting to an endpoint called `/sign-payload` should be considered an RPC call IMO.

packages/common/src/http/httpClient.ts
dphilipson: As above, yeah I'll change it to keep terminology consistent, although I find our terminology a little silly.

packages/auth/src/authClient.ts
dphilipson: It's technically correct, but should say "at least one" for clarity.

packages/common/src/utils/headers.ts
dphilipson: Huh, I actually appreciate the tests catching behavior changes like this. Will change.

packages/auth/src/authClient.ts
jakehobbs: ahhh thanks. i think my brain initially automatically read it as "_only_ one of ... must be provided"

packages/common/src/utils/headers.ts
thebrianchen: ah yeah i ported some of those tests straight from v4, others were autogenned.

jakehobbs (approved):
looks great!

thebrianchen (approved):
thanks for taking this on!

PR #2090, opened 2025-09-22
refactor(v5): renames state persisting helper methods

linnall:
# Pull Request Checklist


thebrianchen (approved):

packages/auth/src/authClient.ts
```diff
@@ -438,7 +439,7 @@ export class AuthClient {
   public async loadAuthSessionState(
     state: string,
   ): Promise<AuthSession | undefined> {
-    const parsedState = JSON.parse(state);
+    const parsedState: AuthSessionState = JSON.parse(state);
```
thebrianchen: it's not safe enough to just assume that the user is going to pass in a string that is the correct state. Instead of doing a blind JSON.parse and cast to `AuthSessionState`, we should add an additional method to validate that the state passed in can be parsed to a valid `AuthSessionState` and error if the srting is invalid.

linnall: added a `deserializeState` function. should we log the parsed state in the error if it can successfully be parsed as JSON but isn't a valid `AuthSessionState` object?

thebrianchen: lgtm! but nah don't have to, i think that's something that the end dev can easily try/catch and see for themselves if they want to debug.

PR #2096, opened 2025-09-23
feat: add persistence to alchemyAuth

thebrianchen:
Adding `AuthSession` persistence for `alchemyAuth` connector. This allows the connected auth session to persist across page loads. The connector stores the serialized auth session with wagmi's local storage API and deserializes and recreates the auth session when reconnecting.

Will fully add jsdoc comments and unit tests in a subsequent PR. wanting to get this code in first to unblock downstream components.

You can test this out in the sample app by logging in and refreshing the page. There are also debug tools at the bottom to view/clear/test local storage contents.

Screenshot 2025-09-23 at 9 17 02 AM

# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing the persistence and management of authentication sessions in the `alchemyAuth` connector. It introduces new storage management functions and updates the UI for viewing local storage contents.

### Detailed summary
- Added `PersistedAuthSession` type and schema for session validation.
- Implemented functions to get, set, and clear stored authentication sessions.
- Enhanced error handling for session storage operations.
- Introduced a UI section for viewing local storage data in `connectors-web-example`.
- Updated `alchemyAuth` to manage session restoration and persistence.
- Added event listener for storage changes to synchronize sessions across tabs.


packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -25,6 +27,107 @@ export interface AlchemyAuthOptions
...
+  chainId: number;
+};
+
+// Storage configuration
+const STORAGE_KEY = "alchemyAuth.authSession" as const;
+
+// Zod schema for validating persisted auth session
+const PersistedAuthSessionSchema = z.discriminatedUnion("type", [
+  z.object({
+    v: z.literal(1),
```
linnall: is there a particular reason why we are adding a versioning field here? otherwise we can move the `AuthSessionStateSchema` in `authClient` into `utils.ts` and export it for use here

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -197,9 +421,11 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
-        // Check if we have a valid authSession instance
-        return authSessionInstance !== undefined;
+      async isAuthorized(): Promise<boolean> {
+        if (authSessionInstance) return true;
+
+        const stored = await getStoredAuthSession(config.storage);
+        return !!stored && Date.now() < stored.expirationDateMs;
       },
 
       async switchChain({ chainId }) {
```
blakecduncan, line 361: Do we need to persist the new chain when this is called?

examples/connectors-web-example/src/main.ts
```diff
@@ -254,6 +261,27 @@ function handleClearStorage() {
   }
 }
 
+// Helper function to resolve Alchemy connector
+function resolveAlchemyAuthConnector(config: any) {
+  return config.connectors.find((c: any) => c.id === "alchemyAuth");
+}
```
jakehobbs: nit (i don't care that much since this is just an example app though)
```suggestion
function resolveAlchemyAuthConnector(config: Config) {
  return config.connectors.find((c) => c.id === "alchemyAuth");
}
```

jakehobbs: any reason for not just importing this from the other pkg btw?
```ts
import { resolveAlchemyAuthConnector } from "@alchemy/connectors-web";
```

packages/auth/src/utils.ts
```diff
@@ -241,3 +243,28 @@ export function getOauthProviderUrl(args: GetOauthProviderUrlArgs): string {
 
   return `${urlPath?.replace(/\/$/, "")}?${searchParams}`;
 }
+
+/**
+ * Zod schema for validating User objects
+ */
+const UserSchema = z.object({
+  email: z.string().optional(),
```
jakehobbs: not sure if we care to be this strict, but just wanted to flag that zod has an `email()` we can use here, if we want
```suggestion
  email: z.string().email().optional(),
```

packages/auth/src/utils.ts
```diff
@@ -241,3 +243,28 @@ export function getOauthProviderUrl(args: GetOauthProviderUrlArgs): string {
...
+  claims: z.record(z.unknown()).optional(),
+});
+
+/**
+ * Type guard to validate User objects at runtime using Zod
+ *
+ * @param {any} obj - The object to validate as a User
+ * @returns {obj is User} True if the object is a valid User, false otherwise
+ */
+export function isValidUser(obj: any): obj is User {
```
jakehobbs: nit: i prefer to almost always use `unknown` here instead of `any`
```suggestion
export function isValidUser(obj: unknown): obj is User {
```

examples/connectors-web-example/src/main.ts
```diff
@@ -489,19 +531,344 @@ function setupSessionControls() {
...
+    if (isConnected && addressMatch && chainMatch && notExpired) {
+      updateStatus(
+        "test-status",
+        "✅ Persistence test PASSED - Session restored successfully!",
+      );
+    } else {
+      const issues = [];
+      if (!isConnected) issues.push("not connected");
+      if (!addressMatch) issues.push("address mismatch");
+      if (!chainMatch) issues.push("chain mismatch");
```
jakehobbs: seems like chain persistence has a bug:
```
Chain consistency: FAIL
Persistence test FAILED: chain mismatch
```

packages/connectors-web/src/alchemyAuth.ts
blakecduncan: We could possibly put it in the `onChainChanged`.  Not sure if you have any thoughts. I'm not sure what is better off the top of my head

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+          const authSessionState: AuthSessionState =
+            JSON.parse(authSessionStateJson);
+
+          const toPersist: PersistedAuthSessionV1 =
+            authSessionState.type === "passkey"
+              ? {
+                  v: 1,
+                  type: "passkey",
+                  expirationDateMs: authSessionState.expirationDateMs,
+                  chainId: currentChainId || config.chains[0]?.id || 1,
```
blakecduncan: Do we need to fallback to chain ID 1 here?  should we throw if nether `clearStoredAuthSession` or `config.chains[0]?.id` are defined?

packages/connectors-web/src/alchemyAuth.ts
dphilipson: Agreed with Linna, I think this is misplaced. `connectors-web` shouldn't be inspecting the contents of the persisted schema, it should be treating it like a black box. The idea behind the auth interfaces is that they say "if you want to persist state, store this blob, and then later give me back the blob and I'll rehydrate." Consumers should not be concerned with the contents of the blob: if they are, it hurts our ability to change it in the future.

Any validation of the contents of the stored state should occur in the `auth` package.

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+            // (it doesn't fire in the same tab that made the change)
+            if (e.key?.endsWith(STORAGE_KEY) && e.newValue === null) {
+              // Our auth session was cleared in another tab - disconnect this tab too
+              // This prevents scenarios where Tab A logs out but Tab B stays "connected"
+              // with a stale session that would fail on actual usage
+              void this.disconnect();
+            }
+          };
+
+          window.addEventListener("storage", handleStorageChange);
```
dphilipson: We should remove this listener when disconnecting.

examples/connectors-web-example/src/main.ts
```diff
@@ -489,19 +531,344 @@ function setupSessionControls() {
...
+          try {
+            await connect(config, { connector: alchemyConnector });
+          } catch (connectError) {
+            // If direct connect fails, try reconnect as fallback
+            const reconnectPromise = reconnect(config);
+            const timeoutPromise = new Promise((_, reject) => {
+              setTimeout(
+                () => reject(new Error("Reconnect timeout after 15s")),
+                15000,
+              );
```
jakehobbs: curious the reasoning for trying `connect` first, then failing over to `reconnect`. i would expect that we just need to call `reconnect` here based on [wagmi docs](https://wagmi.sh/core/api/actions/reconnect).

this seems like a lot of logic to expect apps to implement to get reconnection working w/ the alchemy connector.

also, at the react level wagmi exposes a `reconnectOnMount` option (defaults to `true`) that automatically kicks off the reconnection in react, so we need to make sure that works w/ our connector. [it just calls `reconnect`](https://github.com/wevm/wagmi/blob/331db72fe42f9774403addc0806182b85a0103d7/packages/core/src/hydrate.ts#L53).

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+          "No auth session available. Please authenticate first using sendEmailOtp and submitOtpCode, or loginWithOauth.",
         );
 
+        // Refresh persisted snapshot (e.g., new expiration after token refresh)
+        try {
+          // Get the real auth session state
+          const authSessionStateJson =
+            authSessionInstance!.getSerializedState();
+          const authSessionState: AuthSessionState =
+            JSON.parse(authSessionStateJson);
```
dphilipson: As above, this is a misuse of the auth session API. We're returning the session state as a string and not an object for a reason.

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+
+          const toPersist: PersistedAuthSessionV1 =
```
blakecduncan: nit - it looks like this logic is duplicated between `connect` and `setAuthSession`.  I think a helper function here could help with maintainability

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -25,6 +27,107 @@ export interface AlchemyAuthOptions
   apiKey?: string;
 }
 
+// Persistence types - minimal, versioned, encrypted snapshot for storage
+export type PersistedAuthSessionV1 = AuthSessionState & {
+  /** Version for migration handling */
+  v: 1;
```
jakehobbs: thoughts on naming of this? is this standard? slight preference for `version` or `ver`, if not. or maybe even `_v` to signify it's an internal versioning thing?

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -25,6 +27,107 @@ export interface AlchemyAuthOptions
...
+  /** Chain ID for connector-specific context */
+  chainId: number;
+};
+
+// Storage configuration
+const STORAGE_KEY = "alchemyAuth.authSession" as const;
+
+// Zod schema for validating persisted auth session
+const PersistedAuthSessionSchema = z.discriminatedUnion("type", [
```
jakehobbs: how often do we expect the version of this to change? think it's worth first discriminating on the version field, then discriminating on the type?

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -25,6 +27,107 @@ export interface AlchemyAuthOptions
...
+    expirationDateMs: z.number(),
+    chainId: z.number(),
+    user: z.any().refine(isValidUser, "Invalid user object"),
+  }),
+  z.object({
+    v: z.literal(1),
+    type: z.literal("passkey"),
+    expirationDateMs: z.number(),
+    chainId: z.number(),
+    user: z.any().refine(isValidUser, "Invalid user object"),
```
jakehobbs: is this `isValidUser` helper used anywhere other than within a zod schema? if not, feels a little odd to me to use `z.any().refine()` instead of just `user: UserSchema` directly, unless there's another reason i'm missing.

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -197,9 +421,11 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
       },
 
-      async isAuthorized() {
-        // Check if we have a valid authSession instance
-        return authSessionInstance !== undefined;
+      async isAuthorized(): Promise<boolean> {
+        if (authSessionInstance) return true;
+
+        const stored = await getStoredAuthSession(config.storage);
+        return !!stored && Date.now() < stored.expirationDateMs;
```
dphilipson: Hm, it's interesting that we need to check this here- if this is required, we probably should store expiration time separately from the persistence blob. But it's unclear to me if this is the right way to handle expiration, since it may end up being redundant with the expiration handling in the auth packages that we'll be writing. Might want to think this over with @linnall.

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -25,6 +27,107 @@ export interface AlchemyAuthOptions
...
+    type: z.literal("passkey"),
+    expirationDateMs: z.number(),
+    chainId: z.number(),
+    user: z.any().refine(isValidUser, "Invalid user object"),
+    credentialId: z.string().optional(),
+  }),
+]);
+
+// Type guard for runtime validation of persisted auth session using Zod
+function isPersistedAuthSession(obj: any): obj is PersistedAuthSessionV1 {
```
jakehobbs: nit: it doesn't really make much of a difference here, but i like to avoid `any` as much as reasonably possible. (i also feel like it makes more sense here semantically to go from `unknown` to some known type.)
```suggestion
function isPersistedAuthSession(obj: unknown): obj is PersistedAuthSessionV1 {
```

jakehobbs: although i guess i could see the argument that `any` makes sense here since `JSON.parse` returns `any`, if you feel strongly

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -25,6 +27,107 @@ export interface AlchemyAuthOptions
...
+): Promise<void> {
+  if (!storage) return;
+
+  try {
+    await storage.removeItem(STORAGE_KEY);
+  } catch (error) {
+    console.warn("Failed to clear stored session:", error);
+  }
+}
+
```
jakehobbs: nit: any thoughts on moving some of the session stuff out to another file?

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -25,6 +27,107 @@ export interface AlchemyAuthOptions
   apiKey?: string;
 }
 
+// Persistence types - minimal, versioned, encrypted snapshot for storage
+export type PersistedAuthSessionV1 = AuthSessionState & {
```
blakecduncan: nit: I think it would be nice to have a `store` folder and put some of this logic in there.  I assume we may be able to reuse some of this logic between connectors.  Also will help keep this file shorter.  No worries if you don't feel that's worth it though

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+            await client.restoreAuthSession(sessionStateJson);
+
+          if (authSessionInstance) {
+            currentChainId = persisted.chainId;
+            return true;
+          } else {
+            await config.storage?.removeItem(STORAGE_KEY);
+            return false;
+          }
+        } catch (loadError) {
```
jakehobbs: `loadError` is never used

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+          await config.storage?.removeItem(STORAGE_KEY);
+          return false;
+        }
+      } catch (error) {
```
jakehobbs, line 116: `error` is never used

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+        // Check if session is still valid (not expired)
+        if (Date.now() >= persisted.expirationDateMs) {
+          await clearStoredAuthSession(config.storage);
+          return false;
+        }
+
+        // Attempt to restore the auth session
+        const client = getAuthClient();
+        try {
+          const sessionStateJson = JSON.stringify(persisted);
```
jakehobbs: it feels a little odd to me that we already have a parsed persisted auth session, then we turn it back into a string so that someone else can re-parse it.

jakehobbs: thinking about this a little more, i wonder if it might be nice for `client.restoreAuthSession` to take `string | AuthSessionState`, or if that is overgeneralizing it.

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+          authSessionInstance =
+            await client.restoreAuthSession(sessionStateJson);
+
+          if (authSessionInstance) {
+            currentChainId = persisted.chainId;
+            return true;
+          } else {
+            await config.storage?.removeItem(STORAGE_KEY);
+            return false;
+          }
```
jakehobbs: style nit: this already has quite a bit of nesting, so maybe an early return & avoiding the `else` here would make it a little easier to follow:
```ts
if (!authSessionInstance) {
  await config.storage?.removeItem(STORAGE_KEY);
  return false;
}

currentChainId = persisted.chainId; 
return true;
```

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+
+        // Check if session is still valid (not expired)
+        if (Date.now() >= persisted.expirationDateMs) {
+          await clearStoredAuthSession(config.storage);
+          return false;
+        }
+
+        // Attempt to restore the auth session
+        const client = getAuthClient();
+        try {
```
jakehobbs: can we get rid of the nested `try` blocks here? looks like we have the exact same error handling for both of them (removing the storage & returning `false).

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+
+          window.addEventListener("storage", handleStorageChange);
+        }
       },
 
       async connect({ chainId } = {}) {
-        // Connection is handled through the auth flow (sendEmailOtp -> submitOtpCode)
-        // This method is called by wagmi after authentication completes
-        // TODO(v5): Update error message to reflect different auth methods available.
+        // Perform session restoration here if needed
```
jakehobbs: not sure if it helps you much or changes anything, but just want to flag that `connect` receives an `{isReconnecting: true}` param whenever `reconnect` is called. i think there's likely some behavior we want to differ based on if we are explicitly reconnecting or not, but i'll defer to you.

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
-          "No signer available. Please authenticate first using sendEmailOtp and submitOtpCode, or loginWithOauth.",
+          authSessionInstance,
+          "No auth session available. Please authenticate first using sendEmailOtp and submitOtpCode, or loginWithOauth.",
         );
 
+        // Refresh persisted snapshot (e.g., new expiration after token refresh)
+        try {
+          // Get the real auth session state
+          const authSessionStateJson =
+            authSessionInstance!.getSerializedState();
```
jakehobbs: nit: shouldn't need this `!` since you used `assertNotNullish`
```suggestion
            authSessionInstance.getSerializedState();
```

packages/connectors-web/src/alchemyAuth.ts
jakehobbs: prob should use `??` here b/c of @dphilipson 's point about some funny person creating a chain w/ an id of `0`.
```suggestion
                  chainId: currentChainId ?? config.chains[0]?.id ?? 1,
```

jakehobbs: i also don't looooove the idea of defaulting to mainnet & would prob personally do `const chainId currentChainId ?? config.chains[0]?.id` above & assert it's not nullish.

packages/connectors-web/src/alchemyAuth.ts
jakehobbs: think it's worth creating a helper func for this since we do it in a couple of different places?

packages/auth/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export { AuthClient } from "./authClient.js";
 export { OauthCancelledError } from "./errors.js";
+export { isValidUser } from "./utils.js";
```
jakehobbs: do we actually want to expose this or is it just something used internally by this pkg?

examples/connectors-web-example/src/main.ts
jakehobbs: i think it's b/c the `connect` function should probably prio the persisted chainId when reconnecting, even if there's another `chainId` passed to it?

packages/connectors-web/src/alchemyAuth.ts
linnall: +1 here, the nested `try` blocks also make the code more difficult to read

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -71,36 +174,152 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
+    // Silent resume logic to try to restore auth session from storage
+    async function tryResume(): Promise<boolean> {
+      if (authSessionInstance) return true;
+
+      try {
+        const persisted = await getStoredAuthSession(config.storage);
+        if (!persisted) return false;
+
+        // Check if session is still valid (not expired)
+        if (Date.now() >= persisted.expirationDateMs) {
```
linnall: do we need to check expiration here if we do so in `authClient.restoreAuthSession`?

packages/connectors-web/src/alchemyAuth.ts
jakehobbs: (this is prob irrelevant after linna/dphil's comments)

packages/auth/src/index.ts
thebrianchen: this was used internally by `connectors-web`, but refactoring to avoid this since connectors-web should be agnostic of how authsession is stored

packages/auth/src/utils.ts
thebrianchen: deleted this file

packages/auth/src/utils.ts
thebrianchen: deleted this file

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: agree. we will need to a deeper dive into how we are handling expiration dates. will tackle this in a separate PR

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: started tracking the listener to disconnect

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: done

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: renamed to `version`

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: good suggestion, moved

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: N/A using straight type now

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: yeah, this is my bad. moving it over to string

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: N/A, just using direct zod object now

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: switched over.

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: moved

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: N/A anymore. this is the current flow?
  1. Get stored auth session
  2. Pass opaque state to restoreAuthSession()
  3. Auth package validates expiration internally
  4. If restoreAuthSession() returns undefined (expired/invalid), clean up storage
  5. If it returns a valid session, use it

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: removed this try/catch

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: removed and refactored

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: added `assertNotNullish` check

thebrianchen: not sure what the best behavior here is. this feels like a run-time issue where the storage is full/unavailable, but the auth session is technically still valid and usable in memory, so the user can still continue their current session. 

i did add a throw for no chain id though

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: yeah, agreed. changed to use authsessionstate

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: extracted into helper

thebrianchen: extracted into helper

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: i think it makes more sense to persist only in `switchChain()` since iiuc `onChainChanged()` gets called afterwards, which might cause double persistence. but i'm not sure how often the chain would switched from a call not originating from `switchChain()`. 

For now, going to implement persistence in switchChain() only. can add logic down the road to persist for `onChainChanged()` only if `switchChain()` wasn't called, but that feels unnecessarily complex for now.

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: removed

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: yeah n/a now

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: super helpful, thanks for flag! added in `connect`.

examples/connectors-web-example/src/main.ts
thebrianchen: added import

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: done

examples/connectors-web-example/src/main.ts
thebrianchen: oh what it was passing for me when i ran it, but thanks for catching and investigating!

thebrianchen: actually end up deleting the persistence tests since the strings are double stringified in json (once by authclient and once by wagmi), which makes parsing them a pain.

examples/connectors-web-example/src/main.ts
thebrianchen: i couldn't get it to reconnect and was hacking around it. adding `isReconnecting` makes things simpler though.

packages/connectors-web/src/store/authSessionStorage.ts
```diff
@@ -0,0 +1,108 @@
...
+  /** Version for migration handling */
+  version: 1;
+  /** Chain ID for connector-specific context */
+  chainId: number;
+  /** Opaque serialized auth session state - contents managed by auth package */
+  authSessionState: string;
+};
+
+// Storage configuration
+const STORAGE_KEY = "alchemyAuth.authSession" as const;
```
thebrianchen, line 15: extracting here for now knowing that's not really modular or reusable. punting on that until we have a clear use case for how to reuse storage

packages/connectors-web/src/alchemyAuth.ts
linnall: happy to discuss this. also wanted to callout that we have [an asana ticket](https://app.asana.com/1/1129441638109975/project/1210112085973163/task/1211126074565873?focus=true) for this

packages/connectors-web/src/alchemyAuth.ts
linnall: yes that is the current flow. to ensure this flow holds, we will need to ensure proper error handling when calling/within `loginWithPasskey`, `completeAuthWithBundle`, and any other method we use to create the restored `AuthSession` so we also return `undefined` when those methods fail

examples/connectors-web-example/src/main.ts
```diff
@@ -489,19 +504,91 @@ function setupSessionControls() {
...
+  setTimeout(async () => {
+    try {
+      const alchemyConnector = resolveAlchemyAuthConnector(config);
+      if (alchemyConnector && (await alchemyConnector.isAuthorized())) {
+        // If Alchemy connector is authorized, connect directly to it
+        await connect(config, { connector: alchemyConnector });
+      } else {
+        // Otherwise try standard reconnect for other connectors
+        await reconnect(config);
+      }
```
jakehobbs: i think we are going to have problems once we get to the react level if just calling `reconnect` doesn't work.

do you want to wait & worry about that when we get to the other connector since this one won't be directly used much?

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -197,9 +325,13 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
...
-      async isAuthorized() {
-        // Check if we have a valid authSession instance
-        return authSessionInstance !== undefined;
+      async isAuthorized(): Promise<boolean> {
+        if (authSessionInstance) {
+          return true;
+        }
+
+        const stored = await getStoredAuthSession(config.storage);
+        return !!stored;
```
jakehobbs, line 358: is it correct to say we're authorized just b/c we have a stored auth session? i would expect to not be "authorized" until the session is restored, but maybe my understanding of this isn't 100%.

examples/connectors-web-example/src/main.ts
jakehobbs: seems like calling the `reconnect` action never actually calls `connect()` on our connector. unsure why.

jakehobbs: dug into this a bit more. looks like wagmi calls `getProvider` (and `isAuthenticated`) on our connector before if actually calls `connect` on it.

jakehobbs: as per discussion offline, maybe we should kick off the `tryResume` in `setup`, then store the promise within the connector. then when a function is called that requires the session (i.e. `connect` or `getProvider`, it can await the promise if there's no auth session.

packages/connectors-web/src/alchemyAuth.ts
thebrianchen: this is incomplete. we'll have something for tracking expiration time later in #2106

PR #2106, opened 2025-09-24
feat: add expiration time ms tracking

thebrianchen:
Adding tracking for expiration in auth sessions with a default of 15mins. Updated example web app to show expiration time.

note that we still need to add functionality for letting user provide the expiration time, not included in this PR.

# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing session management and authentication in the `alchemyAuth` connector, including session persistence, expiration handling, and improved session state management.

### Detailed summary
- Added `DEFAULT_SESSION_EXPIRATION_MS` constant for default expiration time.
- Implemented session duration handling in `AuthClient` and `AuthSession`.
- Enhanced `alchemyAuth` to persist and restore session state across page reloads.
- Updated tests to verify session restoration and expiration behavior.
- Improved session status display in the UI.

> The following files were skipped due to too many changes: `.cursor/rules/system-prompt.mdc`, `v4_v5.md`


jakehobbs (approved):

packages/auth/src/authSession.ts
```diff
@@ -47,6 +52,8 @@ export type CreateAuthSessionParams = {
   authType: AuthType;
   /** Credential ID for passkey authentication */
   credentialId?: string;
+  /** Expiration timestamp - if not provided, defaults to 15 minutes from now */
+  expirationDateMs?: number;
```
dphilipson: The comment and variable name are a bit imprecise: they make it sound like you're providing the timestamp at which the session expires, rather than the length of the session. I'd consider naming the variable either just `expirationMs` or maybe something like `sessionDurationMs`, and adjusting the comment.

(I realize the variable name already existed in the code, but I didn't notice it before and now it's becoming part of the public API so needs more attention.)

Edit: Oh whoops, I realize that what I'm describing is in fact what the variable actually does, so the naming and comment are correct. In that case, I'll say that I think it would be more intuitive to have this take the duration instead, since that's what devs are thinking in terms of in 99% of situations.

dphilipson (commented):
This is great! Left a suggestion about naming.

I think we can do a little more in relation to expiration time, which can possibly go in a separate PR. We should also set a timer in the auth client which triggers when the session expires, disconnects the client, and emits a "disconnected" event.

I was thinking of something like the interface in the [design doc](https://www.notion.so/alchemotion/Signer-v5-Interface-Review-246069f2006680098221e0b20c8715d0?source=copy_link#246069f200668012b80cdbdffbc9b115), particularly the elements `disconnect()`, `on("disconnect", handler)`, and `isDisconnected`. As mentioned, it might make sense to do all this together in a separate PR.

packages/auth/src/authSession.ts
linnall: I agree with dphil here. My mental model was that the parameter passed into the `AuthSession` constructor would be called something like `sessionLengthMs` and it would be used to set the private element on `AuthSession` called `expirationDateMs`

jakehobbs: +1

packages/auth/src/authClient.ts
```diff
@@ -500,6 +500,7 @@ export class AuthClient {
       orgId,
       idToken,
       authType: type,
```
linnall, line 398: as mentioned on `authSession.ts`, can we rename this to something like `sessionLengthMs`? assuming that the user will be passing in their desired session length and we are responsible for then calculating the expiration date from it

packages/auth/src/authSession.ts
thebrianchen: changed to `sessionDurationMs`

packages/auth/src/authClient.ts
thebrianchen: added logic to recalculate session length based on the expiration date that's deserialized

PR #2111, opened 2025-09-26
fix: only run CI on affected packages

thebrianchen:
# Pull Request Checklist

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -102,10 +102,14 @@ jobs:
         run: yarn build:libs
 
       - name: Unit Test
-        run: yarn test:ci
+        run: |
+          # Run tests only for packages that changed since the base branch
+          yarn turbo test:ci --filter='[origin/main...HEAD]'
         
       - name: Typecheck Test
-        run: yarn test:typecheck
+        run: |
+          # Run typecheck only for packages that changed since the base branch
+          yarn turbo test:typecheck --filter='[origin/main...HEAD]'
     
   build_ios:
     name: Build iOS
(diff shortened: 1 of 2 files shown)


.github/workflows/on-pull-request.yml
```diff
@@ -102,37 +102,42 @@ jobs:
         run: yarn build:libs
 
       - name: Unit Test
-        run: yarn test:ci
+        run: |
+          # Run tests for packages that changed and their dependents
+          BASE_BRANCH="${{ github.event.pull_request.base.ref }}"
+          yarn turbo test:run --filter="./packages/*" --filter="[origin/$BASE_BRANCH...HEAD]" --continue
```
jakehobbs: ❗ i believe this will cause turbo to start multiple instances of vitest (one per pkg), which will lead to the global setup running multiple times. this is most likely what's causing the tests to fail.

thebrianchen: great catch, using a bit of a hacky method right now to pull the changed packages with turbo, but then run the regular tests with `vitest`

scripts/run-affected-tests.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Use turbo dry-run to identify which packages need testing
+# This includes both changed packages and their dependents
+TURBO_OUTPUT=$(yarn turbo run test:run --filter="./packages/*" --filter="[origin/$BASE_BRANCH...HEAD]" --dry-run 2>&1)
+
+# Extract directories of packages that have test:run tasks
+# 1. Find lines like "@alchemy/package-name#test:run"
+# 2. Look ahead to find the Directory line
+# 3. Extract the directory path and remove the "packages/" prefix
+AFFECTED_DIRS=$(echo "$TURBO_OUTPUT" | \
```
jakehobbs, line 18: this feels like it might be a little brittle (i.e. if turbo changes the log output). fine w/ it for now, but not sure if it's worth revisiting later.

jakehobbs (approved):

scripts/run-affected-tests.sh
thebrianchen: yeah agreed. will keep an eye out. worst case, we just revert to what we had before with the standard test run suite.

PR #2115, opened 2025-09-26
feat(connectors-web): add alchemySmartWallet connector

dphilipson:
Basic implementation of a connector which uses the Wallet API to perform actions using an Alchemy smart account, wrapping another connector to perform signing.

This version is tested to work with the Metamask connector, although there is additional work that can done to clean up code and polish edge cases, as marked with TODO's.

# Pull Request Checklist

---

## PR-Codex overview
This PR primarily focuses on enhancing the `@alchemy/connectors-web` library by introducing a new `alchemySmartWallet` connector, updating existing connectors, and refining the integration of Alchemy's Wallet API functionalities.

### Detailed summary
- Deleted `resolveConnector.ts` and `test-smart-wallet.ts` files.
- Added `@alchemy/wagmi-core` and `@alchemy/wallet-apis` dependencies.
- Introduced `alchemySmartWallet` connector in `alchemySmartWallet.ts`.
- Updated `resolveAuthConnector` to return `connectAlchemyAuth`.
- Refined `sendEmailOtp`, `submitOtpCode`, and `handleOauthRedirect` functions to use destructured connector.
- Added new documentation for `createSmartWalletClientAndRequestAccount` and `createEip1193Provider`.
- Updated example configurations to utilize the new `alchemySmartWallet`.
- Enhanced error handling and logging in various functions.
- Improved type definitions and interfaces for better type safety.


packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,233 @@
...
+    // `config` contains an event emitter that all of Wagmi listens to. Don't
+    // let the inner connecter get its hands on it: we want to fully encapsulate
+    // it and not let it affect global state. Note that the uid doesn't matter
+    // since we're only using this internally.
+    //
+    // TODO: Listen for inner connect and disconnect events and update state accordingly.
+    const innerEmitter = new Emitter("inner-uid");
+    const innerConnector = options.baseConnector({
+      ...config,
+      emitter: innerEmitter,
```
blakecduncan, line 77: Nice, good idea

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,233 @@
...
+     *
+     * TODO: Should we return the same provider instance every time, with that
+     * instance being a proxy object routing to either this or the real provider
+     * if it exists? We would need to do that if Wagmi held on to a provider
+     * that it first got during initialization. It doesn't seem to do that
+     * today, but what if it changed behavior?
+     */
+    const loggedOutProvider: Provider = {
+      on: () => {},
+      removeListener: () => {},
```
blakecduncan, line 93: Not sure if this is overkill/needed, do you think we should have some code to remove the listeners?

```
const listeners = new Map();
const loggedOutProvider: Provider = {
  on: (event, listener) => {
    if (!listeners.has(event)) listeners.set(event, new Set());
    listeners.get(event)?.add(listener);
    return () => listeners.get(event)?.delete(listener);
  },
  removeListener: (event, listener) => {
    listeners.get(event)?.delete(listener);
  },
};
```

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,233 @@
...
-        void chainId;
-        throw new Error("Not implemented: switchChain");
+      async switchChain(params): Promise<Chain> {
+        const { chainId } = params;
+        const currentChainId = await this.getChainId();
+        if (chainId === currentChainId) {
+          return getChainFromConfig(config, chainId);
+        }
+        resetState();
+        const baseChain = await innerConnector.switchChain?.(params);
```
blakecduncan, line 314: Maybe not an issue but I wonder if it's worth reseting the state after successfully switching chains?

If the innerconnector switch chan is rejected then we will be stuck in a state where the `smartWalletClientPromise` is undefined.

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,233 @@
...
+      smartWalletClient.account = {
+        type: "json-rpc",
+        address: account.address,
+      };
+      return smartWalletClient;
+    };
+
+    const initializeSmartWalletClient = async (): Promise<void> => {
+      smartWalletClientPromise = createClient();
+      await smartWalletClientPromise;
```
blakecduncan: I think on every page load, setup is called and then connect is called right after by wagmi.  I think we might be calling createClient twice because of that.  The setup will initialize it and then the connect will reset it and initialize it again.

In the connect function, if we are already connected, I wonder if we can skip this part

```
resetState();
await innerConnector.connect(params);
```

then check if the chainId is correct and if not update it.

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,233 @@
...
-      async isAuthorized() {
-        throw new Error("Not implemented: isAuthorized");
+      async isAuthorized(): Promise<boolean> {
+        return innerConnector.isAuthorized();
       },
 
-      async switchChain({ chainId }) {
-        void chainId;
-        throw new Error("Not implemented: switchChain");
+      async switchChain(params): Promise<Chain> {
```
blakecduncan, line 312: Haven't had a chance to debug it but I noticed in the demo that when I switch chains when connected with the Alchemy Smart Wallet that this function is called but the new chain ID isn't reflected in the UI

blakecduncan (approved):
This is awesome, tested it out in the demo and it's working well!

Few comments but happy for any changes to be made in subsequent PRs if you wanna merge this

packages/connectors-web/src/alchemySmartWallet.ts
dphilipson: Since right now this provider never emits events, what I wrote should be equivalent behavior and I think there's no reason to store listeners that won't be called.

(We could forward listeners to the inner provider, but that feels like it could have unpredictable consequences.)

packages/connectors-web/src/alchemySmartWallet.ts
dphilipson: You're right, I'll reset after successfully swapping the inner I think.

packages/connectors-web/src/alchemySmartWallet.ts
dphilipson: That's not what happened in my testing: `connect` was never called on page load, even on reloading into an already-connected state. `connect` would only be called on the user interaction that requested to connect to this provider, at least in the wagmi demo app you get from their template.

Although it's entirely there could be a different setup that does call connect on load, let me think about this one.

packages/connectors-web/src/alchemySmartWallet.ts
dphilipson: Ooh good catch, this was working in some version of my code but looks like it broke. Let me figure out what's wrong.

dphilipson: Oh I know what happened: it's the isolated event emitter. Previously, the inner connector was publishing its chain change events to the global state, which is wrong but kinda gave the behavior we wanted. Let me fix this.

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,233 @@
...
 
 export type AlchemySmartWalletOptions = {
-  createBaseConnector: CreateConnectorFn;
+  /** The base connector which will sign for the smart wallet. */
+  baseConnector: CreateConnectorFn;
+  /** API key for Alchemy authentication */
+  apiKey?: string;
+  /** JWT token for authentication */
+  jwt?: string;
+  /** Custom RPC URL (optional - defaults to Alchemy's API URL, but can be used to override the chain's Alchemy URL) */
```
thebrianchen: nit: not necessarily RPC url right? taking a step back though, how does the URL work for rpc vs rest calls? doesn't this connector need to do both?

packages/connectors-web/src/alchemySmartWallet.ts
thebrianchen: not sure if this is applicable here, but in `alchemyAuth` we used a `resumePromise` ([link](https://github.com/alchemyplatform/aa-sdk/blob/bc/ci-filter/packages/connectors-web/src/alchemyAuth.ts#L78)) as a standard promise that we wait for on reload. 

does the `getProvider()` only get called on reload, which is the case that the logged out provider is supposed to be for?

packages/connectors-web/src/alchemySmartWallet.ts
blakecduncan: Ohh yeah that makes sense.  Nice find

packages/connectors-web/src/alchemySmartWallet.ts
blakecduncan: Dang that's strange - I'm currently using MM to connect in the `connectors-web-example`.  I don't think it's a huge deal though.  If I get some time in a little I'll see if I can inspect why connect is getting called for me

packages/connectors-web/src/alchemySmartWallet.ts
dphilipson: This connector is only making JSON-RPC calls to the Wallet API. It doesn't make any rest calls or eth method calls.

You're highlighting a real problem I'm worried about across the SDK re: confusion around the different transports and urls needed. I want to keep an eye on this and see if we can make it clearer somehow.

packages/connectors-web/src/alchemySmartWallet.ts
dphilipson: Oh that's really interesting, I assumed that wagmi wouldn't accept us until we produced a provider, but it seems like just a promise of a provider is good enough for wagmi. Let me try that, and also see what built-in connectors are doing.

examples/connectors-web-example/src/wagmi.ts
```diff
@@ -1,16 +1,17 @@
...
-  apiKey: import.meta.env.VITE_ALCHEMY_API_KEY,
-});
+const apiKey = import.meta.env.VITE_ALCHEMY_API_KEY;
+
+const transport = alchemyTransport({ apiKey });
 
 export const config: Config = createConfig({
   chains: [arbitrumSepolia, mainnet, sepolia],
   connectors: [
+    alchemySmartWallet({ baseConnector: injected(), apiKey }),
```
jakehobbs: curious if you thought much about this `baseConnector` name. def won't block on it, but i'm not sure if it's the most clear. i know we are trying to avoid the word "signer" too though. maybe "owner" would be more clear than "base"?

packages/connectors-web/src/alchemySmartWallet.ts
jakehobbs: yeah, this is smart.

packages/connectors-web/src/alchemySmartWallet.ts
jakehobbs: `reconnect` awaits the provider: https://github.com/wevm/wagmi/blob/9651e18c3d140e37e16dcc80aabc29624344b7c5/packages/core/src/actions/reconnect.ts#L68

jakehobbs: the only time we would need the dummy provider during this implementation is during `reconnect`, right?

when `getProvider` is called here, could we await the inner connector being ready, then await the smart wallet client creation so that we have the provider here?

packages/connectors-web/src/alchemySmartWallet.ts
jakehobbs: hmm, wagmi's `reconnect` action should definitely be calling `connect`: https://github.com/wevm/wagmi/blob/9651e18c3d140e37e16dcc80aabc29624344b7c5/packages/core/src/actions/reconnect.ts#L81

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,233 @@
...
+        await innerConnector.setup?.();
+        if (await innerConnector.isAuthorized()) {
+          await initializeSmartWalletClient();
+        }
       },
 
-      async connect(parameters) {
-        await baseConnector.connect(parameters);
-        throw new Error("Not implemented: connect");
+      async connect(params): Promise<{ accounts: Address[]; chainId: number }> {
```
jakehobbs: not sure if this would change anything for your implementation, but just want to flag that `connect` has an `{ isReconnecting: true }` param whenever it's called via wagmi's `reconnect` action.

packages/connectors-web/src/alchemySmartWallet.ts
jakehobbs: +1

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,233 @@
...
+  type Chain,
+  type EIP1193Provider,
+  type ProviderConnectInfo,
+} from "viem";
 import { createConnector, type CreateConnectorFn } from "wagmi";
 
 export type AlchemySmartWalletOptions = {
-  createBaseConnector: CreateConnectorFn;
+  /** The base connector which will sign for the smart wallet. */
+  baseConnector: CreateConnectorFn;
```
jakehobbs: could we go ahead and add a `paymasterPolicyIds` param to the `AlchemySmartWalletOptions` now too? would be good for testing sending calls, and should be a super simple pass-through to the `smartWalletClient`.

side note: we have some type issues w/ setting call-specific policyId: https://alchemyinsights.slack.com/archives/C094T1JRHDH/p1758924853918919, but we can circle back on that later imo.

jakehobbs (commented):
left a couple of questions. good work so far! cool to see this coming along.

could i request that we update the "send call via smart wallet client" button in the connector-web-example to just be "send call" and use the wagmi action to send a call via the connector?

examples/connectors-web-example/src/wagmi.ts
dphilipson: Owner is a good idea, changed.

packages/connectors-web/src/alchemySmartWallet.ts
dphilipson: Thinking about it more, I'm not sure I like the idea of returning a suspended promise that resolves once the user logs in. That seems like it could be unexpected behavior for callers, who might call `await connector.getProvider()` and then be surprised that their program hangs forever. I think the expected behavior would be to return a provider that reflects the logged out state. What do you think?

packages/connectors-web/src/alchemySmartWallet.ts
dphilipson: After some refactoring I did see connect getting called on load. Not sure why it didn't come up before. In any case, this is resolved now because our `setup` no longer initializes the smart wallet client, only `connect` does.

packages/wagmi-core/src/utils/resolveAuthConnector.ts
```diff
@@ -0,0 +1,54 @@
...
+ * the Alchemy Smart Wallet connector if wrapped.
+ *
+ * @param {Config} config - The Wagmi config containing connectors
+ * @returns {ResolvedAuthConnector} The Alchemy Auth connector and a function to connect to it
+ * @throws {Error} If the connector is not found
+ */
+export function resolveAlchemyAuthConnector(
+  config: Config,
+): ResolvedAuthConnector {
+  console.log("CONFIG", config);
```
blakecduncan: nit but just noting these couple debugging log statements

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,332 @@
...
+  accounts?: Address[];
+};
+
+/**
+ * Creates a Wagmi connector that uses Alchemy's Wallet API to perform actions
+ * using a smart account. It wraps a base connector and uses it to sign
+ * transactions.
+ *
+ * @param {AlchemySmartWalletOptions} options - The options for the Alchemy Smart Wallet connector.
+ * @param {CreateConnectorFn} options.baseConnector - The base connector to use to sign transactions.
```
blakecduncan: looks like this was renamed below to options.ownerConnector

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,332 @@
...
+    };
+
+    const createClient = async (): Promise<SmartWalletClient> => {
+      const chainId = await innerConnector.getChainId();
+      const innerAccounts = await innerConnector.getAccounts();
+      if (innerAccounts.length === 0) {
+        throw new Error("No accounts found in base connector");
+      }
+      // TODO: do we always take the first account?
+      const signerAddress = innerAccounts[0];
```
blakecduncan, line 131: Yeah I think it's a pretty fair assumption that the first account in the list is the currently active one.  I don't think the spec for eth_accounts specifies an ordering of the accounts but I do believe most wallets behave this way

packages/wagmi-core/package.json
```diff
@@ -49,7 +49,6 @@
   },
   "dependencies": {
     "@alchemy/common": "*",
-    "@alchemy/connectors-web": "*",
```
blakecduncan, line 52: 🙏

packages/wagmi-core/src/utils/resolveAuthConnector.ts
dphilipson: oh thank you, I thought I got all of those 😞

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,333 @@
...
+        transport: alchemyTransport({
+          apiKey: options.apiKey,
+          jwt: options.jwt,
+          url: options.url ?? "https://api.g.alchemy.com/v2",
+        }),
+        chain,
+        policyId: options.policyId,
+        policyIds: options.policyIds,
+      });
+      // TODO: Calling `client.getProvider()` has the side effect of
```
jakehobbs: we should probably remove this from the provider soon. this is also kinda preventing us from specifying account options (i.e. 7702) here.

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,333 @@
...
+      }
+      // TODO: do we always take the first account?
+      const signerAddress = innerAccounts[0];
+      const chain = getChainFromConfig(config, chainId);
+      const provider = (await innerConnector.getProvider()) as EIP1193Provider;
+      const signer = createWalletClient({
+        account: signerAddress,
+        chain,
+        transport: custom(provider),
+      });
```
jakehobbs, line 137: ❗ if `innerConnector.getClient` is defined, we should use that to get the signer client. using `innerClient.getProvider()` and wrapping it in a new client will not allow signing 7702 authorizations.

jakehobbs (commented):
left a couple of comments, connector works perfectly (when not using 7702) from my testing! 🚀 

the "test send calls" button in the connector web example is no longer working though. would you mind updating that to simply use the `sendCalls` wagmi action instead of the workaround we had before? i put up a pr here if you just want to merge it into yours: https://github.com/alchemyplatform/aa-sdk/pull/2152

packages/connectors-web/src/alchemySmartWallet.ts
dphilipson: Oh you're right. Let me think about what to do with that.

packages/connectors-web/src/alchemySmartWallet.ts
dphilipson: Oh really good idea, I didn't know that was an option.

packages/connectors-web/src/alchemySmartWallet.ts
```diff
@@ -1,85 +1,333 @@
...
+        address: account.address,
+      };
+      return smartWalletClient;
+    };
+
+    const initializeSmartWalletClient = async (): Promise<void> => {
+      smartWalletClientPromise = createClient();
+      const smartWalletClient = await smartWalletClientPromise;
+      currentChainId = smartWalletClient.chain.id;
+      currentInnerAccount = (await innerConnector.getAccounts())[0];
```
linnall: won't this throw an indexing error if `getAccounts` returns an empty array or is that case impossible?

dphilipson: Oh good call. Normally this should always have an account in it since we won't try to initialize without a connected inner connector, but it's very possible that due to race conditions it won't have an account by the time we get here. Let me change that.

jakehobbs (approved):
tested w/ connecting, disconnecting, switching chains, signing, and sending calls using connector web example & v5 react example (w/ 7702 and sma-b). everything is working perfectly! 🥳

PR #2124, opened 2025-09-26
fix: immediately opens popup then set resolved auth url to avoid popup blocking

linnall:
# Pull Request Checklist


packages/auth-web/src/createWebAuthClient.ts
```diff
@@ -100,15 +101,37 @@ export function createWebAuthClient({
...
-          // const eventEmitter = this.eventEmitter;
+          if (!popup) {
+            throw new Error(
+              "Popup blocked by browser. Please allow popups for this site.",
+            );
+          }
+          // Wait for auth URL to be ready, then set url in popup
+          const finalAuthUrl = await (typeof authUrl === "string"
+            ? authUrl
+            : authUrl);
```
jakehobbs: nit: the type of this should always be a promise due to how this function is typed.
```suggestion
          const finalAuthUrl = await authUrl;
```

jakehobbs: also, even if the type were not narrowed yet, it's totally fine to call `await` on an `any | Promise<any>`. awaiting a non-promise just returns the value immediately.

packages/auth-web/src/createWebAuthClient.ts
```diff
@@ -100,15 +101,37 @@ export function createWebAuthClient({
...
+    ) => {
+      const { authUrl, mode } = params;
       switch (mode) {
         case "popup":
+          // Open popup immediately to avoid popup blockers
           const popup = window.open(
-            authUrl,
-            "_blank",
-            "popup,width=500,height=600",
+            "about:blank",
```
jakehobbs, line 113: happy w/ this for now, but i wonder if we should show some sort of loading page eventually instead of just an empty window.

packages/auth-web/src/createWebAuthClient.ts
```diff
@@ -100,15 +101,37 @@ export function createWebAuthClient({
...
+            "oauth-popup",
+            "width=500,height=600",
           );
-          // const eventEmitter = this.eventEmitter;
+          if (!popup) {
+            throw new Error(
+              "Popup blocked by browser. Please allow popups for this site.",
+            );
+          }
```
jakehobbs, line 126: i would suggest moving this check to after we await the promise, to be a bit more defensive against the user closing the popup before the promise resolves.

packages/auth-web/src/createWebAuthClient.ts
jakehobbs: actually it might be super easy to just do something like this:
```ts
const doc = popup.document;

if (doc.body) {
  // body already exists → just replace contents
  doc.body.textContent = "Loading sign-in…";
} else {
  // no body yet → create one and append
  const body = doc.createElement("body");
  const p = doc.createElement("p");
  p.textContent = "Loading sign-in…";
  p.style.cssText = "font:14px system-ui; margin:16px;";
  body.appendChild(p);
  doc.appendChild(body);
}
```

packages/auth-web/src/createWebAuthClient.ts
```diff
@@ -161,14 +184,13 @@ export function createWebAuthClient({
...
                   reject(new Error(`Unknown status: ${status}`));
               }
-            }; // handleMessage
+            };
 
             window.addEventListener("message", handleMessage);
 
             const checkCloseIntervalId = setInterval(() => {
               if (popup?.closed) {
                 cleanup();
```
thebrianchen, line 187: don't we need to reject the promise here?

packages/auth-web/src/createWebAuthClient.ts
thebrianchen: i think this can be removed now

thebrianchen (commented):
don't know enough to stamp this but left some drive-by comments. nasty bug to run into though, thanks for taking it on

packages/auth-web/src/createWebAuthClient.ts
linnall: you're right, `isSignup` was added to help with handling the case that a user logs in with a new oauth account and indicated that they would like to add a passkey after sign up. will comment out this query param for now and add a TO DO comment for how it should eventually be used

packages/auth-web/src/createWebAuthClient.ts
linnall: from testing locally, rejecting here throws an error. this is unwanted behavior since it is valid for the user to close the popup

packages/auth-web/src/createWebAuthClient.ts
linnall: out of curiosity, in what cases would the body already exist?

jakehobbs: the body existed when i was testing (in chrome). i'm not sure, but chatgpt suggested checking if it exists or not, so maybe some browsers behave differently.

jakehobbs: looks like safari/chrome/firefox at least all already create the body in the blank window.

jakehobbs: looks like safari/chrome/firefox at least all already create the body in the blank window.

linnall: added this simple loading page in the popup
Screenshot 2025-09-29 at 3 58 46 PM

packages/auth-web/src/createWebAuthClient.ts
```diff
@@ -100,15 +101,39 @@ export function createWebAuthClient({
...
+      const { authUrl, mode } = params;
       switch (mode) {
         case "popup":
+          // Open popup immediately to avoid popup blockers
           const popup = window.open(
-            authUrl,
-            "_blank",
-            "popup,width=500,height=600",
+            "about:blank",
+            "oauth-popup",
```
jakehobbs, line 114: interesting, TIL about this custom target name feature of `window.open` 👍

packages/auth/src/authClient.ts
```diff
@@ -245,32 +248,21 @@ export class AuthClient {
...
+          mode: params.mode,
+        });
+        return await this.processOAuthResponse(redirectResponse);
+      default: // popup
+        const authUrlPromise = this.buildAuthUrl(params);
+        const popupResponse = await this.handleOauthFlow({
+          authUrl: authUrlPromise,
+          mode: params.mode,
+        });
+        return await this.processOAuthResponse(popupResponse);
```
jakehobbs: nit: might be a little simpler to just do this instead of repeating everything in the 2 different branches here. just a suggestion though & i'm fine w/ it as-is if you prefer.

```suggestion
        const authUrlPromise = this.buildAuthUrl(params);
        const redirectResponse = await this.handleOauthFlow({
          authUrl: params.mode === "popup" ? authUrlPromise : await authUrlPromise,
          mode: params.mode,
        });
        return await this.processOAuthResponse(redirectResponse);
```

packages/auth/src/authClient.ts
```diff
@@ -459,6 +451,40 @@ export class AuthClient {
...
+  private async buildAuthUrl(params: LoginWithOauthParams): Promise<string> {
+    const { targetPublicKey } = await this.getTekStamper();
+    const oauthConfig = await this.signerHttpClient.request({
+      route: "signer/v1/prepare-oauth",
+      method: "POST",
+      body: { nonce: getOauthNonce(targetPublicKey) },
+    });
+    return getOauthProviderUrl({
+      oauthParams:
+        params.mode === "redirect" ? { ...params, redirectUrl: "/" } : params, // TO DO: clean up redirect vs popup path
```
jakehobbs, line 454: i think this is totally unrelated to this PR and was already a thing, but i'm curious to know why the `redirectUrl` is always `"/"`, if i'm reading this correctly?

jakehobbs (approved):

packages/auth/src/authClient.ts
linnall: for the redirect case that is correct. this makes it so that the auth provider's login page opens on the current tab that the user is on

packages/auth/src/authClient.ts
linnall: this is way cleaner, I like it

packages/auth-web/src/createWebAuthClient.ts
thebrianchen: ahh that makes sense

jakehobbs (approved):

packages/auth-web/src/createWebAuthClient.ts
linnall: update: copilot also says we should reject the promise to prevent leaving an auth flow hanging. throwing this error does not block the user from attempting to login again so will reinstate this for now

PR #2125, opened 2025-09-26
feat: add createAlchemyConfig and AlchemyProvider

blakecduncan:
# Pull Request Checklist


packages/react/src/createAlchemyConfig.ts
```diff
@@ -0,0 +1,135 @@
...
+    return perChain;
+  })();
+
+  // Note: Currently using alchemyAuth; swap to alchemySmartWallet when ready.
+  const defaultConnectors = [
+    // Currently alchemyAuth only supports apiKey
+    ...(apiKey ? [alchemyAuth({ apiKey })] : []),
+  ];
+
+  const wagmiConfig = createConfig({
```
blakecduncan, line 122: I ended up not implementing the `wagmiOverrides` option in this PR.  I wanted to get some opinions on it.  I struggled getting the types to work for some reason, I kept getting a type error when doing a shallow copy into the params.

It got me thinking, I wonder if we should just try to expose options that we see fit (rather than a full `wagmiOverrides` option.  And if anyone needs a bunch of wagmi overrides, we just recommend they build the wagmi config themselves.

yarn.lock
```diff
@@ -77,6 +77,64 @@
   resolved "https://registry.yarnpkg.com/@adraffy/ens-normalize/-/ens-normalize-1.11.0.tgz#42cc67c5baa407ac25059fcd7d405cc5ecdb0c33"
   integrity sha512-/3DDPKHqqIqxUULp8yP4zODUY1i+2xvVWsv8A79xGWdCAG+8sb0hRh0Rk2QyOJUnnbyPUAZYcpBuRe3nS2OIUg==
 
+"@alchemy/aa-accounts@3.14.1":
```
jakehobbs, line 216: do you need this many lockfile changes?

examples/react-v5-example/src/app/config.ts
```diff
@@ -1,21 +1,7 @@
...
-import { alchemyTransport } from "@alchemy/common";
+import { createAlchemyConfig } from "@alchemy/react";
 
-export const config = createConfig({
-  connectors: [
-    alchemyAuth({
-      apiKey: process.env.NEXT_PUBLIC_ALCHEMY_API_KEY,
-    }),
-  ],
+export const config = createAlchemyConfig({
```
jakehobbs, line 4: beautiful

packages/react/src/AlchemyProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+ *       {children}
+ *     </AlchemyProvider>
+ *   );
+ * }
+ * ```
+ */
+export function AlchemyProvider({ config, children }: AlchemyProviderProps) {
+  return (
+    <WagmiProvider config={config.wagmi}>
+      {children as any}
```
jakehobbs: curious why you have to cast to `any` here. i think i've maybe seen this before when there were conflicts w/ different library versions?

packages/react/src/createAlchemyConfig.ts
```diff
@@ -0,0 +1,135 @@
...
+      ...(url && { url }),
+    };
+    const perChain: NonNullable<CreateConfigParameters["transports"]> = {};
+    for (const chain of chains) {
+      perChain[chain.id] = alchemyTransport(baseInit);
+    }
+    return perChain;
+  })();
+
+  // Note: Currently using alchemyAuth; swap to alchemySmartWallet when ready.
```
jakehobbs: nit so we can more easily find this.
```suggestion
  // TODO(v5): Currently using alchemyAuth; swap to alchemySmartWallet when ready.
```

packages/react/src/createAlchemyConfig.ts
```diff
@@ -0,0 +1,135 @@
...
+  const defaultConnectors = [
+    // Currently alchemyAuth only supports apiKey
+    ...(apiKey ? [alchemyAuth({ apiKey })] : []),
+  ];
+
+  const wagmiConfig = createConfig({
+    chains,
+    transports: transports!,
```
jakehobbs: nit: does typing `transports`  as `NonNullable<CreateConfigParameters["transports"]>` above avoid this non-null assertion?

blakecduncan: Turns out I actually don't need the `!` at all.  There was a previous version of this that I needed and didn't refactor to remove the `!`.

packages/react/src/AlchemyProvider.tsx
blakecduncan: I spent a good amount of time trying to figure this out on Friday - I was able to remove the `any` after closing and opening cursor.  I'm guessing cursor hadn't indexed the types or something ☹️

yarn.lock
blakecduncan: I added `@alchemy/connectors-web` as a dependency in the `@alchemy/react` package and yarn updated all these dependencies.

I typically trust the lock updates when I add a package, is there a different way I should be adding packages?  right now I just did 

`yarn add @alchemy/connectors-web`

packages/react/src/AlchemyProvider.tsx
jakehobbs: oof

jakehobbs (approved):

yarn.lock
jakehobbs: gotcha, should be fine then! was just a little surprised that would change so much.

thebrianchen: yeah strange, i ran it locally too and verified this causes a bunch of changes

thebrianchen (approved):
this is beautiful. loving the devEx here

PR #2153, opened 2025-10-03
refactor: improve smart wallet 1193 provider

jakehobbs:
# Pull Request Checklist

---

## PR-Codex overview
This PR introduces a new function `createSmartWalletClientAndRequestAccount` to streamline the creation of a smart wallet client while requesting an account. It also updates the `createEip1193Provider` function to utilize this new function for better account handling.

### Detailed summary
- Added `createSmartWalletClientAndRequestAccount` function to combine client creation with account initialization.
- Updated `createSmartWalletClient` to export the new function.
- Modified `createEip1193Provider` to use `createSmartWalletClientAndRequestAccount`.
- Removed `getProvider` method from `createSmartWalletClient`.
- Adjusted E2E tests to use `createEip1193Provider` instead of `createSmartWalletClient`.

### DIFF
--- a/packages/wallet-apis/src/client.ts
+++ b/packages/wallet-apis/src/client.ts
@@ -9,7 +9,6 @@ import { smartWalletActions } from "./decorators/smartWalletActions.js";
 import { type AlchemyTransport } from "@alchemy/common";
 import type { SignerClient, SmartWalletClient } from "./types.js";
 import { createInternalState } from "./internal.js";
-import { createEip1193ProviderFromClient } from "./provider.js";
 
 export type CreateSmartWalletClientParams<
   TAccount extends Address | undefined = Address | undefined,
@@ -69,8 +68,5 @@ export const createSmartWalletClient = <
       policyIds: _policyIds,
       internal: createInternalState(),
     }))
-    .extend(smartWalletActions<TAccount>(signerClient))
-    .extend((_client) => ({
-      getProvider: () => createEip1193ProviderFromClient<TAccount>(_client),
-    }));
+    .extend(smartWalletActions<TAccount>(signerClient));
 };
--- a/packages/wallet-apis/src/client.ts
+++ b/packages/wallet-apis/src/client.ts
@@ -9,6 +9,7 @@ import { smartWalletActions } from "./decorators/smartWalletActions.js";
 import { type AlchemyTransport } from "@alchemy/common";
 import type { SignerClient, SmartWalletClient } from "./types.js";
 import { createInternalState } from "./internal.js";
+import { CreationOptions } from "@alchemy/wallet-api-types";
 
 export type CreateSmartWalletClientParams<
   TAccount extends Address | undefined = Address | undefined,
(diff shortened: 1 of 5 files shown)


packages/wallet-apis/src/provider.ts
```diff
@@ -37,30 +40,39 @@ export type SmartWalletClient1193Methods = [
...
-    const account = await client.requestAccount();
-    client.account = {
-      type: "json-rpc",
-      address: account.address,
-    };
-    eventEmitter.emit("connect", { chainId: toHex(client.chain.id) });
-  })();
-
   const request = (async ({ method, params }) => {
+    if (!client) {
```
dphilipson: Any harm with a race condition if `request` gets called more than once before the client is initialized? If clients are stateless and `requestAccount` returns the same thing every time, it should be fine that we create multiple clients and just stick with the last one, although it does make some extra `requestAccount` calls.

packages/wallet-apis/src/provider.ts
```diff
@@ -37,30 +40,39 @@ export type SmartWalletClient1193Methods = [
...
 
-export const createEip1193ProviderFromClient = <
-  TAccount extends Address | undefined = Address | undefined,
->(
-  client: BaseWalletClient<SmartWalletActions<TAccount>>,
+export const createEip1193Provider = (
+  clientParams: CreateSmartWalletClientParams<undefined>,
+  accountOptions: CreationOptions | { accountAddress: Address } = {},
 ) => {
+  const unauthedClient = createSmartWalletClient(clientParams);
```
dphilipson: (Nit) Since the unauthed client is only used once during initializing the real client, I'd consider declaring it inside the initialization branch of `request` so that it doesn't stick around, for the sake of code organization and cleaner autocomplete (and very minor memory save from garbage collection).

packages/wallet-apis/src/provider.ts
```diff
@@ -37,30 +40,39 @@ export type SmartWalletClient1193Methods = [
...
+              accountAddress: accountOptions.accountAddress,
+            }
+          : {
+              creationHint: accountOptions,
+            },
+      );
+      client = createSmartWalletClient({
+        ...clientParams,
+        account: account.address,
+      });
```
dphilipson: I wonder if it would make sense to factor this into a separate async function which performs this initialization (create unauthed client, request account, create authed client) and returns the authed client client. That seems like something that would be potentially useful to consumers, who want a logged in client but don't need a 1193 provider.

dphilipson (approved):
This does make sense to me! I left a few thoughts inline.

packages/wallet-apis/src/provider.ts
jakehobbs: good flag

packages/wallet-apis/src/provider.ts
jakehobbs: make sense. i originally had it outside because i was going to use it for other things, but that ended up not being the case.

packages/wallet-apis/src/provider.ts
jakehobbs: after doing that, do you think we should extend the client w/ `.getProvider()` again? or keep it totally separate?

wondering what makes the most sense from your perspective at the connector.

packages/wallet-apis/src/client.ts
```diff
@@ -9,6 +9,7 @@ import { smartWalletActions } from "./decorators/smartWalletActions.js";
 import { type AlchemyTransport } from "@alchemy/common";
 import type { SignerClient, SmartWalletClient } from "./types.js";
 import { createInternalState } from "./internal.js";
+import { CreationOptions } from "@alchemy/wallet-api-types";
```
jakehobbs: ❗ ❗ ❗ ❗ ❗ ❗ should be a type import

packages/wallet-apis/src/client.ts
```diff
@@ -70,3 +71,33 @@ export const createSmartWalletClient = <
...
+
+/**
+ * Creates a smart wallet client and requests an account in a single operation.
+ * This is a convenience function that combines client creation with account initialization.
+ *
+ * @param {CreateSmartWalletClientParams<undefined>} clientParams - Parameters for creating the smart wallet client (without an account).
+ * @param {CreationOptions | { accountAddress: Address }} accountOptions - Options for requesting the account. Can either be creation options for a new account or an object with an existing account address.
+ * @returns {Promise<SmartWalletClient<Address>>} A promise that resolves to a smart wallet client with an initialized account.
+ */
+export const createSmartWalletClientAndRequestAccount = async (
```
jakehobbs, line 83: not thrilled w/ this name, but idk what's better

packages/wallet-apis/src/provider.ts
jakehobbs: i think i prefer to keep it totally separate

packages/wallet-apis/src/provider.ts
```diff
@@ -37,30 +40,29 @@ export type SmartWalletClient1193Methods = [
...
-    const account = await client.requestAccount();
-    client.account = {
-      type: "json-rpc",
-      address: account.address,
-    };
-    eventEmitter.emit("connect", { chainId: toHex(client.chain.id) });
-  })();
-
   const request = (async ({ method, params }) => {
+    client ??= await (async () => {
```
dphilipson, line 53: Does this work? `client` doesn't get set until the `await` completes, so I think multiple requests will still all see the client as absent and try to initialize it.

dphilipson: Gonna merge now so I can try out the new provider in my PR. Will update this there.

jakehobbs: ah you're right

jakehobbs: thanks

PR #2154, opened 2025-10-03
feat: adds an sdk to seamlessly integrate privy with Alchemy's infra

blakecduncan:
# Pull Request Checklist

---

## PR-Codex overview
This PR introduces the `@account-kit/privy-integration` package, enhancing the integration of Alchemy's gas sponsorship and smart wallet features with Privy authentication. It includes new components, configuration files, and documentation for setup.

### Detailed summary
- Added `next.config.js` for Next.js configuration.
- Created `tsconfig.json` for TypeScript settings.
- Introduced `version.ts` for version management.
- Added example environment variables in `.env.example`.
- Configured ESLint rules in `.eslintrc.json`.
- Established `typedoc.json` for documentation generation.
- Updated `package.json` with new dependencies.
- Implemented `vitest.config.ts` for testing configuration.
- Created `CHANGELOG.md` for project change documentation.
- Developed components for account info, transaction sending, and token swapping.
- Enhanced README with setup instructions and project structure.
- Included CSS styles for application layout and components.

> The following files were skipped due to too many changes: `account-kit/privy-integration/README.md`, `yarn.lock`

### DIFF
--- a/docs/docs.yml
+++ b/docs/docs.yml
@@ -201,6 +201,8 @@ navigation:
             path: wallets/pages/transactions/pay-gas-with-any-token/index.mdx
           - page: "[NEW] Swap tokens"
             path: wallets/pages/transactions/swap-tokens/index.mdx
+          - page: "[NEW] Cross-Chain Swaps"
+            path: wallets/pages/transactions/cross-chain-swap-tokens/index.mdx
           - page: Send parallel transactions
             path: wallets/pages/transactions/send-parallel-transactions/index.mdx
           - page: Retry transactions
--- a/docs/pages/transactions/cross-chain-swap-tokens/api.mdx
+++ b/docs/pages/transactions/cross-chain-swap-tokens/api.mdx
@@ -1,0 +1,222 @@
+You will need to fill in values wrapped in curly braces like `{SIGNER_ADDRESS}`.
+
+<Steps>
+
+<Step title="Request an account">
+
+```bash
+curl -X POST https://api.g.alchemy.com/v2/{API_KEY} \
+  -H "Content-Type: application/json" \
+  -d '{
+    "jsonrpc": "2.0",
+    "id": 1,
+    "method": "wallet_requestAccount",
+    "params": [
+      {
+        "signerAddress": "{SIGNER_ADDRESS}"
+      }
+    ]
+  }'
+```
+
+This returns:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "accountAddress": "ACCOUNT_ADDRESS",
+    "id": "ACCOUNT_ID"
+  }
+}
+```
+
+For other potential responses, [check out the API reference!](/docs/wallets/api-reference/smart-wallets/wallet-api-endpoints/wallet-api-endpoints/wallet-request-account)
+
+</Step>
+
+<Step title="Request a cross-chain swap quote">
+
+<Info>
+  **Important**: Cross-chain swaps do not support `postCalls`. You cannot batch
+  additional actions after a cross-chain swap.
+</Info>
+
+Request a cross-chain swap quote by specifying both the source chain (`chainId`) and destination chain (`toChainId`). In addition, just like in [single-chain swaps](/wallets/transactions/swap-tokens) you can specify either a `minimumToAmount` or a `fromAmount`.
+
+<Info>
+  If you're using an EOA or just want the raw array of calls returned, pass the
+  optional parameter `"returnRawCalls": true`, this will return a `calls` array.
+</Info>
+
+```bash
+curl -X POST https://api.g.alchemy.com/v2/{API_KEY} \
+  -H "Content-Type: application/json" \
+  -d '{
+    "jsonrpc": "2.0",
+    "id": 1,
+    "method": "wallet_requestQuote_v0",
+    "params": [
+      {
+        "from": "{ACCOUNT_ADDRESS_FROM_STEP_1}",
+        "chainId": "{SOURCE_CHAIN_ID}",
+        "toChainId": "{DESTINATION_CHAIN_ID}",
+        "fromToken": "{FROM_TOKEN}",
+        "toToken": "{TO_TOKEN}",
+        "fromAmount": "{FROM_AMOUNT_HEXADECIMAL}",
+        "capabilities": {
+          "paymasterService": {
+            "policyId": "{PAYMASTER_POLICY_ID}"
+          }
+        }
+      }
+    ]
+  }'
+```
+
+This returns:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 0,
+  "result": {
+    "rawCalls": false,
+    "chainId": "...",
+    "callId": "0x...",
+    "quote": {
+      "expiry": "EXPIRY",
+      "minimumToAmount": "MINIMUM_TO_AMOUNT",
+      "fromAmount": "FROM_AMOUNT"
+    },
+    "type": "user-operation-v070",
+    "data": "USER_OPERATION_DATA",
+    "signatureRequest": {
+      "type": "personal_sign",
+      "data": {
+        "raw": "..."
+      },
+      "rawPayload": "..."
+    },
+    "feePayment": {
+      "sponsored": true,
+      "tokenAddress": "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee",
+      "maxAmount": "..."
+    }
+  }
+}
+```
+
+Note the `callId` in the response! You'll use this to track the cross-chain swap status. Also note the `signatureRequest` - this is what you need to sign, and the returned `data` field is what you'll need to send the transaction.
+
+</Step>
+
+<Step title="Sign the signature request">
+
+To sign the signature request, sign the `raw` field (note, this is not a string! You need to pass it to your signer as raw bytes, generally like so: `{ raw: "0x..." }`) with your signer of choice.
+
+This should use the `personal_sign` RPC method, as noted by the `type` in the `signatureRequest`.
+
+Alternatively, you can sign the raw payload with a simple `eth_sign` but this RPC method is not favored due to security concerns.
+
+</Step>
+
+<Step title="Send the prepared call">
+
+Pass the `callId` from Step 2 to `sendPreparedCalls`. This makes the response return the same `callId` with additional cross-chain status tracking information:
+
+```bash
+curl -X POST https://api.g.alchemy.com/v2/{API_KEY} \
+  -H "Content-Type: application/json" \
+  -d '{
+    "jsonrpc": "2.0",
+    "method": "wallet_sendPreparedCalls",
+    "params": [
+      {
+        "callId": "{CALL_ID_FROM_STEP_2}",
+        "type": "user-operation-v070",
+        "data": "{DATA_FROM_STEP_2}",
+        "chainId": "{SOURCE_CHAIN_ID}",
+        "signature": {
+          "type": "secp256k1",
+          "data": "{SIGNATURE_FROM_STEP_3}"
+        }
+      }
+    ],
+    "id": 1
+  }'
+```
+
+This returns:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": "1",
+  "result": {
+    "preparedCallIds": ["PREPARED_CALL_ID"]
+  }
+}
+```
+
+The response returns the same `callId` you passed in, which you'll use to track the cross-chain swap status in the next step.
+
+For other potential responses, [check out the API reference!](/docs/wallets/api-reference/smart-wallets/wallet-api-endpoints/wallet-api-endpoints/wallet-send-prepared-calls)
+
+</Step>
+
+<Step title="Track the cross-chain swap">
+
+Use the `wallet_getCallsStatus` endpoint to check the status of your cross-chain swap. Cross-chain swaps may take longer than single-chain swaps due to cross-chain messaging.
+
+```bash
+curl -X POST https://api.g.alchemy.com/v2/{API_KEY} \
+  -H "Content-Type: application/json" \
+  -d '{
+    "jsonrpc": "2.0",
+    "method": "wallet_getCallsStatus",
+    "params": [
+      [
+        "{CALL_ID_FROM_STEP_2_OR_STEP_4}"
+      ]
+    ],
+    "id": 1
+  }'
+```
+
+This returns:
+
+```json
+{
+  "id": "1",
+  "jsonrpc": "2.0",
+  "result": {
+    "id": "CALL_ID",
+    "chainId": "SOURCE_CHAIN_ID",
+    "atomic": true,
+    "status": 200,
+    "receipts": [...]
+  }
+}
+```
+
+Cross-chain swaps have additional status codes to reflect the cross-chain nature of the transaction:
+
+| Code | Title                   |
+| ---- | ----------------------- |
+| 100  | Pending                 |
+| 120  | Cross-Chain In Progress |
+| 200  | Confirmed               |
+| 400  | Offchain Failure        |
+| 410  | Cross-chain Refund      |
+| 500  | Onchain Failure         |
+| 600  | Partial Onchain Failure |
+
+To get your transaction hash, you can access `result.receipts[0].transactionHash`.
+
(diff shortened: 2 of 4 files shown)


account-kit/privy-integration/src/types.ts
```diff
@@ -0,0 +1,209 @@
...
+  rawCalls?: false;
+
+  /** Additional call data */
+  [key: string]: unknown;
+}
+
+/**
+ * Result of preparing a swap
+ */
+export interface PrepareSwapResult {
```
Zer0dot: Is the quote supposed to be here and in the preparedSwapCalls at the same time? Just wondering

blakecduncan: Yeah good catch - this type file was from an early prototype I made and forgot to update.  I'm going to update this to use the types from our sdk

examples/alchemy-privy-integration-example/src/components/TokenSwap.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+  const handleExecute = async () => {
+    if (!preparedSwap) return;
+
+    setValidationError("");
+    setSuccessMessage("");
+
+    try {
+      const result = await submitSwap.submitSwap(preparedSwap.preparedCalls);
+      setSuccessMessage(`Swap successful! Transaction: ${result.txnHash}`);
```
thebrianchen: can we also include a link to basescan as well for the tx hash?

examples/alchemy-privy-integration-example/src/components/SendTransaction.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    try {
+      const result = await sendTransaction({
+        to: recipient as Address,
+        value: parseEther(amount),
+        data: "0x",
+      });
+
+      setSuccessMessage(`Transaction sent! Hash: ${result.txnHash}`);
```
thebrianchen: can we include a link to the basescan url as well?

account-kit/privy-integration/src/types.ts
```diff
@@ -0,0 +1,177 @@
+import type { Address, Hash, Hex } from "viem";
+import type { swapActions } from "@account-kit/wallet-client/experimental";
+
+/**
+ * Configuration for the Alchemy provider
+ */
+export interface AlchemyProviderConfig {
+  /** Alchemy API key for @account-kit/infra transport */
+  apiKey?: string;
```
thebrianchen: this doesn't fully match the v4 spec, which has `apiKey`, `jwt`, and `rpcUrl`. prefer to keep naming consistent

account-kit/privy-integration/src/hooks/useAlchemyClient.ts
```diff
@@ -0,0 +1,222 @@
...
+
+        return {
+          ...unsignedAuth,
+          ...signature,
+        };
+      },
+    };
+
+    // Determine transport configuration
+    const transportConfig = config.url
```
thebrianchen: from what i remember parsing through this in v4: `url` and `jwt` can be provided together, but your current code prioritizes `config.url` over everything else, which means even if both are provided, only `url` gets passed.

I wonder if it's easier to just use `ConnectionConfigSchema.parse()` to creat the config instead.

account-kit/privy-integration/src/hooks/useAlchemyClient.ts
```diff
@@ -0,0 +1,222 @@
...
+} from "@account-kit/wallet-client";
+import { alchemy } from "@account-kit/infra";
+import { useAlchemyConfig } from "../Provider.js";
+import { getChain } from "../util/getChain.js";
+
+/**
+ * Module-level cache for the SmartWalletClient
+ * This ensures a single client instance is shared across all components
+ */
+let cachedClient: SmartWalletClient | null = null;
```
thebrianchen: is it correct to have module-level caches in the top-level for hooks? i was under the impression that module-level state in React can breaks the rendering model and could have edge case errors that result in share cache keys.

Should this be moved to within the AlchemyProvider as component-scoped state instead? curious why you have it out here, but don't spend too much responding to this comment if you're confident

account-kit/privy-integration/src/hooks/useAlchemyPrepareSwap.ts
```diff
@@ -0,0 +1,129 @@
...
+export function useAlchemyPrepareSwap(): UsePrepareSwapResult {
+  const { wallets } = useWallets();
+  const { client: getClient } = useAlchemyClient();
+
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<Error | null>(null);
+  const [data, setData] = useState<PrepareSwapResult | null>(null);
+
+  const getEmbeddedWallet = useCallback(() => {
+    const embedded = wallets.find((w) => w.walletClientType === "privy");
```
thebrianchen: see this logic a lot -- worth extracting into a shared util?

account-kit/privy-integration/src/hooks/useAlchemyPrepareSwap.ts
```diff
@@ -0,0 +1,129 @@
...
+          throw new Error(
+            "Received raw calls instead of prepared calls. Ensure returnRawCalls is not set to true.",
+          );
+        }
+
+        const result: PrepareSwapResult = {
+          quote,
+          // Pass the full response as preparedCalls
+          // Type assertion is safe because we validated rawCalls is not true above
+          preparedCalls: response as PreparedSwapCalls,
```
thebrianchen: nit: you destructure `quote` out then pass in the full `response` back as `preparedCalls`, which is confusing to follow. pref either just returning the response or use the destructured `preparedCalls` in the result.

account-kit/privy-integration/src/hooks/useAlchemyPrepareSwap.ts
```diff
@@ -0,0 +1,129 @@
...
+import type {
+  PrepareSwapRequest,
+  PrepareSwapResult,
+  PreparedSwapCalls,
+  UsePrepareSwapResult,
+} from "../types";
+
+/**
+ * Hook to request swap quotes and prepare swap calls
+ * Part of the two-step swap process: prepare → submit
```
thebrianchen, line 14: pls include a ref to `submitSwap()` here for devEx

account-kit/privy-integration/src/hooks/useAlchemySendTransaction.ts
```diff
@@ -0,0 +1,162 @@
...
+        // Send the transaction
+        const result = await client.sendCalls({
+          from: embeddedWallet.address as Address,
+          calls: [formattedCall],
+          capabilities,
+        });
+
+        // Wait for the transaction to be confirmed
+        const txStatus = await client.waitForCallsStatus({
+          id: result.preparedCallIds[0],
```
thebrianchen, line 123: are we guaranteed that `preparedCallIds` is non-empty? seems risky

account-kit/privy-integration/src/hooks/useAlchemySendTransaction.ts
```diff
@@ -0,0 +1,162 @@
...
+          timeout: 60_000,
+        });
+
+        const txnHash = txStatus.receipts?.[0]?.transactionHash;
```
thebrianchen, line 127: same here for `receipts` array

account-kit/privy-integration/src/hooks/useAlchemySendTransaction.ts
```diff
@@ -0,0 +1,162 @@
...
+  if (isHex(value)) {
+    return value;
+  }
+  // Assume decimal string
+  return `0x${BigInt(value).toString(16)}`;
+}
+
+/**
+ * Hook to send transactions with optional gas sponsorship via Alchemy
+ * Drop-in alternative to Privy's useSendTransaction hook
```
thebrianchen, line 36: can you add some bullets for what additional features alchemy has (gas sponsorship, etc)

account-kit/privy-integration/src/hooks/useAlchemySubmitSwap.ts
```diff
@@ -0,0 +1,121 @@
...
+ */
+export function useAlchemySubmitSwap(): UseSubmitSwapResult {
+  const { client: getClient } = useAlchemyClient();
+
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<Error | null>(null);
+  const [data, setData] = useState<SubmitSwapResult | null>(null);
+
+  const submitSwap = useCallback(
+    async (preparedCalls: PreparedSwapCalls): Promise<SubmitSwapResult> => {
```
thebrianchen: where do we check for quote expiration or is that done by backend? should we check the quote in sdk or let backend handle it? is an expired quote error clearly communicated to dev?

account-kit/privy-integration/src/hooks/useAlchemySubmitSwap.ts
```diff
@@ -0,0 +1,121 @@
...
+      setError(null);
+
+      try {
+        const client = await getClient();
+
+        // Extend client with swap actions
+        const swapClient = client.extend(swapActions);
+
+        // Sign the prepared calls
+        // Note: Gas sponsorship must be configured during the prepare step
```
thebrianchen: a little confused here, since i thought the gas sponsorship is config'd by policy id which is set in the provider level.

account-kit/privy-integration/src/types.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  /** JWT token for authentication */
+  jwt?: string;
+
+  /** Custom RPC URL */
+  url?: string;
```
thebrianchen: would port over some of the comments from `ConnectionConfig` to explain how rpcUrl works when used with `apiKey` and `jwt` just to keep things consistent

account-kit/privy-integration/src/hooks/useAlchemyClient.ts
```diff
@@ -0,0 +1,222 @@
...
+          : undefined;
+
+    if (!transportConfig) {
+      throw new Error(
+        "AlchemyProvider requires at least one of: apiKey, jwt, or url",
+      );
+    }
+
+    // Determine policy ID (use first if array)
+    const policyId = Array.isArray(config.policyId)
```
thebrianchen: in what situations do devs pass in multiple policy ids? when do we use an id other than the first one in the array. i know we take `policyId: string | string[]`, but if we're always going to take the first one, shouldn't we just use `string`? cc @jakehobbs since i remember you made the comment to extend it

account-kit/privy-integration/src/types.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  /** Policy ID(s) for gas sponsorship */
+  policyId?: string | string[];
+
+  /**
+   * Default: true → try to sponsor via Alchemy Gas Manager
+   * Set to false to disable sponsorship by default
+   */
+  defaultSponsored?: boolean;
```
thebrianchen: not sure if there is prior art or other deps, but i strongly prefer options that are default false like `disableSponsorship` or `skipSponsorship`

account-kit/privy-integration/src/types.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Options for sending a transaction
+ */
+export interface SendTransactionOptions {
+  /**
+   * Whether to sponsor the transaction
+   * Default: true if policy ID exists and defaultSponsored is not set to false
+   */
+  sponsored?: boolean;
```
thebrianchen: same here, if it doesn't go against existing patterns, would prefer default false booleans like `skipSponsorship`.

thebrianchen (approved):
great work on this! left suggestions and some questions, but nothing blocking

account-kit/privy-integration/src/hooks/useAlchemyClient.ts
```diff
@@ -0,0 +1,222 @@
...
+    getEmbeddedWallet,
+    getEmbeddedWalletChain,
+    signAuthorization,
+    config.apiKey,
+    config.jwt,
+    config.url,
+    config.policyId,
+  ]);
+
+  return { client: getClient };
```
dphilipson: Maybe name the returned field `client` instead of `getClient`? Every time this is used, you rename the return value to `getClient`, including in the readme example, so it seems like that might be a more accurate name.

account-kit/privy-integration/src/hooks/useAlchemyClient.ts
dphilipson: I believe if you pass both, then it uses your url and includes the `jwt` in the authorization header. That's probably not useful to anyone other than Alchemy devs, though.

account-kit/privy-integration/src/hooks/useAlchemyClient.ts
dphilipson: Good callout, I thought the idea was that if there are multiple policy ids, then it uses the first applicable one in the list (as in, Alchemy's backend checks each policy id in order to see its criteria fit the current call). We shouldn't need to choose one on the frontend.

account-kit/privy-integration/src/hooks/useAlchemyClient.ts
blakecduncan: That is a fair concern.  I just did a little research and my solution may cause some issues in SSR envs.

Essentially I was trying to solve this without putting the data in the context because the AlchemyProvider wraps  the PrivyProvider like below so it doesn't have access to the signer and was throwing errors.

```
    <AlchemyProvider>
      <PrivyProvider>
      </PrivyProvider>
    </AlchemyProvider>
```

maybe I'll just make it so that our provider does inside Privy's so we don't have this issue. Like
```
<PrivyProvider>
    <AlchemyProvider>
    </AlchemyProvider>
</PrivyProvider>
```

account-kit/privy-integration/src/hooks/useAlchemyClient.ts
blakecduncan: This is a great callout - updating it to use `ConnectionConfigSchema.parse()`

account-kit/privy-integration/src/hooks/useAlchemyClient.ts
blakecduncan: Yeah I missed this, I'll update.  Good catch

account-kit/privy-integration/src/hooks/useAlchemyClient.ts
blakecduncan: Yeah I'll update - that's a good point

account-kit/privy-integration/src/hooks/useAlchemySubmitSwap.ts
blakecduncan: Yeah good question, I personally think that should be handled on the backend.  In an effort to keep these hooks pretty thin right now.  I think if we build this out more it could be worth making these more robust on the front end

account-kit/privy-integration/src/hooks/useAlchemySubmitSwap.ts
blakecduncan: Yeah that comment is irrelevant now - that was from a old iteration with claude 😅

blakecduncan: In our main sdk though you can either globally set sponsorship or you can pass a policy id to individual hooks which I'm not currently doing

account-kit/privy-integration/src/hooks/useAlchemySendTransaction.ts
blakecduncan: I mention a lot of the additional features in the readme - I think I'm hesitant to mention mention things that aren't relevant to this specific function in the function jsdoc

account-kit/privy-integration/src/hooks/useAlchemyPrepareSwap.ts
blakecduncan: I fixed this - this was a little confusing after several iterations I made to the swap hooks

account-kit/privy-integration/src/hooks/useAlchemySendTransaction.ts
blakecduncan: We handle this with the `?` .  `txnHash` will just be undefined and it'll throw an error below

account-kit/privy-integration/src/hooks/useAlchemySendTransaction.ts
blakecduncan: Adding an error just in case but I'm not sure.  @jakehobbs do you happen to know if it's safe to assume there will always be a call id in this array?

PR #2156, opened 2025-10-03
feat: cross-chain API docs

Zer0dot:
This PR introduces cross-chain swap API docs.

This marks the client and react integrations as coming soon for now.


docs/docs.yml
```diff
@@ -201,6 +201,8 @@ navigation:
             path: wallets/pages/transactions/pay-gas-with-any-token/index.mdx
           - page: "[NEW] Swap tokens"
             path: wallets/pages/transactions/swap-tokens/index.mdx
+          - page: "[NEW] Cross-Chain Swaps"
```
silviavl, line 204: @Zer0dot why do we have a new page? Thought we would add cross-chain to the same page as single-chain swaps for now 

https://www.notion.so/alchemotion/Crosschain-Docs-Plan-27e069f2006680418517fdd23c84641f?source=copy_link

PR #2159, opened 2025-10-07
feat(v5): e2e passkeys

linnall:
# Pull Request Checklist

---

## PR-Codex overview
This PR enhances the `AuthClient` functionality with new parameters and methods for passkey authentication, OAuth provider management, and session handling. It introduces a `TurnkeyClient` for various authentication types and updates existing methods to support new features.

### Detailed summary
- Added `rpId` parameter for WebAuthn operations.
- Updated `loginWithPasskey` to accept a parameter object.
- Introduced methods for adding and removing OAuth providers.
- Enhanced `AuthSession` to use `TurnkeyClient` for managing passkeys and OAuth providers.
- Improved documentation for authentication methods and parameters.
- Added tests for new functionalities related to passkey and OAuth provider management.


packages/auth/src/authClient.ts
```diff
@@ -317,13 +341,79 @@ export class AuthClient {
...
+        rpId: this.rpId,
+      });
+
+      return AuthSession.create({
+        signerHttpClient: this.signerHttpClient,
+        stamper,
+        orgId: result.orgId,
+        authType: "passkey",
+        credentialId: attestation.attestation.credentialId,
+        idToken: undefined, // TO DO: do we need this OR make it so that I do not have to explicitly pass it in
```
dphilipson: You're right we don't need this here, `idToken` is specific to OAuth login (it's an OIDC id token).

We can make it so that you don't have to explicitly pass it in if you want. That would be changing the field's type definition from
```ts
idToken: string | undefined;
```
to
```ts
idToken?: string;
```
That's a stylistic choice: the advantage of requiring it explicitly is that it means the type system will make sure callers don't forget to pass this field in situations where they're supposed to be passing it, at the cost of extra boilerplate when they don't need it like here. I don't feel strongly about this, so feel free to change it if you want.

packages/auth/src/authSession.ts
```diff
@@ -401,10 +404,50 @@ export class AuthSession {
...
+    const { attestation, challenge } =
+      await this.getWebAuthnAttestationInternal({
+        username: this.user.email || "TO DO: anonymous",
+      });
+
+    const createdAt: number = Date.now();
+
+    // 2. Call Turnkey directly to add authenticator to existing org
+    const { activity } = await this.turnkey.createAuthenticators({
+      // TO DO: should we pass the turnkey client we create in AuthClient to the constructor for AuthSession?
```
dphilipson: Good call, yeah, I think we should. Previously the `authClient` didn't need a `TurnkeyClient` so we were creating it in here, but now that it does I think it makes sense to pass a `TurnkeyClient` instead of a `TurnkeyStamper` and do the wrapping on the outside.

Architecturally, I like that too. It's nice when the parameters of a class match or are closer to matching what it keeps internally as dependencies, because it makes it more flexible for testing with mocks.

packages/auth/src/authSession.ts
```diff
@@ -488,6 +539,111 @@ export class AuthSession {
     }
   }
 
+  // eslint-disable-next-line eslint-rules/require-jsdoc-on-reexported-functions
+  protected pollActivityCompletion = async <
```
dphilipson: I'd prefer not to use `protected`: that implies that this class is intended to be subclassed, which it's not (if it were, we'd need to think about that aspect of its design a lot more, happy to talk about this in detail if you're interested). I'd suggest `private` instead, or if you want this logic to be reusable then extract it into a separate helper function (you can use `TurnkeyClient` in place of `typeof this.turnkey` if you do).

Really nice helper function, though!

linnall: Thanks! It's Moldy's 😅

packages/auth/src/authClient.ts
linnall: I will keep it as is to keep its necessity for creating an `AuthSession` via login with oauth

packages/auth/src/authSession.ts
linnall: changed to be `private`

examples/auth-web-example/app/page.tsx
```diff
@@ -45,6 +43,15 @@ export default function Home(): ReactElement {
...
   }, [otpCode]);
 
+  const handleAddPasskey = useCallback(async () => {
+    await authSession?.addPasskey();
+  }, [authSession]);
+
+  const handleLoginWithPasskey = useCallback(async () => {
+    const authSession = await authClient.loginWithPasskey({ username: email });
+    setAuthSession(authSession);
+  }, [email]);
```
dphilipson, line 53: This seems like it should have `authClient` in its dependency list.

packages/auth/src/authClient.ts
```diff
@@ -323,25 +353,101 @@ export class AuthClient {
...
+
+      // create Turnkey org
+      const result = await this.signerHttpClient.request({
+        route: "signer/v1/signup",
+        method: "POST",
+        body: {
+          passkey: {
+            challenge: base64UrlEncode(attestation.challenge),
+            attestation: attestation.attestation,
+          } as any, // TO DO: remove use of as any!!
```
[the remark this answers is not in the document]
dphilipson: Oh sorry, this one is on me, I have the wrong type for the passkey field in `SignerHttpSchema`. I can make a quick fix.

packages/auth/src/authClient.ts
```diff
@@ -323,25 +344,106 @@ export class AuthClient {
...
+        turnkey: new TurnkeyClient(
+          { baseUrl: "https://api.turnkey.com" },
+          stamper,
+        ),
+        orgId: result.orgId,
+        authType: "passkey",
+        credentialId: attestation.attestation.credentialId,
+        idToken: undefined,
+        expirationDateMs:
+          Date.now() + (sessionExpirationMs ?? DEFAULT_SESSION_EXPIRATION_MS),
```
dphilipson, line 510: It may be the case that there's no need to ever expire a passkey session, since user interaction is required for every privileged action. If we do keep an expiration, we'll want to let the caller configure it, similar to the other session expirations since https://github.com/alchemyplatform/aa-sdk/pull/2164 merged. Likewise below.

packages/auth/src/authSession.ts
```diff
@@ -406,14 +418,53 @@ export class AuthSession {
...
   public async addPasskey(
-    params: CredentialCreationOptions,
+    params?: CredentialCreationOptionOverrides,
   ): Promise<PasskeyInfo> {
     this.throwIfDisconnected();
-    return notImplemented(params);
+
+    const { attestation, challenge } = await getWebAuthnAttestationInternal(
+      {
+        username: this.user.email || "TO DO: anonymous",
```
dphilipson: Reminder to change this string before merging.

packages/auth/src/authSession.ts
```diff
@@ -497,6 +557,56 @@ export class AuthSession {
     }
   }
 
+  // eslint-disable-next-line eslint-rules/require-jsdoc-on-reexported-functions
```
dphilipson: This eslint directive is probably not needed for private functions, although I could be wrong.

packages/auth/src/authClient.ts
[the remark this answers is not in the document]
dphilipson: Fix is here: https://github.com/alchemyplatform/aa-sdk/pull/2175

PR #2173, opened 2025-10-10
feat(v5): adds addOauthProvider and removeOauthProvider implementations

linnall:
# Pull Request Checklist

---

## PR-Codex overview
This PR introduces the functionality to add and remove OAuth providers in the `AuthSession` class, along with corresponding API routes and tests to ensure these features work as intended.

### Detailed summary
- Added new API routes for adding (`signer/v1/add-oauth-provider`) and removing (`signer/v1/remove-oauth-provider`) OAuth providers.
- Implemented `addOauthProvider` method in `AuthSession` to handle adding providers.
- Implemented `removeOauthProvider` method in `AuthSession` to handle removing providers.
- Created unit tests for both `addOauthProvider` and `removeOauthProvider` methods to verify functionality.

### DIFF
--- a/packages/aa-infra/src/schema.ts
+++ b/packages/aa-infra/src/schema.ts
@@ -218,7 +218,15 @@ export type SignerHttpSchema = [
     Route: "signer/v1/signup";
     Method: "POST";
     Body: {
-      passkey?: PublicKeyCredential;
+      passkey?: {
+        challenge: string;
+        attestation: {
+          credentialId: string;
+          clientDataJson: string;
+          attestationObject: string;
+          transports: string[];
+        };
+      };
       email?: string;
       emailMode?: "magicLink" | "otp";
       phone?: string;
--- a/packages/auth/tests/auth-client.test.ts
+++ b/packages/auth/tests/auth-client.test.ts
@@ -12,6 +12,10 @@ import { DEFAULT_SESSION_EXPIRATION_MS } from "../src/utils.js";
 import type { SignerHttpSchema } from "@alchemy/aa-infra";
 import type { AlchemyRestClient } from "@alchemy/common";
 import { TurnkeyClient } from "@turnkey/http";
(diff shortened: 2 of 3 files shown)


packages/auth/src/authClient.ts
thebrianchen, line 167: just curious -- why do we need this value for root org id? can you add a comment?

linnall: this is a good question. I pulled this root org id from the v4 implementation of passkeys. It seems like we use this as an admin credential to lookup the actual org id of the current user that is trying to login

linnall: cc: @dphilipson

packages/auth/src/authClient.ts
```diff
@@ -123,6 +139,7 @@ export class AuthClient {
    * @param {string} [params.apiKey] - API key for authentication with Alchemy services
    * @param {string} [params.jwt] - JWT token for authentication with Alchemy services
    * @param {string} [params.url] - Custom URL (optional - defaults to Alchemy's chain-agnostic URL)
+   * @param {string} [params.rpId] - Relying Party ID for WebAuthn operations
```
dphilipson, line 142: Typo:
```suggestion
   * @param {string} [params.rpId] - Relaying Party ID for WebAuthn operations
```

packages/auth/src/authClient.ts
dphilipson: Yeah, that's right. Since the user doesn't store their org id, we need to look it up once they present their passkey, which means doing a lookup in the parent org (the root org that we control).

packages/auth/src/authClient.ts
dphilipson: Actually, never mind, I'm wrong. I've been wrong about what rpId stands for the entire time I've worked on this, haha.

PR #2102, opened 2025-09-24
feat(v5): hooks for preparing & sending calls (including swaps)

jakehobbs:
# Pull Request Checklist

- [ ] Did you add new tests and confirm existing tests pass? (`yarn test`)
- [ ] Did you update relevant docs? (docs are found in the `site` folder, and guidelines for updating/adding docs can be found in the [contribution guide](https://github.com/alchemyplatform/aa-sdk/blob/main/CONTRIBUTING.md))
- [ ] Do your commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [ ] Does your PR title also follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard?
- [ ] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples)? (e.g. `feat!: breaking change`)
- [ ] Did you run lint (`yarn lint:check`) and fix any issues? (`yarn lint:write`)
- [ ] Did you follow the [contribution guidelines](https://github.com/alchemyplatform/aa-sdk/blob/main/CONTRIBUTING.md)?

---

## PR-Codex overview
This PR focuses on enhancing the `@alchemy/wallet-apis` and `@alchemy/react` libraries by adding new functionalities for handling smart wallet clients, improving API requests, and introducing better hooks for preparing and sending transactions.

### Detailed summary
- Removed `packages/wallet-apis/src/index.tsdocs-site`.
- Added `alchemyAuthClient` and `alchemySmartWalletClient` names in client configuration.
- Updated TypeScript configuration in `tsconfig.build.json`.
- Introduced new utility functions for merging client capabilities.
- Added new hooks: `usePrepareSwap`, `useSendPreparedCalls`, and `usePrepareCalls`.
- Implemented type guard functions: `isSmartWalletClient` and `assertSmartWalletClient`.
- Enhanced `sendPreparedCalls` and `prepareSwap` functionalities.
- Updated documentation for several functions and hooks.
- Improved error handling in the `alchemySmartWallet` connector.
- Added new exports for actions and decorators in `wallet-apis`.

[...]

PR #2181, opened 2025-10-14
feat(v5): add sms auth support

thebrianchen:
Porting over sms auth: send OTP, verify OTP, lookup user, modify phone number
- added actions and hooks in `wagmi-core` and `react`
- updated example in `connectors-web-example` and `react-v5-example`

# Pull Request Checklist

---

## PR-Codex overview
This PR introduces SMS OTP authentication functionality, allowing users to request and verify one-time passwords via SMS. It includes updates to the `AuthClient`, new hooks for React, and documentation changes to reflect the new SMS features.

### Detailed summary
- Added `sendSmsOtp` method to `AuthClient` for SMS OTP requests.
- Implemented `lookupUserByPhone` method for checking phone registration.
- Created `useSendSmsOtp` hook for React integration.
- Updated schemas and types to include phone number support.
- Enhanced documentation for new SMS functionalities and usage examples.
- Updated UI components to handle SMS authentication.
- Added tests for SMS OTP functionality and phone number management.

### DIFF
--- a/packages/react/src/hooks/useSendSmsOtp.ts
+++ b/packages/react/src/hooks/useSendSmsOtp.ts
@@ -1,29 +1,18 @@
 "use client";
 
-import type { MutateOptions } from "@tanstack/query-core";
 import { useMutation } from "@tanstack/react-query";
 import type { UseMutationParameters, UseMutationReturnType } from "wagmi/query";
 import { useConfig } from "wagmi";
 import {
-  sendSmsOtp,
   type SendSmsOtpParameters,
   type SendSmsOtpReturnType,
 } from "@alchemy/wagmi-core";
 import type { ConfigParameter } from "../types";
-
-type SendSmsOtpMutate = (
-  variables: SendSmsOtpParameters,
-  options?:
-    | MutateOptions<SendSmsOtpReturnType, Error, SendSmsOtpParameters>
-    | undefined,
-) => void;
-
-type SendSmsOtpMutateAsync = (
-  variables: SendSmsOtpParameters,
-  options?:
-    | MutateOptions<SendSmsOtpReturnType, Error, SendSmsOtpParameters>
-    | undefined,
-) => Promise<SendSmsOtpReturnType>;
+import {
+  sendSmsOtpMutationOptions,
+  type SendSmsOtpMutate,
+  type SendSmsOtpMutateAsync,
+} from "../query/sendSmsOtp.js";
 
 export type UseSendSmsOtpParameters = ConfigParameter & {
   mutation?:
@@ -94,12 +83,11 @@ export function useSendSmsOtp(
 
   const config = useConfig(parameters);
 
+  const mutationOptions = sendSmsOtpMutationOptions(config);
+
   const { mutate, mutateAsync, ...result } = useMutation({
     ...mutation,
-    mutationKey: ["sendSmsOtp"],
-    mutationFn: (variables) => {
-      return sendSmsOtp(config, variables);
-    },
+    ...mutationOptions,
   });
 
   return {
--- a/packages/react/src/query/sendSmsOtp.ts
+++ b/packages/react/src/query/sendSmsOtp.ts
@@ -1,0 +1,34 @@
+import {
+  sendSmsOtp,
+  type SendSmsOtpParameters,
+  type SendSmsOtpReturnType,
+} from "@alchemy/wagmi-core";
+import type { MutateOptions, MutationOptions } from "@tanstack/react-query";
+import type { Config } from "wagmi";
+
+export type SendSmsOtpMutate = (
+  variables: SendSmsOtpParameters,
+  options?:
+    | MutateOptions<SendSmsOtpReturnType, Error, SendSmsOtpParameters>
+    | undefined,
+) => void;
+
+export type SendSmsOtpMutateAsync = (
+  variables: SendSmsOtpParameters,
+  options?:
+    | MutateOptions<SendSmsOtpReturnType, Error, SendSmsOtpParameters>
+    | undefined,
+) => Promise<SendSmsOtpReturnType>;
+
+export function sendSmsOtpMutationOptions(config: Config) {
+  return {
+    mutationKey: ["sendSmsOtp"],
+    mutationFn: (variables) => {
+      return sendSmsOtp(config, variables);
+    },
+  } as const satisfies MutationOptions<
+    SendSmsOtpReturnType,
+    Error,
+    SendSmsOtpParameters
+  >;
+}


examples/react-v5-example/.env.example
```diff
@@ -1 +1,2 @@
 NEXT_PUBLIC_ALCHEMY_API_KEY=
+NEXT_PUBLIC_PAYMASTER_POLICY_ID=
```
thebrianchen, line 2: adding this since it was omitted in an earlier PR

packages/auth/src/authClient.ts
```diff
@@ -50,11 +50,21 @@ export type SendEmailOtpParams = {
...
+  /** Length of the session in milliseconds. Defaults to 15 minutes. */
+  sessionExpirationMs?: number;
+};
+
 /**
  * Parameters for submitting an OTP code for verification
  */
 export type SubmitOtpCodeParams = {
-  /** The OTP code received via email */
+  /** The OTP code received via email or SMS */
```
thebrianchen, line 67: seemed like the OTP code flow was the same for email or sms, but wanted to flag to double check

packages/auth/src/authClient.ts
```diff
@@ -191,20 +201,109 @@ export class AuthClient {
...
+   * ```ts twoslash
+   * const user = await authClient.lookupUserByPhone("+12025551234");
+   * if (user) {
+   *   console.log("User exists with orgId:", user.orgId);
+   * } else {
+   *   console.log("Phone number not registered");
+   * }
+   * ```
+   */
+  public async lookupUserByPhone(
```
thebrianchen, line 276: also wanted thoughts on exposing this in authclient all the way up to wagmi core. this used to live on the signer directly as a method in v4, wondering if this pattern makes sense

packages/auth/src/authSession.ts
```diff
@@ -108,7 +109,7 @@ export class AuthSession {
   private constructor(
     private readonly signerHttpClient: AlchemyRestClient<SignerHttpSchema>,
     private readonly turnkey: TurnkeyClient,
-    private readonly user: User,
+    private user: User,
```
thebrianchen, line 112: removing readonly since sms auth methods allow modifying the phone number. we'll eventually need this for email as well. wanted to find a way to not remove this, but couldn't find a good one.

wanted to flag in case there are downstream effects we need to change as a result

packages/react/src/hooks/useLookupUserByPhone.ts
```diff
@@ -0,0 +1,124 @@
+"use client";
+
+import type { MutateOptions } from "@tanstack/query-core";
```
thebrianchen: @jakehobbs pls review the react ones more deeply. these are mostly AI-created, and i lack the skill to properly review them beyond checking that they work in the react app

packages/auth/src/authClient.ts
jakehobbs: correct

packages/auth/src/authClient.ts
jakehobbs: curious what @dphilipson thinks. i haven't dealt w/ this much.

packages/auth/src/authSession.ts
jakehobbs: i think it makes perfect sense to not be readonly since it contains properties that can be updated

packages/react/src/hooks/useLookupUserByPhone.ts
jakehobbs: oh i was going to add these in a separate PR. in general, i think any "lookup" methods should probably be queries, not mutations.

examples/connectors-web-example/src/main.ts
```diff
@@ -1,5 +1,7 @@
 import {
   sendEmailOtp,
+  sendSmsOtp,
```
jakehobbs, line 3: curious if you think we should continue maintaining both the `connectors-web-example` and the `react-v5-example`. i personally feel that it's pretty safe to assume that the core pkg is working correctly if react is, since react is built on top of it. (but i think there are some things in the core example that aren't in the react example yet.)

examples/react-v5-example/.env.example
jakehobbs: ty

examples/react-v5-example/src/app/page.tsx
```diff
@@ -38,7 +45,31 @@ export default function Home() {
         {account.isConnected ? "Connected" : "Not connected"}
       </p>
       {!account.isConnected ? (
-        <EmailAuthDemo />
+        <div className="flex flex-col gap-4 items-center">
```
jakehobbs: nit: i think it might be worth moving this (and its state) to a subcomponent to keep this component cleaner. (if we add this here, i think it opens up a door for us to start just adding a ton of other stuff at in this top level component too.)

jakehobbs: i'm planning to add oauth login methods to the react demo soon too

examples/react-v5-example/src/app/page.tsx
```diff
@@ -173,6 +204,62 @@ const EmailAuthDemo = () => {
   );
 };
 
+const SmsAuthDemo = () => {
+  const { sendSmsOtpAsync, isSuccess: sentSmsOtp } = useSendSmsOtp();
+  const { lookupUserByPhoneAsync } = useLookupUserByPhone();
```
jakehobbs: if it exists, i think `useLookupUserByPhone` should be a query, not a mutation

examples/react-v5-example/src/app/page.tsx
```diff
@@ -173,6 +204,62 @@ const EmailAuthDemo = () => {
...
+          onClick={async () => {
+            const phoneNumber = prompt(
+              "Enter your phone number (E.164 format, e.g., +15551234567):",
+            );
+            if (!phoneNumber) {
+              return;
+            }
+            // Optional: lookup if phone is registered
+            try {
+              const result = await lookupUserByPhoneAsync({ phoneNumber });
```
jakehobbs: do we really want to expose the lookup step to the user? i believe in v4 we just automatically handled it, but i'm not 100% sure.

jakehobbs: oh, i guess it's optional since the auth flow doesn't change at all based on the user already existing?

jakehobbs: i think in v4, we may have also just had a property like `isNewUser` on the resulting user after logging in if they were newly created. not sure if we are still doing something like that too or not.

jakehobbs: just trying to think of why a dev would want to do a lookup like this

examples/react-v5-example/src/app/page.tsx
```diff
@@ -173,6 +204,62 @@ const EmailAuthDemo = () => {
...
+        </button>
+        {sentSmsOtp && (
+          <button
+            className="cursor-pointer rounded bg-green-500 px-4 py-2 font-bold text-white hover:bg-green-700"
+            onClick={async () => {
+              const otpCode = prompt("Enter the OTP code:");
+              if (!otpCode) {
+                return;
+              }
+              await submitOtpCodeAsync({ otpCode });
```
jakehobbs: nit: you can just use the non-async methods (i.e. `submitOtpCode`) & make this callback non-async, since you aren't actually using the result of this promise.

packages/auth/src/authClient.ts
```diff
@@ -191,20 +201,109 @@ export class AuthClient {
...
+          body: { phone: phoneNumber, targetPublicKey },
+        });
+      } else {
+        // Existing user - login
+        return this.signerHttpClient.request({
+          route: "signer/v1/auth",
+          method: "POST",
+          body: { phone: phoneNumber, targetPublicKey },
+        });
+      }
```
jakehobbs, line 253: ❗ i think we need to submit `expirationSeconds` to `signup`/`auth` endpoints like we do when sending email otp.

packages/auth/src/authClient.ts
jakehobbs: after seeing how it's used, i think i lean slightly towards not exposing it in wagmi core unless someone asks for it.

packages/wagmi-core/src/actions/lookupUserByPhone.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type LookupUserByPhoneParameters = {
+  /** The user's phone number with country code to look up */
+  phoneNumber: string;
+};
+
+export type LookupUserByPhoneReturnType = { orgId: string } | null;
+
+/**
+ * Looks up if a phone number is registered in the system.
```
jakehobbs: nit: "in the system" seems like a slightly weird thing to say. could we be more specific here?

packages/react/src/hooks/useLookupUserByPhone.ts
jakehobbs: i refactored our pattern for the react hooks a bit in #2102. if that gets merged first, we should update these to match.

if this gets merged first, i should update them to match in my pr.

packages/react/src/hooks/useLookupUserByPhone.ts
```diff
@@ -0,0 +1,124 @@
...
+ *         Check Phone
+ *       </button>
+ *       {data && <p>User exists: {data.orgId}</p>}
+ *       {data === null && <p>Phone not registered</p>}
+ *     </div>
+ *   );
+ * }
+ * ```
+ */
+export function useLookupUserByPhone(
```
jakehobbs: i feel fairly strongly that this should be a query, if it exists at all

packages/react/src/hooks/useSendSmsOtp.ts
```diff
@@ -0,0 +1,110 @@
...
+ *
+ *   return (
+ *     <button onClick={handleSendOtp} disabled={isPending}>
+ *       {isPending ? 'Sending...' : 'Send OTP'}
+ *     </button>
+ *   );
+ * }
+ * ```
+ */
+export function useSendSmsOtp(
```
jakehobbs, line 79: this one looks good. 👍 
it just needs to be refactored once my other PR lands.

examples/connectors-web-example/src/main.ts
dphilipson: Yeah, I think we can reasonably move to the react example, although we should port over the functionality if we want to do that and it's not the highest priority for now.

examples/react-v5-example/src/app/page.tsx
dphilipson: Yeah, that's true. I can't really think of a use case for a dev looking up if a phone number already exists. Our login methods are already handling the logic to look up the user and either create a new one or log in to an old one (I'm not sure off the top of my head if it currently returns `isNewUser` though, we should make sure it does).

packages/auth/src/authClient.ts
dphilipson: That's actually sent when submitting the otp code I believe, hence the reason for saving it as `this.pendingExpirationDateMs` here.

packages/auth/src/authClient.ts
dphilipson: I'm not sure we need to expose it from wagmi-core. In v4, every time I needed to access the client directly I felt like it was because there should have been a hook for what I was doing but there wasn't.

packages/auth/src/authClient.ts
jakehobbs: `sendEmailOtp` passes `expirationSeconds` to auth/signup endpoints.

examples/react-v5-example/src/app/page.tsx
thebrianchen: refactored into `AuthenticationDemo` subcomponent

examples/react-v5-example/src/app/page.tsx
thebrianchen: removed

examples/react-v5-example/src/app/page.tsx
thebrianchen: makes sense, removed

examples/react-v5-example/src/app/page.tsx
thebrianchen: done, here and throughout

packages/auth/src/authClient.ts
thebrianchen: removed in wagmi-core and react.

packages/auth/src/authClient.ts
thebrianchen: Used `Math.floor()` to calculate expiration seconds and passing it down in the body

packages/react/src/hooks/useLookupUserByPhone.ts
thebrianchen: deleted this file

packages/react/src/hooks/useLookupUserByPhone.ts
thebrianchen: deleted this file

packages/react/src/hooks/useSendSmsOtp.ts
thebrianchen: how should this be refactored?

packages/wagmi-core/src/actions/lookupUserByPhone.ts
thebrianchen: deleted this file

examples/connectors-web-example/src/main.ts
thebrianchen: yeah, let's port the functionality over. can attempt this in another PR down the road

packages/react/src/hooks/useSendSmsOtp.ts
jakehobbs: some of the types, the `mutationKey`, and the `mutationFn` should be moved to a different file in `src/query`. you can compare to the updated `useSendEmailOtp` in `moldy/v5-base` now.

thebrianchen: done.

PR #2188, opened 2025-10-16
feat(v5): add missing signer react hooks

jakehobbs:
# Pull Request Checklist

---

## PR-Codex overview
This PR primarily enhances the authentication and user management functionalities in the Alchemy Auth connector, introducing new hooks, methods, and documentation for better user experience and handling of authentication sessions.

### Detailed summary
- Added `updateEmail` and `updatePhoneNumber` actions with verification processes.
- Introduced `useUpdateEmail` and `useUpdatePhoneNumber` hooks for React.
- Updated `getAuthSession` to return `AuthSession | null` synchronously.
- Enhanced error handling in authentication methods.
- Added documentation for new methods and hooks.
- Improved user data handling with `watchUser` functionality.
- Created hooks for OAuth login and session management.

> The following files were skipped due to too many changes: `packages/smart-accounts/tests/coinbase-account.test.ts`, `examples/react-v5-example/src/app/page.tsx`, `packages/auth/src/authSession.ts`

### DIFF
--- a/packages/react/src/hooks/useHandleOauthRedirect.ts
+++ b/packages/react/src/hooks/useHandleOauthRedirect.ts
@@ -83,6 +83,7 @@ export function useHandleOauthRedirect(
         setError(null);
       })
       .catch((err: Error) => {
+        console.error(`Error handling OAuth redirect:`, err);
         setError(err);
         setIsSuccess(false);
       })


packages/react/src/hooks/useLoginWithOauth.ts
```diff
@@ -0,0 +1,103 @@
...
+ * - `isIdle`: `boolean` - True if the mutation is in its initial idle state
+ * - `isPending`: `boolean` - True if the mutation is currently executing
+ * - `isSuccess`: `boolean` - True if the last mutation attempt was successful
+ * - `failureCount`: `number` - The failure count for the mutation
+ * - `failureReason`: `Error | null` - The failure reason for the mutation retry
+ * - `isPaused`: `boolean` - True if the mutation has been paused
+ * - `reset`: `() => void` - Function to reset the mutation to its initial state
+ * - `status`: `'idle' | 'pending' | 'error' | 'success'` - Current status of the mutation
+ * - `submittedAt`: `number` - Timestamp for when the mutation was submitted
+ * - `variables`: `LoginWithOauthParameters | undefined` - The variables object passed to the mutation
```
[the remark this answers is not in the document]
jakehobbs: we are doing this other places too. let's revisit before releasing.

packages/react/src/hooks/useHandleOauthRedirect.ts
```diff
@@ -0,0 +1,105 @@
...
+          window.location.pathname,
+        );
+      })
+      .catch((err: Error) => {
+        setError(err);
+        setIsSuccess(false);
+      })
+      .finally(() => {
+        setIsPending(false);
+      });
```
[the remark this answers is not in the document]
jakehobbs: the params are already being cleared by some other function

packages/wagmi-core/src/actions/updateEmail.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  if (!authSession) {
+    throw new BaseError("No active auth session. Please authenticate first.");
+  }
+
+  // Remove email.
+  if ("email" in parameters) {
+    if (parameters.email !== null) {
+      throw new BaseError(
+        "To update email, must provide a verification code instead of an email address.",
```
blakecduncan: nit I think this error message confused me a little.  Wondering if we can be more verbose here

Something like:
```ts
throw new BaseError(
  "The 'email' parameter only accepts null for removal. " +
  "To update email, use { verificationCode: 'code' } after calling sendVerificationCode()."
);
```

packages/wagmi-core/src/actions/updatePhoneNumber.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  if (!authSession) {
+    throw new BaseError("No active auth session. Please authenticate first.");
+  }
+
+  // Remove phone number.
+  if ("phoneNumber" in parameters) {
+    if (parameters.phoneNumber !== null) {
+      throw new BaseError(
+        "To update phone number, must provide a verification code instead of a phone number.",
```
blakecduncan: same comment here as above about a more verbose error message

packages/wagmi-core/src/actions/getAuthClient.ts
```diff
@@ -0,0 +1,20 @@
+import { type Config } from "@wagmi/core";
```
blakecduncan, line 1: These actions are so clean with the new signer interfaces 🔥

packages/auth/src/authSession.ts
```diff
@@ -368,37 +372,172 @@ export class AuthSession {
...
+   * await authSession.sendEmailVerificationCode("user@example.com");
+   * // User receives email with code
+   * const code = prompt("Enter code from email:");
+   * await authSession.setEmail({ verificationCode: code });
+   * ```
+   */
+  public async sendEmailVerificationCode(email: string): Promise<void> {
+    this.throwIfDisconnected();
+
+    const { otpId } = await this.signerHttpClient.request({
```
blakecduncan: Claude recommended this and I think it's a valid point but it might be overkill (so feel free to disregard) - since we store the `pendingEmailOtpId` ourselves, it may be worth protecting against concurrent requests.  Since they will overwrite the `pendingEmailOtpId` value.

```ts
  private isEmailVerificationPending = false;
  private isPhoneVerificationPending = false;

public async sendEmailVerificationCode(email: string): Promise<void> {
  this.throwIfDisconnected();
  
   if (this.isEmailVerificationPending) {
     throw new BaseError(
       "Email verification already in progress. Please wait or complete the current verification.",
     );
   }

   this.isEmailVerificationPending = true;
   try {
    const { otpId } = await this.signerHttpClient.request({
      route: "signer/v1/init-otp",
      method: "POST",
      body: {
        contact: email,
        otpType: "OTP_TYPE_EMAIL",
      },
    });

    this.pendingEmailOtpId = otpId;
   } finally {
     this.isEmailVerificationPending = false;
   }
}
```

then we can make sure to set `this.isEmailVerificationPending = false;` in the `setEmail()` function

examples/react-v5-example/src/app/page.tsx
```diff
@@ -24,13 +24,25 @@ import {
...
   useSendPreparedCalls,
   useSubmitOtpCode,
+  useLoginWithOauth,
+  useHandleOauthRedirect,
+  useUser,
+  useAuthSession,
+  useAuthClient,
+  useSendVerificationCode,
+  useUpdateEmail,
+  useUpdatePhoneNumber,
```
blakecduncan, line 34: 🔥

blakecduncan (commented):
Just some minor comments but this is awesome! Nice job

packages/wagmi-core/src/actions/updateEmail.ts
jakehobbs: sounds good. this is also impossible to hit btw unless they are ignoring the type system, but i figured it would be good to have just in case.

packages/auth/src/authSession.ts
jakehobbs: if we made that change, how would someone request to re-send an OTP?

i think if someone requests multiple OTPs, only the last one requested should be valid. this seems like best practice anywhere that requests an OTP, in my experience.

thoughts?

blakecduncan: Oh yeah, I didn't think about resend.  With that in mind it's probably best to keep it as is.  This protects against an edge case that users really shouldn't run into.  

The case I was thinking of that would be tricky is if a user double clicked to send an email - since it's async we can't guarantee which otp code gets set to the class variable.  But I mean any reasonable frontend dev should protect against allowing a user to double click a button.

One alternative is returning the otp code to the dev so they can call it as many times as they want.  Because I actually don't know if we invalidate an outstanding otp if another one is requested.  But after giving this more thought - I think of all the options, your current implementation is the best

jakehobbs: > The case I was thinking of that would be tricky is if a user double clicked to send an email - since it's async we can't guarantee which otp code gets set to the class variable. But I mean any reasonable frontend dev should protect against allowing a user to double click a button.

oh hmmm, that's a good point. do you think it's overkill to add something like this?

```ts
private isSendingEmailVerificationCode = false;

public async sendEmailVerificationCode(email: string): Promise<void> {
    this.throwIfDisconnected();

    if (this.isSendingEmailVerificationCode) {
      throw new BaseError(
        "Code sending is already in progress. Please wait for it to finish before requesting another code.",
      );
    }

    this.isSendingEmailVerificationCode = true;

    const { otpId } = await this.signerHttpClient.request({
      route: "signer/v1/init-otp",
      method: "POST",
      body: {
        contact: email,
        otpType: "OTP_TYPE_EMAIL",
      },
    });

    this.pendingEmailOtpId = otpId;
    this.isSendingEmailVerificationCode = false;
  }
```

jakehobbs: would need a `finally` in there too in case the request  throws

packages/wagmi-core/src/actions/updateEmail.ts
```diff
@@ -0,0 +1,72 @@
...
+  const authSession = connector.getAuthSession();
+
+  if (!authSession) {
+    throw new BaseError("No active auth session. Please authenticate first.");
+  }
+
+  // Remove email.
+  if ("email" in parameters) {
+    if (parameters.email !== null) {
+      console.log(parameters.email);
```
thebrianchen: did you mean to leave the `console.log` in here?

packages/auth/src/authSession.ts
blakecduncan: No I don't think it's overkill, I think that would be safe to add.  This is pretty much what I was originally thinking but it's an edge case so I didn't have too strong a conviction.

I would add the `try/finally` though.  Because without it, if for some reason the api request fails, we would be stuck with `isSendingEmailVerificationCode = true`

Like:

```ts
   try {
    const { otpId } = await this.signerHttpClient.request({
      route: "signer/v1/init-otp",
      method: "POST",
      body: {
        contact: email,
        otpType: "OTP_TYPE_EMAIL",
      },
    });

    this.pendingEmailOtpId = otpId;
   } finally {
     this.isEmailVerificationPending = false;
   }
   ```

Comment (jakehobbs) [reply packages/auth/src/authSession.ts]:
```ts
public async sendEmailVerificationCode(email: string): Promise<void> {
    this.throwIfDisconnected();

    if (this.isSendingEmailVerificationCode) {
      throw new BaseError(
        "Code sending is already in progress. Please wait for it to finish before requesting another code.",
      );
    }

    this.isSendingEmailVerificationCode = true;

    try {
      const { otpId } = await this.signerHttpClient.request({
        route: "signer/v1/init-otp",
        method: "POST",
        body: {
          contact: email,
          otpType: "OTP_TYPE_EMAIL",
        },
      });
      this.pendingEmailOtpId = otpId;
    } catch (err) {
      throw err;
    } finally {
      this.isSendingEmailVerificationCode = false;
    }
  }
```

packages/react/src/query/sendVerificationCode.ts
```diff
@@ -0,0 +1,42 @@
...
+        SendVerificationCodeReturnType,
+        Error,
+        SendVerificationCodeParameters
+      >
+    | undefined,
+) => Promise<SendVerificationCodeReturnType>;
+
+export function sendVerificationCodeMutationOptions(config: Config) {
+  return {
+    mutationKey: ["sendVerificationCode"],
```
thebrianchen, line 33: can you briefly walk me through how the caching works if there are multiple verification requests that are sent in a row? would there be cache collisions?

packages/auth/src/authSession.ts
jakehobbs: 100%, will do

packages/wagmi-core/src/actions/sendVerificationCode.ts
```diff
@@ -0,0 +1,58 @@
...
+): Promise<SendVerificationCodeReturnType> {
+  const { connector } = resolveAlchemyAuthConnector(config);
+  const authSession = connector.getAuthSession();
+
+  if (!authSession) {
+    throw new BaseError("No active auth session. Please authenticate first.");
+  }
+
+  if (parameters.type === "email") {
+    await authSession.sendEmailVerificationCode(parameters.contact);
```
thebrianchen: as a general rule, do we let the backend do all email and phone validation? what's our current error wrapping behavior?

packages/react/src/hooks/useHandleOauthRedirect.ts
```diff
@@ -0,0 +1,104 @@
...
+        setIsSuccess(true);
+        setError(null);
+        // Clean up OAuth-specific URL parameters while preserving others
+        const url = new URL(window.location.href);
+        url.searchParams.delete("alchemy-bundle");
+        url.searchParams.delete("alchemy-org-id");
+        window.history.replaceState({}, document.title, url.toString());
+      })
+      .catch((err: Error) => {
+        setError(err);
```
thebrianchen, line 87: do we need to delete search params if there's an error?

packages/smart-accounts/tests/coinbase-account.test.ts
```diff
@@ -32,145 +32,157 @@ describe("Viem AA - Coinbase Smart Account", () => {
...
-      address: account.address,
-      message,
-      signature,
-    });
-
-    expect(isValid).toBe(true);
-  }, 60_000);
+  it(
+    "should send a user operation (ETH transfer)",
+    { retry: 3, timeout: 60_000 },
```
jakehobbs, line 37: adding retries here b/c this test fails a lot

packages/wagmi-core/src/actions/updateEmail.ts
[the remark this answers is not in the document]
jakehobbs: good bot

packages/react/src/hooks/useAuthSession.ts
```diff
@@ -0,0 +1,70 @@
...
+): UseAuthSessionReturnType {
+  const config = useConfig(parameters);
+
+  const authSession = useSyncExternalStore(
+    (onChange) => {
+      let disconnectUnsubscribe: (() => void) | undefined;
+
+      // Subscribe to account status changes
+      const unwatchAccount = config.subscribe(
+        (state) => state.status,
```
thebrianchen, line 34: just realized -- i don't think we currently emit a status change on alchemyAuth when a session is restored silently. can you add a "change" event?

packages/connectors-web/src/alchemyAuth.ts
```diff
@@ -522,9 +521,8 @@ export function alchemyAuth(options: AlchemyAuthOptions): CreateConnectorFn {
       // --- Custom methods for Alchemy Auth are defined below this comment ---
       getAuthClient,
 
-      async getAuthSession() {
-        await ensureSession();
```
thebrianchen, line 526: i don't get how we are able to remove `ensureSession()` without potentially breaking the app on reload due to race conditions. 

on reload, `setup` calls `tryResume()` in the background. While `resumePromise` is in flight, `authSessionInstance` is still undefined. during that time, synchronous calls to `getAuthSession()` are going to return `null`. this would break calls like `getUser`, `sendVerificationCode`, and other actions relying on sync behavior, since the call to `getAuthSession` would return `null` and then immediately throw, even though the session would be restored moments later. 

iiuc, we needed to make this sync to be compatible with `useSyncExternalStore` in `useAuthSession`. 

thoughts on adding something along the lines of `ensureAuthSession` for usage in the wagmi actions? open to other ideas, but this current design feels brittle for actions due to the incorrect throwing potential

packages/connectors-web/tests/alchemyAuth.test.ts
thebrianchen, line 13: nit: change type to sync

packages/wagmi-core/src/actions/updateEmail.ts
jakehobbs: nope, removed

packages/react/src/query/sendVerificationCode.ts
jakehobbs: mutations don't have caching, only queries do.

jakehobbs: according to chatgpt:

the mutationKey doesn’t merge behavior, but it does let TanStack:
	•	group them in Devtools (you’ll see both under the same label)
	•	query or filter them via queryClient.getMutationCache().findAll({ mutationKey: [...] })
	•	manually dedupe if you write custom logic (e.g. via a mutation observer)

so if you run two concurrent ['sendMessage'] mutations:
	•	they’ll appear as two separate rows in Devtools,
	•	but both will share the same “key” grouping label.

jakehobbs: Screenshot 2025-10-17 at 4 16 54 PM

jakehobbs: here's an explanation from the authors: https://github.com/TanStack/query/discussions/6093

packages/react/src/hooks/useHandleOauthRedirect.ts
jakehobbs: hmm good question, probably a good idea. thoughts?

packages/connectors-web/src/alchemyAuth.ts
jakehobbs: i'm manually checking for the auth session in the wagmi actions currently. let me extract a helper to make it more consistent. 👍

jakehobbs: i should also test that the alchemyAuth connector works as expected when using it outside of the context of an alchemySmartWallet.

packages/connectors-web/tests/alchemyAuth.test.ts
jakehobbs: good spot

packages/react/src/hooks/useAuthSession.ts
jakehobbs: digging into this

packages/connectors-web/src/alchemyAuth.ts
jakehobbs: hmmm the connectors-web-example is having tons of problems for me now. debugging...

jakehobbs: actually i think i just had some weird caching issue.

packages/react/src/hooks/useHandleOauthRedirect.ts
thebrianchen: did some more research, seems like main pros are:
- cleaner UX that prevents confusion
- avoid refresh and retrying stale params

main cons are that it's harder to debug, but i think we can mitigate this by logging the error.

packages/connectors-web/src/alchemyAuth.ts
jakehobbs: i was a bit concerned b/c in the react demo when reconnecting, the status goes disconnected -> connecting -> connected. but in the wagmi core demo, it goes through a "reconnecting" status.

but then i tested w/ metmask connector and the same thing happens. so i am fairly certain this is just a nuance w/ how wagmi works in react.

packages/react/src/hooks/useAuthSession.ts
jakehobbs: he `onConnect` event doesn't fire from `useAccountEffect` when restoring the existing auth session, but it doesn't happen for metamask either, so i believe this is expected due to how some internal wagmi things work.

jakehobbs: added a `"connect"` event when resuming session.

examples/react-v5-example/src/app/page.tsx
```diff
@@ -126,6 +169,159 @@ const ChainControls = () => {
...
+        <div className="flex gap-2">
+          <button
+            onClick={() => {
+              const email = prompt("Enter new email address:");
+              if (!email) return;
+              sendVerificationCode({ contact: email, type: "email" });
+            }}
+            disabled={
+              isSendingCode && sendVerificationCodeVars.type === "email"
+            }
```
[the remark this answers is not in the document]
jakehobbs: tanstack's types are smart enough to know that the vars are defined if the status is pending 🚀

examples/react-v5-example/src/app/page.tsx
```diff
@@ -256,6 +452,54 @@ const SmsAuthDemo = () => {
...
+          onClick={() => {
+            loginWithOauth({
+              provider: "google",
+              mode: "popup",
+            });
+          }}
+        >
+          {isPending && variables.mode === "popup"
+            ? "Logging in..."
+            : "Google (Popup)"}
```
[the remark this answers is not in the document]
jakehobbs: tanstack's types are smart enough to know that the vars are defined if the status is pending 🚀

packages/connectors-web/src/alchemyAuth.ts
jakehobbs: i think we are okay b/c everyone calling `getAuthSession` is aware that it might return null. this is handled within our internal hooks now using `assertAuthSession`. should be good.

PR #2234, opened 2025-11-05
feat: add logging

thebrianchen:
Add logging capabilities throughout the package + add logs at different log levels where it made obvious sense. Probably still room for more logs to be added.

Sample log output:
```
  [13:44:02.208] [@alchemy/wallet-apis@0.0.0] prepareSign:start {"type":"personal_sign"}
  [13:44:02.309] [@alchemy/wallet-apis@0.0.0] signPreparedCalls:start {"type":"user-operation-v060"}
  [13:44:02.361] [@alchemy/wallet-apis@0.0.0] signPreparedCalls:unexpected-call-type {"type":"unknown-type"}
  [13:44:02.512] [@alchemy/wallet-apis@0.0.0] signSignatureRequest:unknown-type {"type":"some-weird-type"}
  [13:44:02.589] [@alchemy/wallet-apis@0.0.0] prepareCalls:start {"callsCount":3,"hasCapabilities":true}
  [13:44:02.615] [@alchemy/wallet-apis@0.0.0] prepareSign:no-from {"hasClientAccount":false}
```

Logger Features
- level-based logging (error, warn, info, debug, verbose)
- structured data for logging arbitrary json objects
- redaction handler to redact sensitive fields
- namespace filtering to enable logs for only a subset of namespaces
- lazy evaluation to only log expensive computations when the appropriate log level is enabled
- profiling wrapper to measure log execution time for sync/async functions (currently unused within the sdk, but we can wrap common methods later)

Components:
- `createLogger`: standard logger created by each individual package to log 
- helper methods to set/set logger config and check global log level and enabled namespaces
- `InMemorySink`: mainly used for testing the logger within the SDK, but can be used to aggregate logs outside of console output if desired

# Pull Request Checklist

---

## PR-Codex overview
This PR primarily focuses on enhancing the logging functionality across multiple packages by introducing a `LOGGER` instance, improving diagnostics, and documenting various logging methods.

### Detailed summary
- Added `LOGGER` instances to `aa-infra`, `wagmi-core`, `wallet-apis`, and `smart-accounts`.
- Introduced new documentation files for logging methods in `InMemorySink`.
- Enhanced logging in `wallet-apis` actions with debug information.
- Implemented `InMemorySink` for capturing logs during testing.
- Updated `createLogger` and related logging configurations in `common`.

> The following files were skipped due to too many changes: `packages/common/tests/logging/config.test.ts`, `packages/common/tests/logging/createLogger.test.ts`


packages/common/src/logging/config.ts
```diff
@@ -0,0 +1,358 @@
...
+ * const data = {
+ *   apiKey: "secret123",
+ *   userId: "user-456"
+ * };
+ *
+ * const redacted = redactObject(data);
+ * // { apiKey: "[REDACTED]", userId: "user-456" }
+ * ```
+ */
+export function redactObject(
```
blakecduncan, line 251: Just wanted to flag that this will only do a shallow redaction. 

ie this won't be redacted

```typescript
logger.info("user data", { 
  user: { apiKey: "secret123" }
})
```

My be worth adding a comment about this, unless we wanna update it to do a deep redaction

packages/common/tests/logging/config.test.ts
```diff
@@ -0,0 +1,525 @@
...
+      const levels = [
+        { value: "0", expected: LogLevel.ERROR },
+        { value: "1", expected: LogLevel.WARN },
+        { value: "2", expected: LogLevel.INFO },
+        { value: "3", expected: LogLevel.DEBUG },
+        { value: "4", expected: LogLevel.VERBOSE },
+      ];
+
+      for (const { value } of levels) {
+        process.env.AA_LOG_LEVEL = value;
```
blakecduncan: It doesn't look like the config.ts uses `aa_log_level`.  Should this be `ALCHEMY_LOG_LEVEL` ?

packages/common/src/logging/config.ts
```diff
@@ -0,0 +1,358 @@
...
+ * setGlobalLoggerConfig({
+ *   sinks: [consoleSink]
+ * });
+ * ```
+ */
+export function consoleSink(entry: LogEntry): void {
+  const { ts, level, message, data, context } = entry;
+  const timestamp = formatTimestamp(ts);
+  const prefix = `[${timestamp}] [${context.package}@${context.version}]`;
+  const dataStr = formatData(data);
```
blakecduncan, line 345: One thing gpt5 caught is that formatData uses `JSON.stringify()` which will throw an error if a bigint is passed.  I think it's probably pretty likely that at some point a big int will be logged

packages/common/src/logging/config.ts
```diff
@@ -0,0 +1,358 @@
...
+ */
+export function isNamespaceEnabled(namespace: string | undefined): boolean {
+  const { enabledNamespaces } = getGlobalLoggerConfig();
+
+  // If no filter is set or it's an empty array, all namespaces are enabled
+  if (!enabledNamespaces || enabledNamespaces.length === 0) {
+    return true;
+  }
+
+  // If namespace is undefined, only allow if undefined is explicitly in the list
```
blakecduncan: nit: I think this comment is a little confusing.  It doesn't look like we check namespace agains the list at all if namespace is undefined

packages/common/src/logging/config.ts
```diff
@@ -0,0 +1,358 @@
...
+ * Lower numeric values indicate higher priority (more restrictive filtering).
+ *
+ * @example
+ * ```ts
+ * import { LogLevel, setGlobalLoggerConfig } from "@alchemy/common";
+ *
+ * setGlobalLoggerConfig({ level: LogLevel.DEBUG });
+ * ```
+ */
+export enum LogLevel {
```
jakehobbs: strong preference to not use a a real typescript `enum`. would prefer to just use a union of string literals or an enum-like object.

recommended reading:
- https://www.totaltypescript.com/why-i-dont-like-typescript-enums
- https://www.totaltypescript.com/books/total-typescript-essentials/deriving-types#using-as-const-for-javascript-style-enums

packages/common/src/logging/config.ts
```diff
@@ -0,0 +1,358 @@
...
+export function consoleSink(entry: LogEntry): void {
+  const { ts, level, message, data, context } = entry;
+  const timestamp = formatTimestamp(ts);
+  const prefix = `[${timestamp}] [${context.package}@${context.version}]`;
+  const dataStr = formatData(data);
+  const output = `${prefix} ${message}${dataStr}`;
+
+  switch (level) {
+    case LogLevel.ERROR:
+      // eslint-disable-next-line no-console
```
jakehobbs: we aren't using the `no-console` eslint rule, are we?

packages/common/src/logging/config.ts
```diff
@@ -0,0 +1,358 @@
...
+      warn: LogLevel.WARN,
+      info: LogLevel.INFO,
+      debug: LogLevel.DEBUG,
+      verbose: LogLevel.VERBOSE,
+    };
+    const normalized = envLevel.toLowerCase();
+    if (normalized in map) {
+      level = map[normalized];
+    } else {
+      const asNum = Number(envLevel);
```
jakehobbs: do we really need to support both numbers and strings for the env level? imo we can simplify this and just use the strings.

if we do that & stop using the `enum`, we can simply do this here:

```ts
LogLevel[envLevel.toUpperCase() as keyof typeof LogLevel];
```

packages/common/src/logging/config.ts
```diff
@@ -0,0 +1,358 @@
...
+      if (!Number.isNaN(asNum) && asNum >= 0 && asNum <= 4)
+        level = asNum as LogLevel;
+    }
+  }
```
jakehobbs, line 103: if the `envLevel` is an unexpected value, should we error and use the default level?

packages/common/src/logging/config.ts
```diff
@@ -0,0 +1,358 @@
...
+ * import { getGlobalLoggerConfig } from "@alchemy/common";
+ *
+ * const config = getGlobalLoggerConfig();
+ * console.log("Current log level:", config.level);
+ * ```
+ */
+export function getGlobalLoggerConfig(): Required<DiagnosticsConfig> {
+  if (!globalConfig) {
+    globalConfig = defaultConfig();
+  }
```
jakehobbs: shorthand for this, if you want to use it
```suggestion
  globalConfig ??= defaultConfig();
```

packages/common/src/logging/config.ts
jakehobbs: you can still use numbers as the values

packages/common/src/logging/config.ts
jakehobbs: +1

packages/common/src/logging/config.ts
thebrianchen: thanks, added logic and tests for deep redaction

packages/common/tests/logging/config.test.ts
thebrianchen: my bad, was changing them out and missed this

packages/common/src/logging/config.ts
thebrianchen: good catch, thanks

packages/common/src/logging/config.ts
thebrianchen: updated comment

packages/common/src/logging/config.ts
thebrianchen: updated

packages/common/src/logging/config.ts
thebrianchen: removed

packages/common/src/logging/config.ts
thebrianchen: done

jakehobbs (approved):

PR #2270, opened 2025-11-27
fix(V5): widen multi-owner light account owner type

jakehobbs:
# Pull Request Checklist

---

## PR-Codex overview
This PR modifies the `ToMultiOwnerLightAccountParams` type in the `multi-owner-account.ts` file to change the structure of the `owners` property and updates the default value for `owners` in the `toMultiOwnerLightAccount` function.

### Detailed summary
- Updated `owners` type from `OneOf<JsonRpcAccount | LocalAccount>[]` to `[OneOf<JsonRpcAccount | LocalAccount>, ...{ address: Address }[]]`.
- Removed the default value assignment for `owners`, allowing it to be passed directly.

### DIFF
--- a/packages/smart-accounts/src/light-account/accounts/multi-owner-account.test.ts
+++ b/packages/smart-accounts/src/light-account/accounts/multi-owner-account.test.ts
@@ -269,7 +269,7 @@ describe.sequential("MultiOwner Light Account Tests", () => {
   }) => {
     const account = await toMultiOwnerLightAccount({
       client: signer,
-      accountAddress: accountAddress,
+      accountAddress,
       salt: _salt ?? salt++,
       owners: [signer.account, ...(owners ?? [])],
     });


packages/smart-accounts/src/light-account/accounts/multi-owner-account.ts
```diff
@@ -32,7 +32,7 @@ export type MultiOwnerLightAccount = LightAccountBase<
 
 export type ToMultiOwnerLightAccountParams = {
   client: Client<Transport, Chain, JsonRpcAccount | LocalAccount | undefined>;
-  owners: OneOf<JsonRpcAccount | LocalAccount>[];
+  owners: [OneOf<JsonRpcAccount | LocalAccount>, ...{ address: Address }[]];
```
dancoombs, line 35: For my understanding, is the `client` already a signer? Seems odd to pass the signer in twice, could this instead be `additionalOwners` and be just an array of addresses?

dancoombs (approved):

packages/smart-accounts/src/light-account/accounts/multi-owner-account.ts
jakehobbs: the `client` is not necessarily a signer. it's just a client that has a transport + chain, and may or may not have an account on it at all. in all of our SCA implementations, we explicitly pass in the `owner`.

this does feel a little weird, but it's how all of viem's built-in SCA implementations work too: https://viem.sh/account-abstraction/accounts/smart/toCoinbaseSmartAccount

the only thing the `client` is used for is getting the chain id & interacting w/ the chain. the account on it is optional & not used directly (unless the `owner` is a `JsonRpcAccount`, in which case the signing methods on the transport are used).

PR #2466, opened 2026-04-22
feat(v4): track key exports for migration visibility

jakehobbs:
## Summary
- Adds a fire-and-forget call to `/v1/track-key-export` after a successful wallet key export in the signer client
- Sends org ID, account address, and chain type (ETH/SOL) for export tracking
- Adds the new endpoint type definition to `SignerEndpoints`

## Test plan
- [ ] Verify key export flow still works end-to-end
- [ ] Confirm tracking request is sent after successful export
- [ ] Confirm failed tracking request does not break the export flow (fire-and-forget with `.catch(() => {})`)

---

## PR-Codex overview
This PR introduces a new API endpoint for tracking key exports and updates several type definitions in the codebase. It enhances the functionality related to account management and multi-owner operations.

### Detailed summary
- Added new endpoint `"/v1/track-key-export"` with request `Body` containing `orgId`, `accountAddress`, and `type`.
- Updated type definitions for various events and parameters in `account-kit/signer/src/client/types.ts`.
- Adjusted existing documentation references to reflect updated line numbers for type definitions.

### DIFF
--- a/account-kit/signer/src/client/base.ts
+++ b/account-kit/signer/src/client/base.ts
@@ -1553,6 +1553,12 @@ export abstract class BaseSignerClient<
       exported?.activity.result.exportWalletAccountResult?.exportBundle;
     if (!exportBundle) throw new Error("No export bundle found");
 
+    this.request("/v1/track-key-export", {
+      orgId: organizationId,
+      accountAddress: account.address,
+      type: opts.type === "ETHEREUM" ? "ETH" : "SOL",
+    }).catch(() => {});
+
     const parsedExportBundle = JSON.parse(exportBundle);
     const signedData = JSON.parse(
       fromHex(`0x${parsedExportBundle.data}`, { to: "string" }),
--- a/account-kit/signer/src/client/types.ts
+++ b/account-kit/signer/src/client/types.ts
@@ -449,9 +449,7 @@ export type SignerEndpoints = [
       accountAddress: string;
       type: "ETH" | "SOL";
     };
-    Response: {
-      success: boolean;
-    };
+    Response: {};
   },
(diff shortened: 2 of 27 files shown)


account-kit/signer/src/client/base.ts
```diff
@@ -1553,6 +1553,12 @@ export abstract class BaseSignerClient<
       exported?.activity.result.exportWalletAccountResult?.exportBundle;
     if (!exportBundle) throw new Error("No export bundle found");
 
+    this.request("/v1/track-key-export", {
+      orgId: organizationId,
+      accountAddress: account.address,
+      type: opts.type === "ETHEREUM" ? "ETH" : "SOL",
+    }).catch(() => {});
+
```
florrdv, line 1561: The cleaner way of doing this would be to proxy the `exportWalletAccount` call above through our signer service and logging it as it's exported. We do this for a few other endpoints (take in the stamped request and forward the result back)

Given that we're winding this down I feel less strongly, but if we weren't I'd want to do that instead. Your call

jakehobbs: i feel like not doing that since we are winding down.

also feels like slightly better optics to not proxy anything w/ key material through us, even if it is encrypted.

but lmk if you feel strongly & i'll change.

florrdv: pretty sure we alr proxy multi owner key exports FYI. another reason to prefer proxying is that people can't just spam us to pollute our db. happy with this though per our conversation on slack

florrdv (dismissed):

avarobinson (approved):

PR #2473, opened 2026-04-24
feat(wallet-apis): add solana support

jakehobbs:
# Pull Request Checklist

---

## PR-Codex overview
This PR focuses on enhancing the `wallet-apis` package by introducing Solana support, adding new types, actions, and utilities for handling Solana transactions, and improving the existing structure for better integration with Solana wallets.

### Detailed summary
- Added support for `SolanaSmartWalletClient` and `SolanaSigner`.
- Introduced new actions: `signPreparedCalls`, `sendPreparedCalls`, and `getCallsStatus`.
- Updated existing types and interfaces to accommodate Solana functionality.
- Improved error handling with `SolanaSignerError`.
- Refactored `prepareCalls` to handle Solana-specific logic.
- Added documentation for new types and methods.
- Enhanced tests for Solana-related functionalities.

> The following files were skipped due to too many changes: `packages/wallet-apis/src/client.solana.e2e.test.ts`, `packages/wallet-apis/src/adapters/signerEquivalence.test.ts`, `packages/wallet-apis/src/client.ts`, `docs/pages/reference/wallet-apis/src/exports/README.mdx`, `pnpm-lock.yaml`


packages/wallet-apis/package.json
```diff
@@ -56,14 +56,20 @@
   },
   "dependencies": {
     "@alchemy/common": "workspace:*",
-    "@alchemy/wallet-api-types": "^0.1.0-alpha.27",
+    "@alchemy/wallet-api-types": "0.1.0-local.0",
```
jakehobbs: # ⚠️ 

be sure to update this before merging

packages/wallet-apis/src/exports/solana.ts
```diff
@@ -0,0 +1,28 @@
...
+export { waitForCallsStatus } from "../actions/solana/waitForCallsStatus.js";
+
+// Decorator.
+export type { SolanaSmartWalletActions } from "../decorators/solanaSmartWalletActions.js";
+export { solanaSmartWalletActions } from "../decorators/solanaSmartWalletActions.js";
+
+// Adapters.
+export { fromKeypair } from "../adapters/fromKeypair.js";
+export type { SolanaKeypairSigner } from "../adapters/fromKeypair.js";
+export { fromKitSigner } from "../adapters/fromKitSigner.js";
```
blakecduncan, line 25: Merge blocker: `@alchemy/wallet-apis/solana` statically re-exports adapters that import optional peers (`@solana/kit`, `@solana/web3.js`). Since those peers are marked optional, importing Solana actions or `fromKeypair` should not require both packages to be installed.

Please split peer-dependent adapters into separate subpaths/lazy modules, or explicitly change/document the install contract for the `/solana` entrypoint.

packages/wallet-apis/src/adapters/fromKeypair.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @returns {SolanaSigner} A SolanaSigner compatible with `createSmartWalletClient`
+ */
+export function fromKeypair(signer: SolanaKeypairSigner): SolanaSigner {
+  return {
+    address: signer.address,
+    async signTransaction({ transaction }) {
+      const numSigs = transaction[0];
+      const messageStart = 1 + numSigs * 64;
+      const messageBytes = transaction.slice(messageStart);
+      const sig = await signer.signMessage(messageBytes);
```
blakecduncan: Merge blocker: validate the returned Ed25519 signature length before writing it into the transaction. `signedTx.set(sig, offset)` silently leaves part of the 64-byte slot as zeros when `sig.length < 64`, and a non-Ed25519 signer can produce an invalid transaction with no clear error.

Please throw a `BaseError` unless `sig.length === 64`; the same check should be added to `fromKitSigner` for `sigDict[signer.address]`.

packages/wallet-apis/src/client.ts
```diff
@@ -1,11 +1,73 @@
...
+export const SOLANA_CHAINS: Record<SolanaChainId, SolanaChainDef> = {
+  "solana:mainnet": solanaMainnet,
+  "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp": solanaMainnet,
+  "solana:devnet": solanaDevnet,
+  "solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1": solanaDevnet,
+};
+
+// ── Validation ──────────────────────────────────────────────────────────
+
+const BASE58_RE = /^[1-9A-HJ-NP-Za-km-z]{32,44}$/;
```
blakecduncan: Merge blocker: this regex only checks base58 character set and encoded string length, but a valid Solana address must decode to exactly 32 bytes. A 33- or 34-byte value can still match this regex.

Please validate with `Base58.toBytes(address).length === 32` and wrap decode failures in a `BaseError`.

packages/wallet-apis/src/client.ts
```diff
@@ -1,11 +1,73 @@
...
 import { isLocalAccount } from "./utils/assertions.js";
 import type { AlchemyWalletTransport } from "./transport.js";
+import type { SolanaChainId } from "@alchemy/wallet-api-types";
 
-export type CreateSmartWalletClientParams = {
+// ── Solana chain definitions ────────────────────────────────────────────
+
+const solanaMainnet: SolanaChainDef = {
+  ...defineChain({
+    id: 101,
```
blakecduncan: Merge blocker: `client.chain.id` is public runtime state, and `101`/`103` collide with real EVM chain IDs in chain lists. Once released, changing these IDs would be breaking for users who read or compare `client.chain.id`.

Please use large synthetic sentinel IDs or another Solana-branded chain shape before shipping.

packages/wallet-apis/src/client.solana.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    const kitSigner = await generateKeyPairSigner();
+
+    const signer = fromKitSigner(kitSigner);
+
+    client = createSmartWalletClient({
+      transport: apiTransport,
+      chain: "solana:devnet",
+      signer,
+      paymaster: {
+        policyId: process.env.TEST_SOLANA_PAYMASTER_POLICY_ID!,
```
blakecduncan, line 58: Merge blocker: this e2e suite depends on `TEST_SOLANA_PAYMASTER_POLICY_ID`, but I don't see that env wired in `.github/workflows/`. This is currently the only end-to-end coverage of the Solana flow, so CI should either run it intentionally or make the skipped coverage explicit.

Please wire the CI secret/env, or gate the suite with an explicit `skipIf` so the gap is visible.

packages/wallet-apis/src/actions/solana/signSignatureRequest.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function extractSignerSignature(
+  unsignedTx: Uint8Array,
+  signedTx: Uint8Array,
+  signerAddress: string,
+): Uint8Array {
+  const numSigs = signedTx[0];
+
+  for (let i = 0; i < numSigs; i++) {
```
blakecduncan: Non-blocking architecture note: signer-slot resolution currently depends on "first empty slot became filled" rather than resolving the signer slot from `signer.address` in the transaction message. That seems acceptable only if the Wallet API contract is that `sendPreparedCalls` receives a raw 64-byte signature and the server places it using `data.signer`.

Please record that decision in the PR/spec or add a code comment here. If the API ever changes to send/use the fully signed transaction body, this heuristic becomes a correctness bug for multi-signer or non-zero signer-slot transactions.

packages/wallet-apis/src/actions/solana/getCallsStatus.ts
```diff
@@ -0,0 +1,78 @@
...
+
+type SolanaGetCallsStatusResponse = Exclude<
+  GetCallsStatusResponse,
+  { details: unknown }
+>;
+
+function isSolanaResponse(
+  response: GetCallsStatusResponse,
+): response is SolanaGetCallsStatusResponse {
+  return !("details" in response);
```
blakecduncan: Non-blocking: this discriminator is load-bearing on EVM responses always having `details`. If that schema ever becomes optional, a non-Solana response could be treated as Solana.

Consider discriminating on a positive Solana marker instead, e.g. `typeof response.chainId === "string"` for the `solana:*` chain ID shape.

packages/wallet-apis/src/actions/solana/waitForCallsStatus.ts
```diff
@@ -0,0 +1,84 @@
...
+  status?: (result: SolanaGetCallsStatusResult) => boolean;
+  throwOnFailure?: boolean;
+  timeout?: number;
+};
+
+export type SolanaWaitForCallsStatusResult = SolanaGetCallsStatusResult;
+
+/**
+ * Polls getCallsStatus until the call reaches a terminal state.
+ * Mirrors viem's waitForCallsStatus behavior: retry with exponential backoff,
```
blakecduncan: Non-blocking: the JSDoc says this mirrors viem's `waitForCallsStatus`, but the params do not include `AbortSignal` and the retry/polling behavior is custom. Users may infer stronger parity than the implementation provides.

Either add `signal?: AbortSignal` support or soften this wording to something like "similar API to viem's waitForCallsStatus".

packages/wallet-apis/src/adapters/fromWalletAdapter.ts
```diff
@@ -0,0 +1,37 @@
+import type { SolanaSigner } from "../types.js";
```
blakecduncan, line 1: Just looked at your demo code - do you think we should add a `fromWalletStandard` adapter as well.  I think that could reduce some boiler plate code that devs will have to implement

had claude write this up for me as an example
```typescript
import { BaseError } from "@alchemy/common";
import type { SolanaSigner } from "../types.js";

const SIGN_TRANSACTION = "solana:signTransaction";

/**
 * Minimal structural shape of a wallet-standard account.
 *
 * Pass the account object returned by `standard:connect` or `wallet.accounts`.
 * Some wallets rely on fields beyond `address`, so do not construct this
 * object manually from an address string.
 */
export interface WalletStandardAccount {
  address: string;
  publicKey?: Uint8Array;
  chains?: readonly string[];
  features?: readonly string[];
  [key: string]: unknown;
}

/** Minimal structural shape of a wallet-standard wallet. */
export interface WalletStandardWallet {
  features: { readonly [name: string]: unknown };
}

interface SolanaSignTransactionFeature {
  readonly signTransaction: (
    ...inputs: readonly {
      account: WalletStandardAccount;
      transaction: Uint8Array;
    }[]
  ) => Promise<readonly { signedTransaction: Uint8Array }[]>;
}

function isSolanaSignTransactionFeature(
  feature: unknown,
): feature is SolanaSignTransactionFeature {
  return (
    typeof feature === "object" &&
    feature !== null &&
    "signTransaction" in feature &&
    typeof (feature as { signTransaction?: unknown }).signTransaction ===
      "function"
  );
}

/**
 * Adapts a Wallet Standard wallet into a {@link SolanaSigner}.
 */
export function fromWalletStandard(
  wallet: WalletStandardWallet,
  account: WalletStandardAccount,
): SolanaSigner {
  const feature = wallet.features[SIGN_TRANSACTION];

  if (!isSolanaSignTransactionFeature(feature)) {
    throw new BaseError(
      `Wallet does not support the "${SIGN_TRANSACTION}" feature.`,
    );
  }

  if (account.features && !account.features.includes(SIGN_TRANSACTION)) {
    throw new BaseError(
      `Account ${account.address} does not support the "${SIGN_TRANSACTION}" feature.`,
    );
  }

  return {
    address: account.address,
    async signTransaction({ transaction }) {
      const [output] = await feature.signTransaction({ account, transaction });

      if (!output) {
        throw new BaseError("Wallet returned no signed transaction.");
      }

      if (!(output.signedTransaction instanceof Uint8Array)) {
        throw new BaseError("Wallet returned an invalid signed transaction.");
      }

      return output;
    },
  };
}
```

packages/wallet-apis/src/client.solana.e2e.test.ts
jakehobbs: i don't think these e2e tests don't run on this repo. these run from wallet server's repo. i'll double check though.

packages/wallet-apis/src/exports/solana.ts
jakehobbs: lazily importing them

packages/wallet-apis/src/adapters/fromKeypair.ts
jakehobbs: Both adapters now throw a BaseError before signedTx.set(sig, offset) if the signature isn't exactly 64 bytes.

packages/wallet-apis/src/client.ts
jakehobbs: Replaced the regex-only Solana address check (BASE58_RE) with isAddress from @solana/kit (aliased as isSolanaAddress to avoid collision with viem's isAddress). The @solana/kit implementation does a proper base58 decode and asserts exactly 32 bytes, rejecting 33/34-byte values that the old regex would accept. Decode failures are caught internally by isAddress, and our wrapper throws a BaseError on invalid input.

packages/wallet-apis/src/client.ts
jakehobbs: any thoughts on best approach here?

jakehobbs: i think i'm just gonna put 0 since we have a separate `solanaChainId` that we reference internally.

packages/wallet-apis/src/client.solana.e2e.test.ts
jakehobbs: confirmed

packages/wallet-apis/src/actions/solana/signSignatureRequest.ts
jakehobbs: Added the architectural decision comment to both [fromKeypair.ts:36-40](vscode-webview://0olak8elotg355esled0c9p453457b2437iaq7k2857vp7vtohpo/packages/wallet-apis/src/adapters/fromKeypair.ts#L36-L40) and [fromKitSigner.ts:52-56](vscode-webview://0olak8elotg355esled0c9p453457b2437iaq7k2857vp7vtohpo/packages/wallet-apis/src/adapters/fromKitSigner.ts#L52-L56). The comment documents that the "first empty slot" heuristic is safe only because sendPreparedCalls sends a raw 64-byte signature and the server resolves placement via data.signer, and flags the specific breakage scenario (multi-signer / non-zero signer-slot) if the API contract ever changes.

packages/wallet-apis/src/actions/solana/getCallsStatus.ts
jakehobbs: Changed the Solana response discriminator in getCallsStatus from checking for the absence of details (!("details" in response)) to checking for a positive solana: chain ID prefix — both at the type level (Extract<..., { chainId: solana:${string} }>) and at runtime (chainId.startsWith("solana:")). Safe if details becomes optional on EVM or if EVM adopts string chain IDs.

packages/wallet-apis/src/adapters/fromWalletAdapter.ts
jakehobbs: added

packages/wallet-apis/src/actions/solana/waitForCallsStatus.ts
jakehobbs: keeping parity w/ viem, updated jsdoc

packages/wallet-apis/src/actions/solana/getCallsStatus.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  const decoded = fullDecoded;
+  const statusCode = decoded.status;
+  const status = (() => {
+    if (statusCode >= 100 && statusCode < 200) return "pending" as const;
+    if (statusCode >= 200 && statusCode < 300) return "success" as const;
+    if (statusCode >= 300 && statusCode < 700) return "failure" as const;
+    throw new BaseError(`Unknown Solana call status code: ${statusCode}`);
+  })();
```
dancoombs, line 70: nit: Is this different than eth? If not probably a chance to DRY

packages/wallet-apis/src/actions/solana/prepareCalls.ts
```diff
@@ -0,0 +1,122 @@
...
+ *     programId: "11111111111111111111111111111111",
+ *     data: "0x...",
+ *   }],
+ *   capabilities: {
+ *     paymaster: { policyId: "your-policy-id" }
+ *   }
+ * });
+ * ```
+ */
+export async function prepareCalls(
```
dancoombs, line 77: These actions feel like they're duplicating a bunch of code from the EVM actions. Probably not a huge deal while we only support EVM & Solana, but might get annoying when we add the 3rd type.

Is there a possible refactor where we inject the chain specific logic into the existing actions, or is that too messy?

packages/wallet-apis/src/actions/solana/getCallsStatus.ts
jakehobbs: we don't have this for eth at all. we use viem's action, which is incompatible.

packages/wallet-apis/src/actions/solana/prepareCalls.ts
jakehobbs: my initial implementation actually had everything shared, but it got pretty messy. i think this is a lot cleaner.

jakehobbs: this is where i refactored it: https://github.com/alchemyplatform/aa-sdk/pull/2473/commits/be02dfa17678f01d0e79fc361a63690fb4068686

jakehobbs: also note that evm uses `getCallsStatus` and `waitForCallStatus` from viem, which have incompatible types & runtime behavior w/ solana.

dancoombs: Got it, maybe it makes sense to go down this route when we add a 3rd VM/transaction type

dancoombs (approved):
🚢

jakehobbs:
Discovered new bugs w/ typebox, so may release Zod migration before this.

PR #2479, opened 2026-04-30
fix: clarify requestQuoteV0 chain ID errors

Richard-Dang:
# Pull Request Checklist

- [x] Did you add new tests and confirm existing tests pass? (`pnpm test`)
- [ ] Did you update relevant docs? (docs are found in the `docs` folder, and guidelines for updating/adding docs can be found in the [contribution guide](https://github.com/alchemyplatform/aa-sdk/blob/main/CONTRIBUTING.md))
- [x] If you have a breaking change, is it [correctly reflected in your commit message](https://www.conventionalcommits.org/en/v1.0.0/#examples)? (not a breaking change)
- [x] Did you run lint (`pnpm run lint:check`) and fix any issues? (`pnpm run lint:write`)

## Summary

Adds SDK-facing validation for `requestQuoteV0` chain ID params before TypeBox encodes them to raw RPC hex strings. This turns accidental raw-RPC inputs like `chainId: "0xa4b1"` into a clear SDK error that asks for a number like `42161`.

## Testing

- `pnpm --filter @alchemy/common build`
- `pnpm --filter @alchemy/wallet-apis build`
- `pnpm exec prettier --check packages/wallet-apis/src/experimental/actions/requestQuoteV0.ts packages/wallet-apis/src/experimental/actions/requestQuoteV0.test.ts`
- `OPENAI_API_KEY=dummy pnpm exec eslint packages/wallet-apis/src/experimental/actions/requestQuoteV0.ts packages/wallet-apis/src/experimental/actions/requestQuoteV0.test.ts`
- Inline runtime check with `tsx` verified numeric `chainId`/`toChainId` encode to hex and hex strings reject before transport

Note: `pnpm --filter @alchemy/wallet-apis test:run src/experimental/actions/requestQuoteV0.test.ts` is blocked locally by the repo global Vitest setup calling `evm_setAutomine` against the local test RPC before this suite runs.

---

## PR-Codex overview
This PR introduces validation for chain IDs in the `requestQuoteV0` function, ensuring they are numeric before proceeding with requests. It also adds corresponding tests to confirm the functionality and error handling.

### Detailed summary
- Added `assertOptionalChainId` function to validate chain ID parameters.
- Updated `requestQuoteV0` to call `assertOptionalChainId` for `chainId` and `toChainId`.
- Created tests for `requestQuoteV0` to validate chain ID handling.
- Added tests for rejecting invalid hex string chain IDs.

### DIFF
--- a/packages/wallet-apis/src/experimental/actions/requestQuoteV0.ts
+++ b/packages/wallet-apis/src/experimental/actions/requestQuoteV0.ts
@@ -1,4 +1,5 @@
 import type { Address, Prettify } from "viem";
+import { BaseError } from "@alchemy/common";
 import type { DistributiveOmit, InnerWalletApiClient } from "../../types.ts";
 import {
   fromRpcCapabilities,
@@ -20,6 +21,26 @@ import {
 const schema = methodSchema(MethodSchema);
 type BaseRequestQuoteV0Params = MethodParams<typeof MethodSchema>;
 type RequestQuoteV0Response = MethodResponse<typeof MethodSchema>;
+
+/**
+ * Validates chain IDs in the SDK-facing request shape before TypeBox encodes
+ * them to the raw RPC hex-string shape.
+ *
+ * @param {"chainId" | "toChainId"} name - The parameter name shown in the error message.
+ * @param {unknown} value - The SDK-facing chain ID value to validate.
+ */
+function assertOptionalChainId(
+  name: "chainId" | "toChainId",
+  value: unknown,
+): asserts value is number | undefined {
+  if (value === undefined || typeof value === "number") {
+    return;
+  }
+
+  throw new BaseError(
+    `Invalid params: ${name} must be a number when using requestQuoteV0, e.g. 42161. Hex strings like "0xa4b1" are only valid for raw wallet_requestQuote_v0 RPC calls.`,
+  );
+}
 
 export type RequestQuoteV0Params = Prettify<
   WithCapabilities<
@@ -106,6 +127,12 @@ export async function requestQuoteV0(
           client,
           "capabilities" in params ? params.capabilities : undefined,
         );
+
+  assertOptionalChainId("chainId", params.chainId);
+  assertOptionalChainId(
+    "toChainId",
+    "toChainId" in params ? params.toChainId : undefined,
+  );
 
   const { account: _, chainId: __, ...rest } = params;
   const rpcParams = encode(schema.request, {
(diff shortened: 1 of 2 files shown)


pavelm (approved):

0xfourzerofour (approved):

packages/wallet-apis/src/experimental/actions/requestQuoteV0.ts
```diff
@@ -21,6 +22,26 @@ const schema = methodSchema(MethodSchema);
...
 type RequestQuoteV0Response = MethodResponse<typeof MethodSchema>;
 
+/**
+ * Validates chain IDs in the SDK-facing request shape before TypeBox encodes
+ * them to the raw RPC hex-string shape.
+ *
+ * @param {"chainId" | "toChainId"} name - The parameter name shown in the error message.
+ * @param {unknown} value - The SDK-facing chain ID value to validate.
+ */
+function assertOptionalChainId(
```
jakehobbs, line 32: would rather not have this live in here. we have other actions that accept an optional chain id too.

let's test and see how they behave.

jakehobbs (changes requested):

packages/wallet-apis/src/experimental/actions/requestQuoteV0.ts
jakehobbs: `prepareSign`, `formatSign`, `grandPermission`, and `prepareCalls` all do `params.chainId ?? client.chain.id` as well.

jakehobbs: honestly this feels a little weird to me to have to do though. typescript should already be showing errors if someone tries to pass hex where they shouldn't.

jakehobbs: @Richard-Dang imo the PR you did in wallet-server is the only change we should need here. the error i get currently if i pass a hex where a number should be is this: `FAILED: HexNumber encode error: 0x01 is out of 32-bit unsigned range`, which should be improved after your other change lands.

Richard-Dang: Yeah I think you're right. I added the SDK-side check because I was trying to make the requestQuoteV0 error more specific, but the actual bad error is coming from the shared HexNumber codec during Value.Encode

Since prepareCalls / formatSign / requestQuoteV0 all go through the same codec path, fixing it in wallet-server types is the cleaner place. I'll close this PR and keep the wallet-server one

PR #2509, opened 2026-05-15
fix(smart-accounts): make PermissionBuilder hook state non-mutating

jakehobbs:
## Summary
- \`compileInstallArgs\` silently dropped permissions because it never called \`translatePermissions\`. Any caller adding permissions and then invoking \`compileInstallArgs\` got an empty \`hooks\` array — a security-relevant footgun for session-key flows.
- \`compileRaw\` and \`compileDeferred\` translated correctly but mutated \`this.hooks\`, so repeated or interleaved compile calls duplicated hooks.
- Refactored to treat \`this.hooks\` as immutable after construction. All three compile methods now build the final hook array fresh per call via a shared private \`buildHooks(extraHooks?)\` helper.

## Test plan
- [ ] Verify \`compileInstallArgs\` after \`addPermission\` returns hooks matching \`compileRaw\` output
- [ ] Verify calling any compile method multiple times produces identical hook arrays (no duplication)
- [ ] Verify \`compileDeferred\` still injects the TimeRange hook when deadline is set
- [ ] Verify constructor-provided \`hooks\` still pass through to all compile outputs
- [ ] Run existing \`deferralActions.test.ts\` suite end-to-end

---

## PR-Codex overview
This PR refactors the `PermissionBuilder` class in `permissionBuilder.ts` to improve the handling of hooks, particularly in relation to time range modules and install validation. It introduces new methods for compiling install arguments and encoding hooks, enhancing modularity and maintainability.

### Detailed summary
- Introduced `extraHooks` array for managing additional hooks.
- Updated hook handling logic to use `extraHooks` instead of directly modifying `this.hooks`.
- Created `compileInstallArgs` method to compile install arguments with hooks.
- Added `encodeInstallValidationCall` method for shared encoding logic.
- Implemented `buildHooks` method to assemble hooks without mutating existing hooks.
- Added `encodeRawHooks` method to encode raw hooks into the final hook format.

### DIFF
--- a/packages/smart-accounts/src/ma-v2/permissionBuilder.ts
+++ b/packages/smart-accounts/src/ma-v2/permissionBuilder.ts
@@ -427,6 +427,8 @@ export class PermissionBuilder {
     typedData: DeferredActionTypedData;
     fullPreSignatureDeferredActionPayload: Hex;
   }> {
+    const extraHooks: Hook[] = [];
+
     // Add time range module hook via expiry
     if (this.deadline !== 0) {
       if (this.deadline < Date.now() / 1000) {
@@ -436,7 +438,7 @@ export class PermissionBuilder {
         throw new DeadlineOverLimitError(this.deadline);
       }
 
-      this.hooks.push(
+      extraHooks.push(
         TimeRangeModule.buildHook(
           {
             entityId: this.validationConfig.entityId,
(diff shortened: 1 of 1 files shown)


jakehobbs:
Abandoning — the builder is intended to be used fluently (`new PermissionBuilder().addPermission(...).compile...`), not as a long-lived stateful object across multiple compile calls. The 'compileInstallArgs silently drops permissions' issue is real but the practical risk is low under that usage pattern, and the refactor risk on `compileDeferred` isn't worth it.

PR #2550, opened 2026-07-24
fix(deps): drop tar override that broke the publish pipeline

jakehobbs:
## What

Removes the `tar@<7.5.18` pnpm override added in #2549. That override forced `lerna@8.2.4` onto `tar@7`, whose API is incompatible with lerna's `packDirectory`:

```
lerna ERR! TypeError: Cannot read properties of undefined (reading 'create')
lerna ERR!     at packDirectory (.../node_modules/lerna/dist/index.js:6231:39)
```

`lerna publish` bumps versions, commits, and tags **before** packing/publishing, so the v5.0.8 version bump + git tags landed on `main`, but the actual `npm publish` never ran — leaving npm on 5.0.7 with a dangling 5.0.8 tag.

## Change

- Drop `tar@<7.5.18`. `tar` reverts to its prior resolution (6.2.1 / 7.2.0 / 7.5.21) — the same set that shipped 5.0.7 successfully.
- The `js-yaml` overrides (`3.15.0` / `4.3.0`) are **retained** — they were not implicated in the failure (the CI build steps passed; only lerna's pack step crashed on tar).

`node-tar` shipped no 6.x-compatible patch for the range lerna needs, so there is no `tar` bump that both satisfies the dependency-hygiene goal and keeps lerna 8 working. Moving lerna's `tar` forward would require a lerna v9 major upgrade, which should be evaluated on its own rather than bundled into a lockfile change.

## Verification

- `pnpm install --frozen-lockfile` passes on Node 22.20.0.
- `tar` back to the resolution that published 5.0.7; `js-yaml` remains on 3.15.0 / 4.3.0.

---

## PR-Codex overview
This PR focuses on updating the package dependencies in `package.json` and `pnpm-lock.yaml`, particularly the `tar` package and its versions, while also adding new entries for several packages with their respective resolutions and engine requirements.

### Detailed summary
- Updated `tar` package versions from `7.5.21` to `6.2.1` and `7.2.0`.
- Added new packages: 
  - `chownr@2.0.0` and `fs-minipass@2.1.0`.
  - `minipass@5.0.0` and `mkdirp@1.0.4`.
- Specified engine requirements for several packages.
- Included deprecation warnings for older `tar` versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,8 +100,7 @@
       "@noble/curves": "^1.4.0",
       "eslint-plugin-react-hooks": "4.6.0",
       "js-yaml@<3.15.0": "3.15.0",
-      "js-yaml@>=4 <4.3.0": "4.3.0",
-      "tar@<7.5.18": "7.5.21"
+      "js-yaml@>=4 <4.3.0": "4.3.0"
     }
   },
   "workspaces": [
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -12,7 +12,6 @@ overrides:
   eslint-plugin-react-hooks: 4.6.0
   js-yaml@<3.15.0: 3.15.0
   js-yaml@>=4 <4.3.0: 4.3.0
-  tar@<7.5.18: 7.5.21
 
 importers:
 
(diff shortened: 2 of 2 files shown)


jakehobbs:
Superseded by #2551, which upgrades lerna to v9 — that fixes the pack-step crash *and* keeps the tar override (so alert #524 stays closed), instead of reverting it. Reopen this if the v9 upgrade needs to be held and we want to unblock the release with the revert instead.

